PR #7, opened 2016-03-29
Download templates from Researchworks

arokem:


arokem:
Question regarding naming: is `read_templates` and `fetch_templates` too general? That is, do we need to say that these are certain templates (e.g. `read_templates_17` for the fact that it has 17 ROIs?). Anything else we need to do to version this? We are keeping hashes, which will change if/when the data in the templates changes.

PR #8, opened 2016-03-30
Brain extraction, using Pandas `apply`.

arokem:


arokem:
Still WIP: need to see whether this works on multiple subjects (it works well with just one). Also, need to think about how to test this.

arokem:
I can now confirm that this does work well with more than one subject :-)

PR #9, opened 2016-03-30
built a module to fit dki parameters and write out images

jyeatman:
this should be good to go


arokem:
We'll need to sort out a way to test this stuff properly...

PR #12, opened 2016-04-01
DKI documentation

arokem:
And changed the key-word args to make more sense. cc: @libbyhuber

### DIFF
--- a/AFQ/dki.py
+++ b/AFQ/dki.py
@@ -80,9 +80,9 @@ def fit_dki(data_files, bval_files, bvec_files, mask=None, min_kurtosis=0,
     MD = dkifit.md
     AD = dkifit.ad
     RD = dkifit.rd
-    MK = dkifit.mk(min_d, max_d)
-    AK = dkifit.ak(min_d, max_d)
-    RK = dkifit.rk(min_d, max_d)
+    MK = dkifit.mk(min_kurtosis, max_kurtosis)
+    AK = dkifit.ak(min_kurtosis, max_kurtosis)
+    RK = dkifit.rk(min_kurtosis, max_kurtosis)
 
     maps = [FA, MD, AD, RD, MK, AK, RK]
     names = ['FA', 'MD', 'AD', 'RD', 'MK', 'AK', 'RK']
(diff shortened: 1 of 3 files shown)


jyeatman:
We'll make sure to change the matlab calls appropriately. BTW does this
message automatically go to Libby?

arokem:
I don't know. Let's see if she responds :-)

libbyhuber:
:)

thank you!

arokem:
Would someone else please take a look, and merge if this looks ready?

AFQ/dki.py
```diff
@@ -8,22 +8,48 @@
...
     bvec_files : str or list
-        ditto
+        Equivalent to `data_files`.
+    mask : ndarray, optional
+        Binary mask, set to True or 1 in voxels to be processed.
+        Default: Process all voxels.
+    min_kurtosis : float, optional
+        The minimal plausible value of kurtosis. Default: -1.
+    max_kurtosis : float, optional
+        The maximal plausible value of kurtosis. Default: 3.
```
bcipolli: I tend to leave out default values; I believe these get published in the auto-generated function declaration comment, and this can get stale.

AFQ/dki.py
```diff
@@ -8,22 +8,48 @@
...
+        `data_files`.
+
+    Returns
+    -------
+    file_paths : a dict with the derived maps that were computed and full-paths
+    to the files containing these maps.
+
+    Note
+    ----
+    Maps that are calculated: FA, MD, AD, RD, MK, AK, RK
```
bcipolli: I assume these don't need to be spelled out...

AFQ/dki.py
```diff
@@ -8,22 +8,48 @@
...
+
     """
-    # XXX Check that inputs are uniform
+    types = [type(f) for f in [data_files, bval_files, bvec_files]]
```
bcipolli: I'm not sure if this is the best check. This will fail unicode vs. byte string, or different types of iterables (tuple vs. list). Not perfect, but probably fine, right?

bcipolli:
Made a couple of comments, if you're good I'm good--merge it!

arokem:
Thanks for taking a look. I think I'm probably fine with all these things as is.

PR #13, opened 2016-04-13
DTI and model predictions

arokem:
Check-list for things that I still need to do here: 

- [x] model predictions from parameters
- [x] CLI for both models, including predictions


arokem:
OK - I think this is ready to go, but I would appreciate it if someone could take a look.

arokem:
Sorry - wrong button...

PR #15, opened 2016-04-23
Use pytest instead of nose

arokem:


bcipolli:
Some flake8 failures, in addition to failure for `subprocess` to run.

I'm not a big fan of having top-level `tests` and `testing` directories; too similar for me. I have a slight preference (high enough to mention) for `utils.testing` rather than `testing.utils`.

arokem:
Good call. I will change the testing/utils structure into utils/testing. The subprocess failure still mystifies me

arokem:
OK. This seems to work now. The main reason to abandon nose in favor of pytest is that nose might not be maintained in the long run (unless someone decides to support it).

PR #18, opened 2016-04-27
Automate finding new scripts.

arokem:


arokem:
This might also help deal with the subprocess issues we've been seeing on #15

PR #20, opened 2016-05-05
Tractography

arokem:
First pass at tractography: this does deterministic tracking with CSD. Fitting the model and doing the tracking are separated into two different steps, and CSD parameters are stored as a nifti file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,14 +34,3 @@ script:
 
 after_success:
 - coveralls
-
-before_cache:
-# clean unused packages & installed files from conda cache
-# this makes the cache rebuilt less frequently
-- conda clean --tarballs --packages --index-cache
-- rm -rf $HOME/miniconda/pkgs/cache
-- xargs rm <installed_files.txt
-
-cache:
-  directories:
-    - $HOME/miniconda
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
   - 3.5
 
 env:
-  - CONDA_DEPS="pip flake8 pytest coverage numpy matplotlib pandas" PIP_DEPS="coveralls dipy nibabel pytest-cov boto3 scipy"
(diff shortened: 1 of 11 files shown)


arokem:
Check it out @jyeatman : I managed to write one function that does both deterministic and probabilistic tracking from both the tensor and CSD. It will probably also accommodate other kinds of models, as we add them. So, we can use SFM (which has an ODF), and eventually also DKI (which has an ODF as well), but we can already use CSA, QBall, and other models that have SHM representations as well.

arokem:
Ugh. We seem to be running into https://github.com/nipy/dipy/issues/852

jyeatman:
This should work fine on python 2.7 right?

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://depts.washington.edu/bdelab/

arokem:
Yes. It works fine on 2.7.

It seems that MNE has a fix for this, so I will try to apply that fix here,
and then we can also port it into Dipy.

arokem:
A stitch in time, and all that.

arokem:
OK - with a little bit more elbow-grease, I have this now passing! 

Feel free to take a look, and make comments.

bcipolli:
@arokem What's the best way to review this for you? It's a bit much code to really read through. Any targeted places to check, or just looking to have someone else run it locally to validate?

arokem:
The best help is to say whether the general framework makes sense: a function that takes as arguments parameters and infers the kind of model to use for the local models, and then a switch on prob/det?

In other words, focus on `tractography.py` and `test_tractography.py`

arokem:
But also: no need for a very thorough review at this point, I think. We'll get a better sense when we start using these things. These PRs are here more for everyone's situational awareness.

bcipolli:
@arokem makes sense to me, that part of the code looks pretty clean!

Right, I do appreciate the awareness-building, but also want to make sure you don't get backed up on PRs too much either...

arokem:
I think I'll start merging.

PR #21, opened 2016-05-06
Added some tools for registration. Started testing them.

arokem:
But testing probably needs revision, since this takes a while...

### DIFF
--- a/AFQ/dki.py
+++ b/AFQ/dki.py
@@ -71,7 +71,7 @@ def fit_dki(data_files, bval_files, bvec_files, mask=None,
     names = ['FA', 'MD', 'AD', 'RD', 'MK', 'AK', 'RK', 'params']
 
     if out_dir is None:
-        out_dir = op.join(op.split(data_files[0])[0], 'dki')
+        out_dir = op.join(op.split(data_files)[0], 'dki')
 
     if not op.exists(out_dir):
         os.makedirs(out_dir)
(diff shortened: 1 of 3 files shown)


arokem:
This one should follow #20

arokem:
I'll wait for the testing to finish, just to make sure that I didn't cause a mess here, and then merge this as well.

AFQ/dki.py
```diff
@@ -71,7 +71,7 @@ def fit_dki(data_files, bval_files, bvec_files, mask=None,
     names = ['FA', 'MD', 'AD', 'RD', 'MK', 'AK', 'RK', 'params']
 
     if out_dir is None:
-        out_dir = op.join(op.split(data_files[0])[0], 'dki')
+        out_dir = op.join(op.split(data_files)[0], 'dki')
```
bcipolli: I think you should check if `data_files` is a string at the top, and make a list out of it. In that case, this would bea list comprehension.

arokem: I might be missing something, but I think I am doing that here:
https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/utils/models.py#L28

bcipolli: Nice, right. Then, I'm not sure why this code change would work...

arokem: Maybe it doesn't but I am not testing it enough... Let me see...

arokem: Oh, I see. We do that inside the utility function, so in the calling
function, it retains the same type as when it got in here...

arokem: Working on it here: https://github.com/yeatmanlab/pyAFQ/pull/25

But I have to stop for today...

PR #28, opened 2016-06-07
NF: Add a b0_threshold parameter for model fitting.

arokem:
This is useful for HCP data.

### DIFF
--- a/AFQ/dki.py
+++ b/AFQ/dki.py
@@ -18,7 +18,7 @@ if LooseVersion(dipy.__version__) < '0.12':
 
 
 def fit_dki(data_files, bval_files, bvec_files, mask=None,
-            min_kurtosis=-1, max_kurtosis=3, out_dir=None):
+            min_kurtosis=-1, max_kurtosis=3, out_dir=None, b0_threshold=0):
     """
     Fit the DKI model, save files with derived maps
 
@@ -40,8 +40,10 @@ def fit_dki(data_files, bval_files, bvec_files, mask=None,
         The maximal plausible value of kurtosis. Default: 3.
     out_dir : str, optional
         A full path to a directory to store the maps that get computed.
-        Default: maps get stored in the same directory as the last DWI file in
-        `data_files`.
+        Default: maps get stored in the same directory as the last DWI file
+        in `data_files`.
+    b0_threshold : float
+
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


AFQ/dki.py
```diff
@@ -54,7 +56,9 @@ def fit_dki(data_files, bval_files, bvec_files, mask=None,
 
     """
     img, data, gtab, mask = ut.prepare_data(data_files, bval_files,
-                                            bvec_files, mask=mask)
+                                            bvec_files, mask=mask,
+                                            b0_threshold=0)
```
bcipolli: `b0_threshold=0)` => ` b0_threshold=b0_threshold)`, I believe.

arokem: Yikes! Good catch.

bin/pyAFQ_dki
```diff
@@ -28,7 +28,9 @@ parser.add_option("-o", "--out_dir",
                   help="full path to directory for files to be saved (will be created if it doesn't exist)")
 
 parser.add_option("-m", "--mask", help="mask file")
+parser.add_option('-b', '--b0_threshold', help="b0 threshold", default=0)
+
 opts, args = parser.parse_args()
 
 dki.fit_dki(opts.dwi, opts.bval, opts.bvec, mask=opts.mask,
-            out_dir=opts.out_dir)
+            out_dir=opts.out_dir, b0_threshold=float(opts.b0_threshold))
```
bcipolli: Just a (hopefully helpful) note: you can specify the type right on the option. Should give the user a bit more info, fail at a better point for bad inputs, etc.

bcipolli: https://docs.python.org/2/library/optparse.html#the-store-action

arokem: Sure - definitely helpful. Did I get it right?

bcipolli: I think you'd just change to:
```parser.add_option('-b', '--b0_threshold', help="b0 threshold", type="float", default=0)```

Then do away with the float conversion later. In theory :)

arokem: Like so?
https://github.com/yeatmanlab/pyAFQ/pull/28/commits/1be852b5a1f27a15544b021c1a17dc7cbb54291b

bcipolli:
LGTM!

arokem:
Off we go.

arokem:
Thanks for the review!

PR #30, opened 2016-06-15
Slightly more flexible segmentation API

arokem:
Also moved a bit of cruft into `api`, added a `yield` fix for `orient_by_rois` (and used it to orient the streamlines as they are segmented).


AFQ/_fixes.py
```diff
@@ -81,3 +81,72 @@ def spherical_harmonics(m, n, theta, phi):
     val *= np.exp(0.5 * (gammaln(n - m + 1) - gammaln(n + m + 1)))
     val = val * np.exp(1j * m * theta)
     return val
+
+
+def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
+    """Orient a set of streamlines according to a pair of ROIs
```
bcipolli: a comment explaining what this is overriding and why would be helpful. Looks like other functions in this file do so.

AFQ/_fixes.py
```diff
@@ -81,3 +81,72 @@ def spherical_harmonics(m, n, theta, phi):
...
+
+    if affine is not None:
+        roi1 = apply_affine(affine, roi1)
+        roi2 = apply_affine(affine, roi2)
+
+    # Make a copy, so you don't change the output in place:
+    if copy:
+        new_sl = deepcopy(streamlines)
+    else:
+        new_sl = streamlines
```
bcipolli: `new_sl = deepcopy(streamlines) if copy else streamlines` ? I also think the comment can go away, since the parameter is documented already.

AFQ/api.py
```diff
@@ -12,6 +12,28 @@
...
+#                "SLF", "ARC", "UNC"]
+#
+# # Read in the standard templates:
+# AFQ_TEMPLATES = afd.read_templates()
+# # For the arcuate, we need to rename a few of these and duplicate the SLF ROI:
+# AFQ_TEMPLATES['ARC_roi1_L'] = AFQ_TEMPLATES['SLF_roi1_L']
+# AFQ_TEMPLATES['ARC_roi1_R'] = AFQ_TEMPLATES['SLF_roi1_R']
+# AFQ_TEMPLATES['ARC_roi2_L'] = AFQ_TEMPLATES['SLFt_roi2_L']
+# AFQ_TEMPLATES['ARC_roi2_R'] = AFQ_TEMPLATES['SLFt_roi2_R']
+#
```
bcipolli: I prefer to simply use git history to access old code, rather than adding comments. Also want to make sure that this was intentionally commented out when moved.

AFQ/segmentation.py
```diff
@@ -69,22 +69,87 @@ def segment(fdata, fbval, fbvec, streamlines, bundles=AFQ_BUNDLES,
...
+                data,
+                interpolation='nearest'))
+            # This function requires lists as inputs:
+            select_sl = dts.select_by_rois(select_sl,
+                               [warped_ROI.astype(bool)],
+                               [rule])
+        # Next, we reorient each streamline according to an ARBITRARY, but
+        # CONSISTENT order. To do this, we use the first ROI for which the rule
+        # is True as the first one to pass through, and the last ROI for which
+        # the rule is True as the last one to pass through:
```
bcipolli: :+1: thanks for adding the comment!

AFQ/segmentation.py
```diff
@@ -69,22 +69,87 @@ def segment(fdata, fbval, fbvec, streamlines, bundles=AFQ_BUNDLES,
...
+    img : 3D volume
+
+    streamlines : list of arrays, or array
+
+    weighting : 1D array or 2D array (optional)
+
+    """
+    if isinstance(streamlines, list):
+        # Resample each streamline to the same number of points
+        # list => array
```
bcipolli: perhaps `np.array`? I didn't get this at first... but maybe the issue is, I'm not in Python/Numpy enough these days :)

AFQ/segmentation.py
```diff
@@ -69,22 +69,87 @@ def segment(fdata, fbval, fbvec, streamlines, bundles=AFQ_BUNDLES,
...
+
+    # We assume that weights *always sum to 1 across streamlines*:
+    if weighting is None:
+        w = np.ones(values.shape) / values.shape[0]
+
+    tract_profile = np.sum(w * values, 0)
+    return tract_profile
+
+
+# def gaussian_weights(fgarray):
```
bcipolli: Another commented block of code to double-check on

arokem:
I am going to go ahead and merge this one, so we can move on.

PR #33, opened 2016-07-01
WIP: putting together a pipeline

arokem:
More work on the API. Needs a lot more thinking.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -197,13 +197,13 @@ class AFQ(object):
             self.set_brain_mask()
 
         return self.data_frame['brain_mask_img'].apply(
-                                        nib.Nifti1Image.get_data)
+            nib.Nifti1Image.get_data)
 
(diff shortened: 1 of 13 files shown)


AFQ/api.py
```diff
@@ -122,59 +154,165 @@ def __init__(self, raw_path=None, preproc_path=None,
...
+                self.data_frame['dwi_file'].apply(nib.load)
+
+    def get_dwi_data_img(self):
+        self.set_dwi_data_img()
+        return self.data_frame['dwi_data_img']
+
+    dwi_data_img = property(get_dwi_data_img, set_dwi_data_img)
+
+    def _dti(self, row, mask=None):
+        if not op.exists('dti_params_file'):
```
jyeatman: I'm not sure we want this check. Since this is done at a level higher as well, we should assume at this point that you want to recompute and overwrite. Note that the higher level function had an overwrite option and this check will undo that overwrite option

PR #42, opened 2017-01-09
started working on the cleaning. We added an example of the cleaning …

jyeatman:
…function to the notebook and carved out space in the segmentation module


jyeatman:
Note that the weights seem to be scaled incorrectly. The should correspond roughly to zscores. See the stanford data notebook

Issue #44, opened 2017-02-08
Switch from optparse to argparse


arokem:
See https://github.com/yeatmanlab/AFQ-Browser/pull/100

richford:
I'm working on this now but I have a question: right now some of the options expect a comma-separated list of input files. With argparse (using the `nargs` option), we may use a space separated list of files, e.g.:

`$ pyAFQ_dki -d dwi_file_1 dwi_file_2 -l bval_file_1 bval_file_2 -c bvec_file_1 bvec_file_2`

This seems cleaner to me. It would also allow us to enforce the requirement that the number of dwi files is the same as the number of bval files and bvec files. We could catch that early on dring argument parsing rather than deeper in the call stack. But I don't know if this change would break compatibility with other tools in the AFQ pipeline. And if it does, is that a big deal?

arokem:
@libbyhuber, @jyeatman: would this break something that you are doing? I
think that at the worst you would have to replace commas with spaces in
some of the code that you wrote to use this, but let us know if this is
disruptive.

I am otherwise +1 for this change

jyeatman:
That seems like a sensible change. This is where the simple changes would
need to be made in AFQ and it looks like we even have a test function too!
https://github.com/yeatmanlab/AFQ/tree/master/utilities/pyAFQ

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://brainandeducation.com

richford:
Just to be clear, I think I can reproduce the old behavior with argparse. So the switch isn't mandatory. But space separation with `nargs` seems to be how people are handling lists of input these days so I wanted to take the opportunity to update if it isn't too disruptive. But if it is, there's absolutely no problem with just recreating the old behavior.

arokem:
I think this would be an improvement. The comma-separation thing was a
hack.

It seems that it would entail changing three lines of code in the Matlab
function Jason pointed out, so I really don't think it's too disruptive.

jyeatman:
Go for updating it

We will adjust

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://brainandeducation.com

richford:
[...]

PR #46, opened 2017-02-09
Use argparse instead of optparse

richford:
Resolves issue #44. See discussion in #44 for more details.


arokem:
@jyeatman : how do you feel about this API? Users would have to type `pyAFQ_fit dti file.nii.gz file.bval file.bvec` or something like that. 

I think that the previous API is marginally clearer, because the different options are available to users through the tab-completion (i.e. which models they could fit). 

We could reduce code duplication by putting all the duplicate code in a `cli.py` module in the package portion of the library.

richford:
Good point about the tab completion. FYI, use pyAFQ_fit -h or pyAFQ_fit --help to see the new API.

jyeatman:
I'm fine either way. I don't think it will be difficult to update the
matlab code. I think we should go with whatever you think will be the best
for our long term vision of having many models etc

Jason D. Yeatman, PhD
Assistant Professor, Institute for Learning & Brain Sciences (I-LABS)
Department of Speech & Hearing Sciences
University of Washington
http://brainandeducation.com

arokem:
Let's go with this change then.

TBH, it's not that crucial: the CLIs are mostly meant for Matlab users ;-)

And I think that this change will make things slightly easier to extend
with more models.

arokem:
I think this is now subsumed in #94 that was just merged. I'll close this one.

PR #51, opened 2017-04-20
NF: DKI microstructure

RafaelNH:
Add fit_dkimicro to dki.py and pyAFQ_dkimicro in bin

PR #52, opened 2017-04-20
Try installing dipy from the source on Github.

arokem:
Let's see if we can install straight from master on travis, before merging #51 (might make tests there pass, and we can get rid of "fixes"!)


arokem:
It works! Merging

PR #48, opened 2017-03-06
Brainhack 2017 changes

bcipolli:
* Get `force_recompute` working, as a class-level property
* Tweak `requirements.txt`

PR #54, opened 2017-04-22
Brainhack 2017

arokem:
Mostly about using force_recompute.


arokem:
Supersedes #48

arokem:
I think this should work.

PR #63, opened 2018-04-10
Bvals preallocation

### LINKED ISSUE #56: This could be optimized by preallocating instead of using `append`
https://github.com/yeatmanlab/pyAFQ/blob/daa842b77b4bfd92785dfb3db988444f42b67bfb/AFQ/utils/models.py#L48-L50

richford:
Resolves #56.

`prepare_data` returns the last Nifti1Image generated from `data_files` in order to replicate the pre-existing behavior. But was that intended? It seems that the old behavior was to return the Nifti1Image produced by running `nib.load()` on the last element of `data_files`. Is there something special about the last one?

### DIFF
--- a/AFQ/utils/models.py
+++ b/AFQ/utils/models.py
@@ -52,4 +52,4 @@ def prepare_data(data_files, bval_files, bvec_files, mask=None,
 
     gtab = dpg.gradient_table(bvals, bvecs, b0_threshold=b0_threshold)
 
-    return img, data, gtab, mask
+    return img_array[-1], data, gtab, mask
--- a/AFQ/utils/models.py
+++ b/AFQ/utils/models.py
@@ -40,18 +40,16 @@ def prepare_data(data_files, bval_files, bvec_files, mask=None,
     if isinstance(mask, str):
         mask = nib.load(mask).get_fdata().astype(bool)
 
-    data = []
-    bvals = []
-    bvecs = []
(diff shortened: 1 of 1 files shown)


arokem:
I don't think there is anything special about that one. I think that we assumed that the meta-data will be similar across all the nifti files, and we are keeping this around, so that we can get the affine transformation stored in the header of this file. For example: https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/csd.py#L57

Maybe we should check that the `affine` properties of all the img objects are identical and then return just the affine, instead of the `img` object?

AFQ/utils/models.py
```diff
@@ -40,18 +40,16 @@ def prepare_data(data_files, bval_files, bvec_files, mask=None,
...
-        bvecs.append(np.loadtxt(bvec_file))
-
-    data = np.concatenate(data, -1)
-    gtab = dpg.gradient_table(np.concatenate(bvals),
-                              np.concatenate(bvecs, -1),
-                              b0_threshold=b0_threshold)
-
-    return img, data, gtab, mask
+    img_array = [nib.load(dfile) for dfile in data_files]
+    data_array = [img.get_data() for img in img_array]
```
arokem, line 44:
```suggestion
    data_array = [img.get_fdata() for img in img_array]
```

richford:
Rebased on master.

arokem:
Thanks!

Issue #72, opened 2018-07-11
Could we save these distances in the previous loop through bundles?


arokem:
https://github.com/yeatmanlab/pyAFQ/blob/02c68a2452efc4354e258a75431e7cde9fd95dc3/AFQ/segmentation.py#L303

arokem:
Addressed through #76

PR #76, opened 2018-07-12
Store distances from first loop over bundles for reuse in second loop.

arokem:
Addresses #72.


arokem:
Also tested on a whole-brain dataset. Seems to do the right thing.

arokem:
Interestingly, seems to make things slightly *slower* (about 20%), rather than as intended. I might roll this one back.

PR #83, opened 2018-09-05
Split streamlines

arokem:
Another PR


arokem:
Closing. We need to revisit the issue of splitting streamlines, but we'll do so on a separate PR.

PR #85, opened 2018-09-06
updated tracking params

jyeatman:
step size is by voxel not by mm. 2 steps per voxel is reasonable. We could do more. But probably not needed


jyeatman:
@arokem I added a commented line where I think we should write out the ROIs. I actually think the ROIs might be fine and the visualization is the problem. Once you merge this pull request can you show me how that ROI writing should be done?

arokem:
I think that we need to add scikit-image to the requirements file for this to work. I am going to merge this and then fix that on top of that.

PR #91, opened 2018-09-07
Spinning mosaic

kesshijordan:
first-pass at visualization that saves out spinning frames in the mosaic format @akeshavan needs for her QC webapp

Question: can PIL be a dependency?

TODO: 
- abstract out the frame-saving so it takes any render and produces the stack of png's
- make little QC functions that produce useful renders
- remove temp files when mosaic is saved

### DIFF
--- a/AFQ/viz.py
+++ b/AFQ/viz.py
@@ -319,50 +319,29 @@ def qc_multibundle_ren(bundle_list, showme=False):
     return ren
 
 
-def save_spin_single_bundle(bundle, Nframes=18,
-                            savename='temp', showme=False,
-                            savespin=True, size=(200, 200)):
-
-    ren = window.Renderer()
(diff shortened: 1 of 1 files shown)


kesshijordan:
I just modularized this process so you pass a render to the function that does the snapshot-capture.  @arokem: for the purposes of pyAFQ we don't want to save the individual snapshots, just the mosaic (@akeshavan's QC slides a window over the mosaic... it doesn't use the individual images), but we have to save them as pngs until the mosaic is made. What's the preferred way to delete intermediate files (or I there any reason i should leave them?)?

arokem:
If you generate them in a temp folder, the os will take care of them

arokem:
Sorry. I believe that this was eventually superseded by #229. Closing.

PR #103, opened 2018-09-29
Wip demo notebooks for freesurfer/Recobundles combination feature

kesshijordan:
Hi, @arokem. Here's a set of 2 notebooks on the combined method I've been working on Freesurfer/Recobundles that use the dictionary idea we discussed to make them play nicely with pyAFQ. Let me know if anything isn't clear; when I have a bit more time I can clean them up and put in better annotations.

Cheers,
Kesshi


arokem:
I think this one can be closed now.

PR #105, opened 2018-10-02
Apply an affine preregistration step, prior to SyN.

arokem:


arokem:
I believe this working fine, so I will go ahead and merge this.

PR #113, opened 2019-03-21
S3FS I/O

arokem:
@akeshavan : what do you think?


arokem:
Does anyone understand why travis is not seeing the last letter of `s3fs` in the requirements file? 

https://travis-ci.org/yeatmanlab/pyAFQ/jobs/509619335#L340

So odd.

arokem:
Maybe pinning the versions of the requirements will help with that. 

We'll need to update the nibabel version to the release version on Monday when 2.4 comes out including the fix to streamline reading.

Issue #123, opened 2019-06-28
Error:  mapping -> 'tuple' object has no attribute 'forward'


ejcanalesr:
Hi guys,
First of all, thanks for your work!

I am trying to run the "plot_tract_profile.py" example script using my data.

However, I get this error: 'tuple' object has no attribute 'forward', which is related to the lines:
reg.write_mapping(mapping, hardi_out_dir + '/mapping.nii.gz')
and 
mapping_data = np.array([mapping.forward.T, mapping.backward.T]).T

any idea?

I have other issues, for example, the tract "ARC" is not recognized:

bundle_names = ["SLF", "ILF", "UNC", "ARC"]

...

      7     for hemi in ['_R', '_L']:
      8         bundles[name + hemi] = {
----> 9             'ROIs': [templates[name + '_roi1' + hemi],
     10                      templates[name + '_roi2' + hemi]],
     11             'rules': [True, True],

KeyError: 'ARC_roi1_R'

I checked and despite "ARC_L_prob_map" and "ARC_R_prob_map" are in the templates folder, their corresponding "roi" images are not...

Thanks!
Regards,
Erick

arokem:
Hi Erick, 

Thanks for checking this out!

>However, I get this error: 'tuple' object has no attribute 'forward', which is related to the lines:
reg.write_mapping(mapping, hardi_out_dir + '/mapping.nii.gz')

That's puzzling. Is it really a tuple at that point? Is it possible that there is an errant comma somewhere in your code turning what should be a mapping object into a tuple?  If I understand correctly, you modified our script a bit? Would you mind sharing the entire script you are running?

> I have other issues, for example, the tract "ARC" is not recognized:

This should probably be documented more explicitly and maybe we should call out to our API module here. Arcuate has slightly different rules apply to it. Instead of using its own ROIs, it uses SLF ROIs. See: https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/api.py#L117-L120. Notice that SLF also has an exclusion ROI -- see a bit further down that same function. 

Note that this is still pre-release software, so things are still in flux, but we are working hard towards a release this summer. But we'd love hearing what things you run into as you are running the code -- I believe it should already be usable, but part of the reason that we haven't released yet is that we haven't had enough opportunities to see how it works with different datasets, so your feedback is very helpful.

arokem:
[...]

PR #126, opened 2019-07-03
BF: syn_register_dwi returns a tuple as output.

arokem:
Should resolve some of the issues raised in #123


arokem:
Travis seems happy, and so am I.

PR #128, opened 2019-07-03
Use the DIPY I/O for streamlines, instead of our own.

### LINKED ISSUE #127: Use dipy's I/O instead of our own
So we can avoid warnings like this one: https://travis-ci.org/yeatmanlab/pyAFQ/jobs/553808933#L65

We can use functions from here: https://github.com/nipy/dipy/blob/master/dipy/io/streamline.py to replace `utils.streamlinse.read_trk` and `utils.streamlinse.save_trk`

arokem:
Resolves #127

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -14,6 +14,7 @@ from dipy.segment.mask import median_otsu
 import dipy.data as dpd
 import dipy.tracking.utils as dtu
 import dipy.tracking.streamline as dts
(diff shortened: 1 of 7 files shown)


arokem:
This is fine, but might have some downstream effects. For example, see the change to affine arguments that happened in #129. So, we should keep an eye on that. For now, I'll go ahead and merge this one.

PR #129, opened 2019-07-03
Seed for tracking only in the way-point ROIs.

arokem:
This (@36000's idea!) speeds up the example, because tracking was the
bottle-neck. Next, to get CircleCI to build the examples for us!


arokem:
This seems to work fine. I am going to merge this, so that I can move ahead to setting up CircleCI to run the doc build on PRs.

PR #133, opened 2019-07-08
Batch ODF Calculation

36000:
Batched ODF calculation. Numpy needed ~5GB to hold all projections. Changed what was being calculated in numpy to a for loop. This significantly reduces the amount of memory required to the point where it is unnoticeable compared to other operations. However, it still takes time, so I also added a TQDM progress bar.


36000:
Is there a reason pep8speaks hasn't commented yet?

arokem:
I am not sure why @pep8speaks is silent here. Let's see if the @ mention wakes it to life. 

Saves memory, but this is probably much slower, isn't it?

36000:
I'll run some tests on a computer that has enough RAM so swap space isn't an issue. However, in this example there are only ~360 vertices with 1 vertex per batch, so it won't be that much of a difference. Will this be true for other examples? I could also change the batch size so that the for loop goes 100 times with a batch size of (# of vertices/100).

36000:
Another idea: num_batches should be a parameter. It could default to 100 batches. -1 could mean one batch for every vertex. If num_batches != 1, then we use the progress bar!

36000:
With num_batches as 100:      35.48 seconds
With num_batches as 1:          31.46 seconds

So on a computer with enough memory, its certainly faster. But on a computer without, it should be an option to batch.

arokem:
Looks like the CI finished OK, so once you fix up PEP8, we can merge this.

36000:
If these checks pass, it should be ready to merge!

arokem:
Yup! Merging.

PR #134, opened 2019-07-08
Adds an example using RecoBundles to do the segmentation.

arokem:
It's a bit long and windy right now, but at least some of that is because I had to figure out how to use the atlas to seed in the location of the bundles in the subject space. Neverthless, useful. After more testing, we should proceed to bake this stuff into the API.


arokem:
I'm going to go ahead and merge this in and move on ahead with it.

PR #136, opened 2019-08-01
Segmentation Refactoring and Other Updates

36000:
In the process of putting segmentation into its own class. Added logging, option to resample. trying numba.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -398,7 +398,7 @@ def _recobundles(row, wm_labels, bundle_dict, reg_template, odf_model="DTI",
                                         force_recompute=force_recompute)
         tg = nib.streamlines.load(streamlines_file).tractogram
         sl = tg.apply_affine(np.linalg.inv(row['dwi_affine'])).streamlines
-        segmentation = seg.Segment(method='Reco')
+        segmentation = seg.Segmentation(method='Reco')
         bundles = segmentation.segment(bundle_dict, sl)
         tgram = aus.bundles_to_tgram(bundles, bundle_dict, row['dwi_affine'])
         nib.streamlines.save(tgram, bundles_file)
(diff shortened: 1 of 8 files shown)


36000:
@arokem Take a look at the current Segmentation class / file. This should be ready to merge.

AFQ/segmentation.py
```diff
@@ -125,165 +86,257 @@ def gaussian_weights(bundle, n_points=100, return_mahalnobis=False,
...
+
+        Parameters
+        ----------
+        split : boolean
+            If true, classify the streamlines and split those that:
+            1) cross the midline, and 2) pass under 10 mm below
+            the mid-point of their representation in the template space.
+            Default: False
+        nb_points : int
+            Resample streamlines to nb_points number of points.
```
arokem: Docstring would have to change slightly

AFQ/segmentation.py
```diff
@@ -125,165 +86,257 @@ def gaussian_weights(bundle, n_points=100, return_mahalnobis=False,
...
+            If streamlines is None, will use previously given streamlines.
+            Default: None.
+        """
+        if streamlines is None:
+            streamlines = self.streamlines
+
+        self.streamlines = np.array(
+            dps.set_number_of_points(streamlines, nb_points))
+
+    def split_streamlines(self, streamlines=None,
```
arokem: We might need to rename this function, as it doesn't split streamlines (for now)

AFQ/api.py
```diff
@@ -397,7 +398,8 @@ def _recobundles(row, wm_labels, bundle_dict, reg_template, odf_model="DTI",
                                         force_recompute=force_recompute)
         tg = nib.streamlines.load(streamlines_file).tractogram
         sl = tg.apply_affine(np.linalg.inv(row['dwi_affine'])).streamlines
-        bundles = seg.recobundles(sl, bundle_dict)
+        segmentation = seg.Segmentation(method='Reco')
```
arokem: We should refactor the API here so that it uses the Segment API directly. That is, the bundles file should call out to segmentation.segment. 

For this, we need to first make sure that the API for the two segmentation functions (`'reco'` and `'ROI'`) are identical, which they are not currently. 

The idea is that the object knows which function it is going to use, but the calling process is agnostic to that, so that the same API can be reused, and we only need one function at this level.

36000:
@arokem , the only thing this needs is a reco API test. I won't be able to finish it today, but I think I resolved all the other comments if you want to look at them before you leave.

arokem:
Yeah - this all looks good. Let's go ahead and merge this and then add an issue to remind us to go back  and add that test.

PR #137, opened 2019-08-01
Starting to adapt to DIPY 1.0 API.

arokem:


arokem:
This now passes locally on my machine (against current DIPY master branch), so it should pass on the CI, as soon as DIPY 1.0 is released.

36000:
The examples need to be adapted as well. For example, move_streamlines in plot_tract_profile should be transform_tracking_output I think.

arokem:
Maybe we can take that on a separate branch, once everything is passing here?

arokem:
OK - this passes the tests now, with dipy 1.0 released.

@36000 : would it be OK to merge this into master now? I believe that you could then keep going on your work in #136. 

And we can come back to fix the examples for the new API.

36000:
It is OK to merge!

PR #139, opened 2019-08-23
Early track checks

36000:
Move check for streamline length into tracker, so too small or too large streamlines can be thrown out without having to be saved.


AFQ/_fixes.py
```diff
@@ -67,7 +72,12 @@ def _generate_streamlines(self):
                 else:
                     parts = (B[stepsB - 1:0:-1], F[:stepsF])
                     streamline = np.concatenate(parts, axis=0)
-                yield streamline
+
+                len_sl = len(streamline)
+                if len_sl < self.min_length or len_sl > self.max_length:
```
arokem: Can we make this take into account the step size, so that the length is in mm, rather in the number of steps?

arokem:
Looks great! In the long term this is something we should push upstream to DIPY itself (here: https://github.com/nipy/dipy/blob/master/dipy/tracking/local_tracking.py). For now, I'll merge this here.

PR #143, opened 2019-09-13
Trying out Github Actions

arokem:
Just trying to see what this looks and feels like


arokem:
This now seems to work on my fork: https://github.com/arokem/pyAFQ/pull/4. Maybe we should switch over from Travis?? I'll try testing #160 with this included, to see whether it solves the issues we saw there.

PR #150, opened 2019-09-28
Recobundles HCP

arokem:
Example of running recobundles segmentation with HCP data. 

Also, miscellaneous small fixes.


arokem:
I think that this is ready for a review now. Anyone want to take a look or try this out?

arokem:
This is ready to go now!

PR #152, opened 2019-10-14
Towards tracking/segmentation with dask parallelization

arokem:
For now, mostly focused on saving memory in dask workers for now


arokem:
There's nothing really interesting about this. We should revisit at some point, but this is not useful.

PR #154, opened 2019-10-15
Allow image affine to be calculated in advance for afq segmentation

36000:


36000:
I believe the failure doesn't have to do with the changes I just made? Only failed in python 3.6

arokem:
This looks good.  The Travis error was an OOM error, so hopefully restarting the build will be enough to resolve that.

AFQ/segmentation.py
```diff
@@ -219,6 +222,7 @@ def _seg_afq(self, bundle_dict, streamlines, fdata=None, fbval=None,
         Quantification" PloS One 7 (11): e49790.
         """
         self.logger.info("Preparing Segmentation Parameters...")
+        self.img_affine = img_affine
```
arokem: Maybe we need to add some error handling: if users provide both `img_affine` and fdata, that could result in conflicting information.  So maybe users should  be allowed to provide one, or the other, but not both?

AFQ/segmentation.py
```diff
@@ -218,7 +221,18 @@ def _seg_afq(self, bundle_dict, streamlines, fdata=None, fbval=None,
         "Tract Profiles of White Matter Properties: Automating Fiber-Tract
         Quantification" PloS One 7 (11): e49790.
         """
+        if img_affine is not None:
+            if (mapping is None or
+                fdata is not None or
+                fbval is not None or
+                    fbvec is not None):
```
arokem: I think that pep8 is upset about those `or`s being at the end of lines. See: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

PR #159, opened 2019-10-18
Clips streamlines by the inclusion ROI.

arokem:
Address #157.

This is in the right direction, but requires more validation.

Issue #81, opened 2018-09-04
Clip to plane ROIs 


arokem:
To get only the central part of the streamline in each individual. See https://github.com/nipy/dipy/pull/1617. 

Alternatively, think of a way to align the nodes across individuals (@jyeatman has ideas how to do this, I believe).

kesshijordan:
Regarding aligning nodes across individuals: One option is doing streamline linear registration to a bundle template in MNI space, and another is just locking it to the anatomy in MNI space. We just have to be careful in thinking about bundles that may vary in where they stop (e.g. the arcuate projects to the cortex on many adjacent gyri and not every case will extend the same distance into the frontal lobe). Simple sub-sampling in the patient's bundle is fine for structures that tend to terminate in a consistent manner (e.g CST, IFOF), but SLF II, III, and Arcuate are bundles for which that method would result in a mismatch between points, anatomically, that may vary considerably depending on parameters/data/age of the subjects. @jyeatman and I discussed approaching the trunk in a tbss-esque analysis, and the periphery using a combination of values and NaN's (the NaN's giving information about how extensively the model is able to track in the periphery; e.g. an arcuate that doesn't extend as far into the frontal lobe in some subjects may indicate low FA, but that depends on the exact model/tracking parameters).

36000:
Clipping is done in #159 , discussion on realignment can move to #379

PR #160, opened 2019-10-22
Adding TQDM to Segmentation

36000:


AFQ/segmentation.py
```diff
@@ -3,6 +3,7 @@
 from scipy.spatial.distance import mahalanobis, cdist
 
 import nibabel as nib
+from tqdm import tqdm
```
richford: If you use
```
from tqdm.auto import tqdm
```
then tqdm will automatically detect the backend and choose between console and notebook versions so that the progress bar is pretty in all cases.

arokem: Ooh! Good to know!

arokem:
Looks like that memory error was not a fluke. But maybe after you make a few more changes (i.e., per-streamline tqdm) it will resolve itself?

arokem:
This is puzzling. It's hard to believe that tqdm adds that much memory load. I am going to try to push another commit on the  branch I am currently working on (#81) and we can see whether this something  more fundamental (e.g., one of our dependencies has made a release?)

arokem:
BTW, one of the potentially memory-saving things I did in that PR is this: https://github.com/yeatmanlab/pyAFQ/pull/159/files#diff-42ac59cc95334c7a71c2df1698dfeb81R446. Maybe if you applied that change here, we would be able to squeak past this memory bottleneck?

36000:
I find it very puzzling as well. I am going to push a commit with no tqdm at (no changes from master) to see if it passes. I am also trying to profile memory around where it hits the memory error.

36000:
Ok, with no changes, one succeeded and one failed. so it just looks like we are approaching the limits of memory when testing the ROI exporting.

arokem:
Could you please rebase this on master? Let's see if the errors magically go away after the recent changes.

arokem:
Still no joy here, huh? I wonder whether there is something we could do to make these tests less memory demanding. It should also help make them run faster.

arokem:
Looks like Github Actions gives us enough memory to get past this bottleneck. 

This PR: https://github.com/arokem/pyAFQ/pull/7 get's tested both on Travis (fails, same memory error as here) and on Github Actions (passing, of course). 

One potential  way forward here would be to deprecate Travis testing and just using Github Actions. Does anyone have any thoughts on this? I am not 100% sure what the pros/cons are. I've heard some people grumble about MS consolidating power by taking out the third-party vendors (such as Travis), but maybe we're OK with that, if it means Github gets better integrated?

arokem:
Superseded by #178

PR #166, opened 2019-11-06
WIP: Using probability threshold

arokem:
Found a way to  speed up our main loop along the way. Still needs some cleanup, before it can be used, but I am putting up this PR, because I also ran into  some CI issues along the way, and I would like to  resolve those.


arokem:
Superseded by #169. Closing.

PR #167, opened 2019-11-12
Let cross streamlines only require the affine

36000:


arokem:
This is good. I think that the CI failure is unrelated. I am also seeing it on #166. Based on the errors, I think it has to do with versions of dependencies that we need to sort out. But I am not sure why this only happens with the CLIs.

arokem:
I believe that if you rebase on master, the test errors should go away.

arokem:
Yep. Here we go!

Issue #164, opened 2019-11-03
Move transformations  into  the segmentation function


arokem:
This:  https://github.com/yeatmanlab/pyAFQ/blob/master/examples/plot_tract_profile.py#L100

And this:  https://github.com/yeatmanlab/pyAFQ/blob/master/examples/plot_tract_profile.py#L125

Should probably be done inside of the segmentation function, so that we get an API consistent with the RB API.

PR #175, opened 2019-11-26
Moved transformations inside of segmentation for AFQ

36000:
Move transformations into the segmentation function #164


36000:
@arokem 
This commit causes this error on my local machine (the tests online should have the same error, I just don't want to wait for them to finish):
```
        # Test bundles exporting:
        myafq.export_bundles()
        assert op.exists(op.join(myafq.data_frame['results_dir'][0],
                                 'bundles',
                                 'CST_R.trk'))

>       assert tract_profiles.shape == (800, 5)
E       assert (1000, 5) == (800, 5)
E         At index 0 diff: 1000 != 800
E         Use -v to get the full diff

AFQ/tests/test_api.py:148: AssertionError
```
Can you explain why tract profiles would have shapes of either 800 or 1000?

arokem:
Regarding 800 vs. 1000, this is hard-coded based on 100 nodes per tract, two scalars (MD and FA) for each of the tracts segmented in this case.

arokem:
After some mulling, it seems that we have come to the conclusion that to make sure that the results of recobundles and waypoint segmentation are the same, we need to provide inputs in the RASMM space (which is the output of our tracking) and then have both of these provide outputs in the VOX space. This means that we would apply the following transformation:

```streamlines = dts.Streamlines(
    dtu.transform_tracking_output(streamlines,
                                  np.linalg.inv(img.affine)))```

On the input to segmentation by waypoint ROI, before any additional computations and on the output of recobundles, right before returning it. 

This  way, we can forgo additional transformations and save things to the VOX space, regardless of the algorithm.  See: https://github.com/yeatmanlab/pyAFQ/pull/186

arokem:
@36000 : I think that this will all be much simplified once we merge #190. Let's look at that together later today.

36000:
I cannot understand why API is failing. It finds ARC_L, CST_L, CST_R, FP but also SLF_R
	The only difference is the use of transform tracking output instead of apply_affine? Underlying code looks similar:
https://github.com/nipy/nibabel/blob/6cf363d9b7488e2e3069ddfd6f4a0fefbfb2b577/nibabel/affines.py#L76
https://github.com/nipy/dipy/blob/64641b26bd4791ad7c0e8afc6806911292a38232/dipy/tracking/utils.py#L810

arokem:
This is largely superseded by #190. Closing.

Issue #174, opened 2019-11-25
pyAFQ template


Qji12:
Hi 
Thanks for resolving issue# 172. I found that in ./AFQ_data, the roi files and prob maps are not consistent: prob [197 233 189], roi [182 218 182], MNI template [197 233 189]. Is that a problem when you trying to warp roi? Thanks

arokem:
It might be. I am still looking into it. I think that https://github.com/yeatmanlab/pyAFQ/pull/183/commits/b4f1a7a493d6204cded47235965d978f45c9e672 might resolve this.

arokem:
I believe this is resolved in #183. Let me know if you run into any issues.

Issue #168, opened 2019-11-15
Add selection based on cortical endpoints


arokem:
See `AFQ_DefineFgEndpoints` in the  Matlab implementation

jyeatman:
The simplest thing to do here would be to use the exact same template as we do in the matlab version. Then we could have an option to use freesurfer instead

arokem:
Yes. That's what I was thinking of doing as a first pass.

Based on [this line](https://github.com/yeatmanlab/AFQ/blob/50bdc81e060670a83789a4168f9b5e88cea39c2a/functions/AFQ_DefineFgEndpoints.m#L98) it should be this file:  https://github.com/yeatmanlab/AFQ/blob/50bdc81e060670a83789a4168f9b5e88cea39c2a/templates/labelMaps/MNI_AAL_AndMore.nii.gz, correct? What's the difference between that one and this one: https://github.com/yeatmanlab/AFQ/blob/50bdc81e060670a83789a4168f9b5e88cea39c2a/templates/labelMaps/MNI_AAL.nii.gz?

jyeatman:
I think that is correct. But I don't know what the difference is.Beyond the fact that one has AAL And More!

arokem:
More definitely sounds better! Let me see if I can at least have a poke at this code, so that I might have some informed questions to ask on Monday's call.

arokem:
@jyeatman : question about the implementation. [This section of code](https://github.com/yeatmanlab/AFQ/blob/50bdc81e060670a83789a4168f9b5e88cea39c2a/functions/AFQ_DefineFgEndpoints.m#L54-L63) suggests that enpoint location for each bundle is determined based on an ordering within a list. For example, based on the  strings stored here, it looks like number 3 and 4 are the corticospinal tracts. Where is the list with the ordering of all 20 stored? Is it correct to assume that [this](https://github.com/yeatmanlab/AFQ/blob/50bdc81e060670a83789a4168f9b5e88cea39c2a/functions/AFQ_SegmentFiberGroups.m#L199-L208) is the ordering used here as well?

jyeatman:
Not sure I fully understand. But, the default behavior of afq is to segment 20 fiber groups. This segmentation is always done in the same order. Then it looks up the endpoint regions corresponding to each fiber group. The ordering of fiber groups is defined in the function you note.
1. Left Thalamic Radiation
2. Right Thalamic Radiation
3. Left Corticospinal
4. Right Corticospinal
[...]

PR #183, opened 2019-12-03
Use the AAL atlas to filter streamlines

arokem:


AFQ/segmentation.py
```diff
@@ -42,7 +43,9 @@ def __init__(self,
                  b0_threshold=0,
                  prob_threshold=0,
                  rng=None,
-                 return_idx=False):
+                 return_idx=False,
+                 filter_by_endpoints=True,
+                 dist_to_aal=4):
```
arokem: @jyeatman : in your experience, what is a reasonable distance to use for filtering? Using this distance still seems to leave quite a few streamlines that look spurious (at least for CST, at least in the example dataset).

jyeatman: I would have suggested 4mm just looking at fiber endpoints and atlas voxels. The atlas is very rough and just serves to clear up the really bad streamlines

arokem: Great - that helped - calibrating to the voxel size, so the distance is in mm, solved the particular issue I was seeing with that CST, so 4 mm does seem to be a good way to go (but 8 mm is too much 😄)

arokem:
OK - I've been playing around with this all day yesterday, and I'm pretty sure that it's doing the right thing. 

In particular, I think that this resolves both #168, as well as #174 (resolved in https://github.com/yeatmanlab/pyAFQ/commit/b4f1a7a493d6204cded47235965d978f45c9e672). So I  think that this is ready to merge.

PR #184, opened 2019-12-03
Add fiber group class

36000:


36000:
Move to own module (bundles.py); have everything be stateful tractograms that keep track of space; find a convenient way to specify file paths; use bundles everywhere.

36000:
Add tests

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+            see DIPY.
+            Default: Origin.NIFTI
+
+        space : string, optional.
+            see DIPY.
+            Default: Space.RASMM
+
+        bundles : dict, optional.
+            Keys are names of the bundles.
+            If using_idx is False, values are Streamline objects.
```
arokem, line 40:
```suggestion
            If using_idx is False, values are StatefulTractogram objects.
```

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+            Else, values are dictionaries with Streamline objects and indices.
```
arokem, line 41:
```suggestion
            Else, values are dictionaries with StatefulTractogram objects and indices.
```

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+        sls = dtu.transform_tracking_output(sft.streamlines, affine)
+        return StatefulTractogram(sls,
+                                  reference,
+                                  self.space,
+                                  origin=origin,
+                                  data_per_streamline=sft.data_per_streamline)
+
+    def apply_affine(self, affine, reference, origin=Origin.NIFTI):
+        """
+        Appliy a linear transformation, given by affine, to all
```
arokem, line 161:
```suggestion
        Apply a linear transformation, given by affine, to all
```

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+            Whether to verify that the bounding box is valid in voxel space.
+            Default: False
+        """
+        if space is not None:
+            space_temp = self.space
+            self.to_space(space)
+
+        for bundle_name, bundle in self.bundles.items():
+            save_tractogram(bundle,
+                            f'{file_path}{bundle_name}{file_suffix}',
```
arokem, line 226: I would use `os.path.join` here, instead of this.

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+            Default: np.eye(4)
+        bbox_valid_check : boolean, optional.
+            Whether to verify that the bounding box is valid in voxel space.
+            Default: False
+        """
+
+        for bundle_name in bundle_names:
+            if self.reference == 'same':
+                sft = load_tractogram(
+                    f'{file_path}{bundle_name}{file_suffix}',
```
arokem, line 260: Same here

AFQ/bundles.py
```diff
@@ -0,0 +1,325 @@
...
+            Default: 'FA'
+
+        n_points : int
+            Number of points to resample to.
+            Default: 100
+        """
+        self.to_space(Space.VOX)
+        profiles = []
+        for bundle_name, bundle in self.bundles.items():
+            weights = gaussian_weights(bundle.streamlines, n_points=n_points)
```
arokem, line 313: We might want to make the weighting optional, as a boolean key-word argument.

AFQ/bundles.py
```diff
@@ -0,0 +1,338 @@
...
+            Whether to verify that the bounding box is valid in voxel space.
+            Default: False
+        """
+        if space is not None:
+            space_temp = self.space
+            self.to_space(space)
+
+        for bundle_name, bundle in self.bundles.items():
+            save_tractogram(bundle,
+                            os.path.join(file_path, bundle_name, file_suffix),
```
arokem, line 229: This wouldn't work, because the result would be something like: 

`'./path/to/bundle_name/.trk'`

```suggestion
                            os.path.join(file_path, f"{bundle_name}{file_suffix}"),
```

AFQ/bundles.py
```diff
@@ -0,0 +1,338 @@
...
+            Default: '.trk'
+        affine : array_like (4, 4), optional.
+            The mapping from the file's reference to this object's reference.
+            Default: np.eye(4)
+        bbox_valid_check : boolean, optional.
+            Whether to verify that the bounding box is valid in voxel space.
+            Default: False
+        """
+
+        full_path = os.path.join(file_path, bundle_name, file_suffix)
```
arokem, line 260: And here as well.

PR #186, opened 2019-12-03
Save the results in the VOX space

arokem:


arokem:
On further thought, let's stick to RASMM for now, as this is the coordinate frame that tracking produces.

PR #190, opened 2019-12-11
Transform streamlines using SFT methods.

arokem:
We should use StatefulTractogram methods for spatial transformations wherever possible (instead of `transform_tracking_output`).


arokem:
Also simplifies `patch_up_roi` to remove Gaussian smoothing.

arokem:
I'll go ahead and merge this now, so we can keep going. Still need to make sure the RecoBundles example works with these changes.

PR #192, opened 2019-12-18
plot_recobundles example with SFT

arokem:


arokem:
OK. This is working as expected.

PR #193, opened 2019-12-19
Use a config file as input to CLI. Simplify default setting.

arokem:
Along the way, I think that this would be the place to implement saving of metadata files to go along with the generated derivative data.


arokem:
TODO:

- [ ] Metadata for each derived file
- [ ] Test for new CLI 
- [ ] Separate segmentation outputs by algorithm
- [ ] Documentation of the CLI

arokem:
More TODO: 

- [ ] Make sure that BIDSish names have only one `desc` field.

arokem:
OK. I believe that this now does everything I would like it to. Off we go!

PR #195, opened 2019-12-26
More work on docs etc. for release

arokem:


arokem:
Turns out it was that simple. Every push on a PR now builds the documentation, which can then be downloaded from the artifacts tab and inspected.

arokem:
Once this comes back green, I will merge it. Then, need to confirm that the pypi upload workflow does the right thing (presumably uploads a version to test.pypi.org upon merge and in every push) and I think that we will be ready for a 0.1 tag/release.

PR #199, opened 2020-01-02
Use only the setuptools_scm version string. Removing all others.

arokem:


arokem:
I believe the documentation build failure is spurious. I am going to go ahead and merge this, in the hopes that it resolves by itself, when on master.

PR #204, opened 2020-01-16
`clean_by_endpoints` should be able to return the indices.

arokem:
For now, just the failing unit test. Implementation to follow.


arokem:
OK. This is now ready to go. Once merged and integrated, I will push out a new release.

PR #206, opened 2020-01-22
CLI fixes and creation of a Boutiques descriptor

gkiar:
Hi!

I've created a [boutiques descriptor](https://boutiques.github.io) for pyAFQ, and would like your help towards finishing it up! Having a descriptor for a tool like this is nice because it allows us to do pre-runtime validation of arguments/config settings, output validation, output querying, and sweeping across parameter settings + logging of performance (when paired with [Clowdr](https://github.com/clowdr/clowdr/tree/develop)).

The current version of the descriptor has all of the options present within the configuration file specified with default values corresponding to the template. It requires two arguments: the `dmriprep_directory` and `config_filename` (which Boutiques will write from the descriptor). To test this, you can do it either in two independent steps or combine these into a single step, both shown below:

```bash
$ # First, install boutiques
$ pip install --upgrade boutiques 

$ # Option 1: simulate the Boutiques "invocation", save it, and use it to generate a config file
$ bosh example pyafq_cli.json > invocation.json
$ # You can look in the file created above and see a JSON object that looks something like this:
$ # {
$ #    "config_file": "str_config_file_sJ",
$ #    "dmriprep_path": "str_dmriprep_path_wI"
$ # }
$ # Then, you can provide this to the simulate command...
$ bosh exec simulate pyafq_cli.json -i invocation.json

$ # Option 2: Combining the above on the command line (removing spaces in the produced JSON)
$ bosh exec simulate pyafq_cli.json -i $(bosh example pyafq_cli.json | tr -d [:space:])
```

After performing either of the above, you'll see a `.toml` file produced in your current directory which contains the generated configuration. You can provide additional arguments (overriding defaults) in your invocation file as `key-value` pairs using the "id" of the field you wish to override. Once you're happy with your config, this tool can be run with:

```bash
bosh exec launch pyafq_cli.json invocation.json  # ... with whatever other args you want, like volume mounts
```

### What I need from you
- [x] Validation of input types between enums and lists (in particular, for `bundles` and `wm_labels`, it is unclear if these are strict enums or lists that are extensible and in fact correspond to values stored within a parcellation file. I suspect strict enum for the former and extensible list for the latter?)
- [x] Validation of the options available to `stats`,  `bundles_seg_algo`, `scalars_model`, and other enumerated fields ensuring that all valid options are included
- [x] (optional) Description of the outputs which will be produced
- [x] (optional) A description for each of the input values
- [x] (optional) For increased consistency in the schema, replacing the `None` default values, for ids such as `sphere`, to a string such as `"None"`. The reason for this is that the default value and provided arguments are currently of different types, a reserved word `None` which does not need quotes, or a string which does require quotes. This makes validation/substitution of this parameter tricky. I think this should be a relatively trivial change in your parser, and can happily make the PR myself if you're willing to do this change.


arokem:
Thanks Greg! My schedule today is packed, but I will try to take a look
tomorrow.

arokem:
OK - I've taken a look. This looks very promising -- I've never used Boutiques before, so I am still learning. Let me see if I can answer your questions: 

> Validation of input types between enums and lists (in particular, for bundles and wm_labels, it is unclear if these are strict enums or lists that are extensible and in fact correspond to values stored within a parcellation file. I suspect strict enum for the former and extensible list for the latter?)

IIUC, you are correct. The bundles are selected from a rather strictly defined set. Would it be helpful if  the possible values of these strings were stored somewhere? 

For wm_labels, it's a bit more flexible. It depends on what the user provides as a white matter segmentation. But it has to be a list of numbers. The values provided as defaults are the integers that identify wm in the FS segmentation, which is detected as the default segmentation.

>Validation of the options available to `stats`, `bundles_seg_algo`, `scalars_model`, and other enumerated fields ensuring that all valid options are included

Do you mean `stat`? The statistic used in the cleaning algorithm? That's left a bit flexible here, but maybe we should restrict it to be one of `'median'` or `'mean'`. It really doesn't make much sense to  use any other statistic to detect the central tendency of the  coordinates of streamlines within a putative bundle. the only options for `bundles_seg_algo` are "AFQ" and "Reco" (and  that should stay like that for the foreseeable future). `scalars_model` is just "DTI" and "DKI" for now (and "DKI" is not really fully supported for now). In the long run, we'd like to also support inputs like ["DTI", "DKI"], where more than one model is fit, but we don't yet support that. Any others I should look at?

> (optional) Description of the outputs which will be produced

If you run this on one subject's data you should see everything that is produced. It's elaborate :-)

> (optional) A description for each of the input values

For now, these things are hidden in the docstrings of the various called functions, but I agree that would be good to have in the documentation.

arokem:
Sorry: missed that last one.

> (optional) For increased consistency in the schema, replacing the None default values, for ids such as sphere, to a string such as "None". The reason for this is that the default value and provided arguments are currently of different types, a reserved word None which does not need quotes, or a string which does require quotes. This makes validation/substitution of this parameter tricky. I think this should be a relatively trivial change in your parser, and can happily make the PR myself if you're willing to do this change.

Yeah - I think we can do that. Let me try a PR with that.

arokem:
Did that answer all your questions?

gkiar:
Hey @arokem - thanks for the all the replies! You did indeed hit every question, and I've got just a few follow-ups... I'll post some here and some in line with code chunks to make it more clear.

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+    },
+    {
+        "id": "bundles",
+        "name": "bundles",
+        "type": "String",
+        "list": true,
+        "min-list-entries": 1,
+        "list-separator": "', '",
+        "value-key": "[BUNDLES]",
+        "value-choices": ["ATR", "CGC", "CST", "HCC", "IFO", "ILF",
```
gkiar: > IIUC, you are correct. The bundles are selected from a rather strictly defined set. Would it be helpful if the possible values of these strings were stored somewhere?
I currently have the set of all possible bundles set here, which is also the default value. If there are other bundles not in the default we can just add them to this list in `value-choices` but not `default-value`.

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "id": "wm_labels",
+        "name": "wm labels",
+        "type": "Number",
+        "integer": true,
+        "list": true,
+        "min-list-entries": 1,
+        "list-separator": ", ",
+        "value-key": "[WM_LABELS]",
+        "value-choices": [250, 251, 252, 253, 254, 255, 41, 2, 16, 77],
+        "default-value": [250, 251, 252, 253, 254, 255, 41, 2, 16, 77],
```
gkiar: > For wm_labels, it's a bit more flexible. It depends on what the user provides as a white matter segmentation. But it has to be a list of numbers. The values provided as defaults are the integers that identify wm in the FS segmentation, which is detected as the default segmentation.

@arokem ok, so this should still have `"list": true`, but no `"value-choices"` set (meaning any value that is an integer will be accepted), and the `"default-value"` set as-is, with a description added to state that this is assuming the `aparc+aseg` label set from Freesurfer. Agreed?

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "value-key": "[MIN_SL]",
+        "default-value": 20,
+        "optional": true
+    },
+    {
+        "id": "stat",
+        "name": "stat",
+        "type": "String",
+        "value-key": "[STAT]",
+        "value-choices": ["mean", "median"],
```
gkiar: > Do you mean stat? The statistic used in the cleaning algorithm? That's left a bit flexible here, but maybe we should restrict it to be one of 'median' or 'mean'. It really doesn't make much sense to use any other statistic to detect the central tendency of the coordinates of streamlines within a putative bundle.

@arokem I did indeed mean stat -- oops! ok, leaving as `mean` and `median` then. 👍

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "value-choices": ["false", "true"],
+        "default-value": "false",
+        "optional": true
+    },
+    {
+        "id": "segmentation_seg_algo",
+        "name": "segmentation seg algo",
+        "type": "String",
+        "value-key": "[SEGMENTATION_SEG_ALGO]",
+        "value-choices": ["AFQ"],
```
gkiar: Similar to `bundles_seg_algo`, does `segmentation_seg_algo` only have `AFQ` and `Reco` as the valid options?

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "value-key": "[MAX_LENGTH]",
+        "default-value": 1000,
+        "optional": true
+    },
+    {
+        "id": "odf_model",
+        "name": "odf model",
+        "type": "String",
+        "value-key": "[ODF_MODEL]",
+        "value-choices": ["DTI"],
```
gkiar: Similar to `scalars_model`, are the options here `DTI` and `DKI`?

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "value-choices": ["dti_fa", "dti_md"],
+        "default-value": ["dti_fa", "dti_md"],
+        "optional": true
+    },
+    {
+        "id": "directions",
+        "name": "directions",
+        "type": "String",
+        "value-key": "[DIRECTIONS]",
+        "value-choices": ["det", "prob"],
```
gkiar: Am I correct in assuming that the options for `directions` are `det` and `prob`?

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "value-choices": ["AFQ"],
+        "default-value": "AFQ",
+        "optional": true
+    },
+    {
+        "id": "scalars_model",
+        "name": "scalars model",
+        "type": "String",
+        "value-key": "[SCALARS_MODEL]",
+        "value-choices": ["DTI"],
```
gkiar:
```suggestion
        "value-choices": ["DTI", "DKI"],
```

per:

>scalars_model is just "DTI" and "DKI" for now (and "DKI" is not really fully supported for now). In the long run, we'd like to also support inputs like ["DTI", "DKI"], where more than one model is fit, but we don't yet support that.

Eventually, we can turn this into a list so that the behaviour you'd like is possible

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "default-value": ["ATR", "CGC", "CST", "HCC", "IFO", "ILF",
+                          "SLF", "ARC", "UNC", "FA", "FP"],
+        "optional": true
+    },
+    {
+        "id": "bundles_seg_algo",
+        "name": "bubndles seg algo",
+        "type": "String",
+        "value-key": "[BUNDLES_SEG_ALGO]",
+        "value-choices": ["AFQ"],
```
gkiar:
```suggestion
        "value-choices": ["AFQ", "Reco"],
```

per: 

>the only options for bundles_seg_algo are "AFQ" and "Reco" (and that should stay like that for the foreseeable future).

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
...
+        "default-value": "det",
+        "optional": true
+    },
+    {
+        "id": "max_angle",
+        "name": "max angle",
+        "type": "Number",
+        "value-key": "[MAX_ANGLE]",
+        "minimum": 0.0,
+        "maximum": 180.0,
```
gkiar: Are `[0.0, 180.0]` correct extreme values for the angle?

gkiar:
OK @arokem - I've added comments on the descriptor addressing each of your responses. I can also go and make a PR with the suggestion in the last bullet, since it shouldn't take long, and will do so in a different branch so we can look at them independently.

pyafq_cli.json
```diff
@@ -0,0 +1,501 @@
+{
+    "name": "pyAFQ",
+    "tool-version": "0.1.1",
+    "description": "Automated Fiber Quantification ... in Python",
+    "command-line": "pyAFQ [CONFIG_FILE]",
+    "schema-version": "0.5",
+    "container-image": {
```
gkiar: ***ToDo @gkiar***: add `environment-variables` field here to make explicit the `fuzzy` parameters.

gkiar:
@arokem : Looking a little closer at the parser, it seems that options set as `None` (without quotes, as is exactly specified in the default config from the help-text), crashes my version of `pyAFQ` with the following issue:

```
Traceback (most recent call last):
  File "/Users/gkiar/code/env/pyafq/bin/pyAFQ", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/gkiar/code/gkiar/pyafq/bin/pyAFQ", line 100, in <module>
    config = toml.load(opts.config)
  File "/Users/gkiar/code/env/pyafq/lib/python3.7/site-packages/toml/decoder.py", line 112, in load
    return loads(ffile.read(), _dict, decoder)
  File "/Users/gkiar/code/env/pyafq/lib/python3.7/site-packages/toml/decoder.py", line 458, in loads
    raise TomlDecodeError(str(err), original, pos)
toml.decoder.TomlDecodeError: This float doesn't have a leading digit
```

So maybe it needs to be in quotes, anyways?

gkiar:
Hey @arokem - there are a few more notes here that would be great to have your feedback on before I finalize the remaining details. Thanks! :)

arokem:
Sorry for the slowness. This will unfortunately take another couple of days.

gkiar:
A couple more questions about the tool in general that may make sense to resolve prior to merging in the descriptor:
- Is there an option to specify which subjects or sessions to process, so that an entire dataset isn't processed en masse if the user wants to parallelize across subjects? Naming these `participant_label` and `session_label` would be consistent with the existing BIDS app spec.
- Is there a way to specify an output directory that is not the same as the input directory? Alternatively, just adding a suffix for files produced by the run (i.e. with a parcellation name, date, or in my case a perturbation run ID, etc.)?
- Is there an option to either use or ignore existing derivatives found inside the working directory? Something like `no_cache = True|False`, where pieces are recycled if existing (I think this is already there) but that can be turned off if we want a clean slate.

pyafq_cli.json
arokem: This seems fine for now. 

It does get a bit complicated: First, we might have additional bundles in the future. Second, the Recobundles segmentation algorithm uses a different list as its default. How do you think we should handle this?

pyafq_cli.json
arokem: Yes - that's correct!

pyafq_cli.json
arokem: I believe so.

pyafq_cli.json
arokem: Confusingly, it's "DTI" and "CSD" for this one.

pyafq_cli.json
arokem: Yes.

pyafq_cli.json
arokem: Sounds good!

pyafq_cli.json
gkiar: Is the list overlapping with this one? Perhaps we could merge the two lists in the `value-choices` field, just to cover all the bases we currently have, and then leave the `default` as-is. Thoughts?

pyafq_cli.json
```diff
@@ -0,0 +1,507 @@
...
+        "value-key": "[MAX_LENGTH]",
+        "default-value": 1000,
+        "optional": true
+    },
+    {
+        "id": "odf_model",
+        "name": "odf model",
+        "type": "String",
+        "value-key": "[ODF_MODEL]",
+        "value-choices": ["DTI"],
```
gkiar:
```suggestion
        "value-choices": ["DTI", "CSD"],
```

pyafq_cli.json
```diff
@@ -0,0 +1,507 @@
...
+        "value-choices": ["false", "true"],
+        "default-value": "false",
+        "optional": true
+    },
+    {
+        "id": "segmentation_seg_algo",
+        "name": "segmentation seg algo",
+        "type": "String",
+        "value-key": "[SEGMENTATION_SEG_ALGO]",
+        "value-choices": ["AFQ"],
```
gkiar:
```suggestion
        "value-choices": ["AFQ", "Reco"],
```

pyafq_cli.json
arokem: There is some overlap. The full list for recobundles is: `['CST_R', 'C_L', 'F_L', 'UF_R', 'MCP', 'AF_R', 'CCMid', 'F_R', 'UF_L', 'AF_L', 'CC_ForcepsMajor', 'CC_ForcepsMinor', 'CST_L', 'IFOF_L', 'C_R', 'IFOF_R']`.  Note that even where there is overlap (e.g., `'CST'`), it is not clear that they stand for exactly the same thing. That's part of the work we are doing now: figuring out how these segmentation  algorithms differ in terms of their results, the reliability of these results and so on.

arokem: BTW, this list is derived from: 

```
import AFQ.data as afd
reco_bundles = afd.read_hcp_atlas_16_bundles()      
reco_bundles.keys()
```

gkiar: Cool, thanks! For now, I've added this list to the possible value choices and we can re-evaluate if we find a change needs to be made

arokem:
@gkiar : Is this PR ready to merge from your point of view?

gkiar:
@arokem Yep! I'll migrate remaining questions here (about usage but not CLI functionality) to issues.

arokem:
OK. Off we go!

PR #207, opened 2020-01-24
Api Usability

36000:
As I work through using the API, I am putting any changes I make in this pull request. So far:
1. Allowing relative paths for dwi_folder, anat_folder (sub_dir is redundant as sess contains full path already)
2. Small additions to documentation
3. Using f strings for paths
4. Moved helper functions into the class
5. added force recompute

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -867,7 +867,7 @@ class AFQ(object):
             self.reg_template = reg_template
         # This is the place in which each subject's full data lives
         self.dmriprep_dirs = glob.glob(op.join(dmriprep_path,
-                                               '%s*' % sub_prefix))
+                                               f"{sub_prefix}*"))
 
         # This is where all the outputs will go:
         self.afq_dir = op.join(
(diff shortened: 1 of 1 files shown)


AFQ/api.py
```diff
@@ -890,20 +894,17 @@ def __init__(self,
 
                 os.makedirs(results_dir_list[-1], exist_ok=True)
 
-                dwi_file_list.append(glob.glob(op.join(sub_dir,
-                                                       ('%s/%s/%s.nii.gz' %
-                                                        (sess, dwi_folder,
-                                                         dwi_file))))[0])
+                dwi_file_list.append(glob.glob('%s/%s/%s.nii.gz' %
```
36000: Use f strings instead or path lib

arokem:
Move all the helper functions to be methods of the AFQ object. 

Then, we can revisit `force_recompute` (see also #220)

36000:
How should force_recompute work? Should it be a parameter that the user can turn on or off at any time? Or an argument to the functions that the user can call?

arokem:
> How should force_recompute work? Should it be a parameter that the user can turn on or off at any time? Or an argument to the functions that the user can call?

I think that it should be an attribute of the object, set to `False` per default. If the user sets this to `True`, everything that is already computed gets over-written.

arokem:
This looks good!  @36000: Is it ready to merge, from your point of view?

arokem:
Awesome. With this now merged, I am going to make another point release, to see that we get a sane version number on PyPI and that Zenodo uploads work.

PR #223, opened 2020-02-14
Another hack to generate a PyPI acceptable version string.

arokem:


arokem:
Looks like this hack worked! I think we might be ready for a minor release.

arokem:
Unfortunately, this resulted in a rather strange version released on PyPI: https://pypi.org/project/pyAFQ/0.1.263611229

Might have to rethink this approach.

PR #224, opened 2020-02-14
Read the MNI template from templateflow, instead of DIPY.

arokem:
This has the added benefit that we can get a brainmask for our registration from the same place (which is the default behavior implemented here).

### DIFF
--- a/AFQ/__init__.py
+++ b/AFQ/__init__.py
@@ -1,6 +1,3 @@
-from .data import *  # noqa
-from .utils import *  # noqa
-from .api import *  # noqa
 from .version import version as __version__  # noqa
 
 _ga_id = "UA-156363454-3"
--- a/AFQ/__init__.py
+++ b/AFQ/__init__.py
@@ -1,3 +1,6 @@
+from .data import *  # noqa
+from .utils import *  # noqa
+from .api import *  # noqa
 from .version import version as __version__  # noqa
 
 _ga_id = "UA-156363454-3"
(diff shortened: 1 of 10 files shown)


arokem:
This might explain why this fails: 

```
In [1]: import nibabel as nib                                                   

In [2]: img1 = nib.load('/Users/arokem/.dipy/mni_template/mni_icbm152_t2_tal_nli
   ...: n_asym_09a.nii.gz')                                                     

In [3]: img2 = nib.load('/Users/arokem/.cache/templateflow/tpl-MNI152NLin2009cAs
   ...: ym/tpl-MNI152NLin2009cAsym_res-01_T2w.nii.gz')                          

In [4]: img2.shape                                                              
Out[4]: (193, 229, 193)

In [5]: img1.shape                                                              
Out[5]: (197, 233, 189)

In [6]: img1.affine                                                             
Out[6]: 
array([[   1.,    0.,    0.,  -98.],
       [   0.,    1.,    0., -134.],
       [   0.,    0.,    1.,  -72.],
       [   0.,    0.,    0.,    1.]])

In [7]: img2.affine                                                             
Out[7]: 
array([[   1.,    0.,    0.,  -96.],
       [   0.,    1.,    0., -132.],
       [   0.,    0.,    1.,  -78.],
       [   0.,    0.,    0.,    1.]])

```

arokem:
OK - I am pretty sure that this represents a substantial improvement in terms of registration of individual data to the template. I ran this on two very different acquisitions already, to good effect (for example, this seems to resolve #227). I will run it on one more dataset, before merging. After this is merged and after merging #238, I think that we should make a new release.

arokem:
OK - just ran it on one more dataset. This is good to go!

Issue #220, opened 2020-02-11
Enable caching of previous derivatives?


gkiar:
Is there an option to either use or ignore existing derivatives found inside the target output directory (crossref: #219)? Something like `cache = True|False`, where pieces are recycled if existing but that can be turned off if we want a clean slate.

arokem:
We used to have a `force_recompute` key-word argument in the API module, but it added a lot of complexity to the code. Maybe we should revisit?

arokem:
Per default, we do not over-write outputs if they already exist. #207 implements a `force_recompute` key-word argument and #225 implements a config option for the CLI that allow over-riding this default, and recompute everything.

36000:
Moved to #406

PR #225, opened 2020-02-22
Restore force recompute

arokem:


arokem:
Addresses #220

arokem:
OK - I think this works.

PR #228, opened 2020-03-07
More API Fixes

36000:
reg_prealign is sometimes used and assumed to exist. This causes errors when use_prealign is set to False.


arokem:
Could you please add a test that would fail without this change?

PR #232, opened 2020-04-06
Fix warping

Anneef:
Hi, 
this fix corrects the warping of ROI and probability maps in segmentation.py. It had been done with using the forward direction of the warp operation used to warp native on template. Now it is done with the inverse, resulting in ROI and probmap, which were previously registered to template, being warped into native space. 
Also, added debugging lines, which are currently commented-out, in preparation for an optional debugging feature mode that allows step by step tracking of problems with image registration.
_aNNe


arokem:
Thanks! I will go ahead and merge this and then go on to address the issues brought up in #233, as a way to better handle #235 and #234.

PR #234, opened 2020-04-06
registration of AAL to template

Anneef:
just a code suggestion / reminder. Currently the AAL Atlas is not registered to the template, before the warping from template space to native space is applied. This is to say the cleaning by endpoints cannot properly operate. 
This change would fix that. However, I suggest to finally move that registration code to some other, better suited location in the code. One possibility would be the code that registers ROIs and Probmaps. 
Please let me know, when to remove this pull request.


arokem:
I believe this was addressed in https://github.com/yeatmanlab/pyAFQ/commit/0eb99a8bc33412a3d026f488ce62557890de074a

PR #235, opened 2020-04-06
swap the endpoint assignment

Anneef:
assignment of streamline begin to cortical target0 and streamline end to cortical target1 did not result in valid streamlines in our data for ATR. Swapping did. 
This has been briefly tested for a few other tracts, but I am far from sure that this change is valid for all. In fact, I believe that the assignment should be tested routinely as part of the debug-mode: distances between zero end and both cortical target and distances of -1 end and both targets are recorded and if the distribution of distances is clearly different, and does not follow the expected order , the user should be notified to arrange the targets (or rois) in a different sequence.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -832,7 +832,7 @@ def clean_by_endpoints(streamlines, targets0, targets1, tol=None, atlas=None,
             dist0ok = True
         else:
             dist0ok = False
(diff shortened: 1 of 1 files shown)


arokem:
This is now addressed in https://github.com/yeatmanlab/pyAFQ/pull/283/commits/aa39fc3bf181ab0bc5f3acc95e979f52df38d4f0.  Closing

PR #238, opened 2020-04-10
Add dki to api

36000:


36000:
@arokem  dipy.reconst.dki does not have a TensorModel function, while dti does. Could we discuss what this function is doing in our next meeting?

arokem:
Yes! For now I will just mention that the equivalent class for the dki module is [`dipy.reconst.dki.DiffusionKurtosisModel`](https://github.com/dipy/dipy/blob/master/dipy/reconst/dki.py#L1527). It has essentially the same API as the `TensorModel` class.  See also [this example](https://dipy.org/documentation/1.1.1./examples_built/reconst_dki/).

36000:
@arokem if I test the dki model using the stanford hardi data as set up in test_AFQ_data_waypoint I get this error:
image

Do I need data that includes more b values or are we just not using them for this test?

arokem:
Yes. This is because the Stanford data only has one non-zero b-value. You
can try using the cfin dataset that is used in the DIPY Examples

https://dipy.org/documentation/1.1.1./examples_built/reconst_dki/

PR #250, opened 2020-04-28
BF: Actually use the user-provided path for saving HCP files.

### LINKED ISSUE #248: Specifying path for fetch hcp causes error
This line should be moved into the if clause of the above if/else statement: https://github.com/yeatmanlab/pyAFQ/blob/bd01a7df18fc2a87b05e17c56abe3e5b712e40ce/AFQ/data.py#L357
Or the else clause should be
my_path = path

Depending on intended functionality.

arokem:
Fixes #248

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -350,11 +350,9 @@ def fetch_hcp(subjects,
     bucket = s3.Bucket(hcp_bucket)
 
     if path is None:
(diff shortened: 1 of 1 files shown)


arokem:
@36000 : does this fix the issue? I think this logic is more in line with the docstring.

PR #251, opened 2020-04-28
BF: Use Tableau 20 colors for the 20 waypoint-defined bundles.

### LINKED ISSUE #247: Color dictionary needs redone
https://github.com/yeatmanlab/pyAFQ/blob/bd01a7df18fc2a87b05e17c56abe3e5b712e40ce/AFQ/viz.py#L61

@arokem This is the dictionary we currently use for assigning color values.

arokem:
Fixes #247

It turns out that we don't need 40 colors. This is only needed in AFQ-Browser
because we need to be able to highlight the picked bundles.

This means that the code can be simplified quite a bit. This commit also
clarifies and simplifies the logic, including comments. And adds a docstring.

Changes in the API module to remain consistent.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1030,9 +1030,10 @@ class AFQ(object):
         scene = viz.visualize_volume(fa_img,
                                      inline=False,
                                      interact=False)
+
         scene = viz.visualize_bundles(bundles_file,
-                                      affine_or_mapping=row['dwi_affine'],
-                                      bundle_names=self.bundle_dict,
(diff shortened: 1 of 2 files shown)


arokem:
sub-01_sess-01_dwi_space-RASMM_model-DTI_desc-det-AFQ_viz

arokem:
I find the white background to be slightly more visible. Anyone want to bike-shed with me?

sub-01_sess-01_dwi_space-RASMM_model-DTI_desc-det-AFQ_viz

PR #257, opened 2020-05-07
Use xvfb_wrapper for aws

36000:
Uses xvfb_wrapper to handle virtual display on aws in api

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -253,8 +253,8 @@ class AFQ(object):
             diffeomorphic mapping in registration. Default: True
 
         virtual_frame_buffer : bool, optional
-            Whether to use a virtual fram buffer. This is neccessary if generating
-            GIFs in a headless environment. Default: False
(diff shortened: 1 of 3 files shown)


36000:
Remember if statement @36000 !

AFQ/api.py
```diff
@@ -210,6 +211,7 @@ def __init__(self,
                  scalars=["dti_fa", "dti_md"],
                  wm_labels=[250, 251, 252, 253, 254, 255, 41, 2, 16, 77],
                  use_prealign=True,
+                 use_xvfbwrapper=False,
```
arokem, line 214: How about a more explicit name? Also, we should add this to the docstring.


```suggestion
                 virtual_frame_buffer=False,
```

arokem:
Looks like there's still one more small PEP8 issue here. Also, what do you think about changing the keyword argument to `virtual_frame_buffer` instead of `use_xvfb`? Finally, would you mind adding this kwarg and its explanation into the docstring and the CLI?

Issue #258, opened 2020-05-08
API naming convention


36000:
Prepend session and subject id to file names in api. This will make it more convenient to distinguish files from different subjects.

PR #269, opened 2020-06-23
Api File Naming

36000:
Show session and subject in all file names. Moved common inclusions for file names into fname function. Some pep8 fixes.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1046,7 +1046,6 @@ class AFQ(object):
                 f'_viz.gif',
                 include_track=True,
                 include_seg=True))
-        fname = op.join(fname[0], row['results_dir'], fname[1])
 
         scene = viz.scene_rotate_forward(scene)
         viz.create_gif(scene, fname)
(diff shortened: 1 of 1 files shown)


36000:
resolves #258

36000:
@arokem if the dmriprep files already have the sub-XX_sess-XX prefix like in BIDs format shown below, then we do not need this pull request, the api already grabs this prefix from the file. 
[image](https://user-images.githubusercontent.com/9562159/85584547-1825ce80-b5f4-11ea-9c2c-8e33f2b0cd27.png)
Should I use this PR to add in the sub-XX_sess-XX prefix if it doesn't exist, or should I drop this all together?

arokem:
Oh yeah - let's leave the output like the input. But in our own work, let's make sure the inputs are fully spelled out. 

I think that means we can close this?

36000:
No on closing. I also moved common inclusions for file names into the fname function. This won't change functionality but makes the code nicer. For example, Instead of having this in every function:
`
odf_model = self.tracking_params['odf_model']
directions = self.tracking_params['directions']
...
fname = fname + (
                f'_space-RASMM_model-{odf_model}'
                f'_desc-{directions}'
            )
 `
You can instead use the include_track argument when you want to include that information in the file name:
`self._get_fname(row, suffix, include_track=True)`

36000:
I rebased so now this PR only includes the convenient function and some pep8 fixes.

arokem:
LGTM! I love it when a PR removes more lines of code than it adds :-)

PR #270, opened 2020-06-23
Added random_seed parameter in tractography

36000:
This parameter sets the random_seed when random_seeds is True.

### DIFF
--- a/AFQ/tractography.py
+++ b/AFQ/tractography.py
@@ -14,7 +14,7 @@ from AFQ._fixes import VerboseLocalTracking, tensor_odf
 
 
 def track(params_file, directions="det", max_angle=30., sphere=None,
-          seed_mask=None, n_seeds=1, random_seeds=False, random_seed=None,
+          seed_mask=None, n_seeds=1, random_seeds=False, rng_seed=None,
           stop_mask=None, stop_threshold=0, step_size=0.5, min_length=10,
           max_length=1000, odf_model="DTI"):
     """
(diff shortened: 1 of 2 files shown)


36000:
@arokem Does the documentation always fail? Double check I updated everything correctly in the cli.json file before merging.

AFQ/tractography.py
```diff
@@ -14,9 +14,9 @@
 
 
 def track(params_file, directions="det", max_angle=30., sphere=None,
-          seed_mask=None, n_seeds=1, random_seeds=False, stop_mask=None,
-          stop_threshold=0, step_size=0.5, min_length=10, max_length=1000,
-          odf_model="DTI"):
+          seed_mask=None, n_seeds=1, random_seeds=False, random_seed=None,
```
arokem, line 17: This is a very confusing variable name, considering we also already have `random_seeds` 😄

How about `rng_seed`?

36000:
For the ProbabilisticDirectionGetter, I believe that the random_seed set in local tracking is what it ends up using. I just set that to rng_seed.

arokem:
I think that you need to also add this parameter as an option in [bin/pyAFQ](https://github.com/yeatmanlab/pyAFQ/blob/master/bin/pyAFQ)

36000:
> I think that you need to also add this parameter as an option in [bin/pyAFQ](https://github.com/yeatmanlab/pyAFQ/blob/master/bin/pyAFQ)

This is how segmentation parameters are handled in bin/pyAFQ:
image
I believe this means it is loaded from pyafq_cli.json ?

arokem:
No - pyafq_cli.json is here for [boutiques](https://boutiques.github.io/)

But I think that I was wrong. All the arguments get automatically read in https://github.com/yeatmanlab/pyAFQ/blob/master/bin/pyAFQ#L42

So I believe this is good to go.

PR #272, opened 2020-06-24
Bring fetch_hcp into alignment with other fetch functions

36000:
Our last attempt to fix the path names in fetch_hcp introduced another bug. This PR makes it use the same path names as the other fetch functions.


36000:
It looks like right now there are no tests for the fetch_ functions. Should I add some or would that be resource intensive?

arokem:
Thanks! We should definitely look into #275 at some point.

PR #279, opened 2020-06-26
Profile format changed to be less tall and more wide, like mAFQ

### LINKED ISSUE #277: Profile output format in CSV should match mAFQ (more wide)
Specifically, there should not be a scalar column with the names "dti_fa", "dti_md" and a profiles column with their corresponding values. Instead, each scalar should have its own column.

36000:
Instead of a scalar column with the names "dti_fa", "dti_md" and a profiles column with their corresponding values, each scalar gets its own column. closes #277

### DIFF
--- a/AFQ/tests/test_api.py
+++ b/AFQ/tests/test_api.py
@@ -192,7 +192,7 @@ def test_AFQ_data_waypoint():
         'sub-01_sess-01_dwi_space-RASMM_model-DTI_desc-det-AFQ-clean_tractography_idx.json'))  # noqa
 
     tract_profiles = pd.read_csv(myafq.tract_profiles[0])
-    assert tract_profiles.shape == (800, 5)
+    assert tract_profiles.shape == (400, 5)
 
     # Before we run the CLI, we'll remove the bundles and ROI folders, to see
     # that the CLI generates them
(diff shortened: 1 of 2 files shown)


AFQ/tests/test_api.py
```diff
@@ -192,7 +192,7 @@ def test_AFQ_data_waypoint():
         'sub-01_sess-01_dwi_space-RASMM_model-DTI_desc-det-AFQ-clean_tractography_idx.json'))  # noqa
 
     tract_profiles = pd.read_csv(myafq.tract_profiles[0])
-    assert tract_profiles.shape == (800, 5)
+    assert tract_profiles.shape == (400, 5)
```
arokem, line 195: Hmm. Why are there only half the numbers here?

richford:
No reason other than the appeal to tidy data. That's just the format that we used for AFQ-Browser and AFQ-Insight. Inside of AFQ-insight, we want the feature matrix to be wide (n x p, where n in the number of subjects and p is the number of features), but I can definitely do that conversion inside of AFQ-Insight. I don't remember if we had good reasons to prefer the wide format inside of AFQ-Browser other than the appeal to tidy data.

richford:
Sorry for the lingo. I'm referring to [this definition of tidy](https://www.jstatsoft.org/index.php/jss/article/view/v059i10/v59i10.pdf). Hadley gets extra points for the Tolstoy quote.

arokem:
OK - sounds good. Let's go with the wide format proposed here (and already implemented in AFQ-Browser and AFQ-Insight).  

Only still need to understand the changes introduced in the tests, but otherwise, LGTM.

AFQ/tests/test_api.py
36000: Two columns were removed: scalar and profile. scalar was the label (ie "dti_fa") and profile was the number. Now, instead, there are two columns of numbers: "dti_md", "dti_fa". This means it is only half the length. For each node, there is only one row instead of one for each scalar.

arokem: Gotcha! That makes sense.

arokem:
With that question, I believe this is ready to merge? Is it rebased on master?

36000:
I just rebased it. I believe it can merge when these checks complete.

PR #281, opened 2020-06-30
make_bundle_dict should only be called after reg_template is settled

36000:
I found this bug while looking at FA registration. @arokem  I wonder if this is the source of some of our registration problems.


36000:
Registration test passes locally. I wonder if bids just updated.

arokem:
Good catch!

Issue #264, opened 2020-05-19
Use nitransforms for spatial transformations


arokem:
See this notebook for concepts and API: https://github.com/poldracklab/nitransforms/blob/master/docs/notebooks/NiTransforms%20-%20ISBI%202020.ipynb

arokem:
And use the x5 file format therein to store the transforms that are generated. I believe this covers both the affine transform and the non-linear transforms that are currently stored as niftis

arokem:
Make it so that you can accept registrations and warps created by other pipelines (e.g., qsiprep)

PR #283, opened 2020-06-30
Allow user to customize registration targets, fix some other aspects of registration

### LINKED ISSUE #226: Use SLR to register to MNI, instead of image-based registration with SyN
I believe that we already have an MNI-aligned whole-brain tractography in the recobundles atlas, which can be used for this purpose.

### LINKED ISSUE #243: Use FA template for alignment
There's a nice-looking MNI-registered FA template here: https://www.fmrib.ox.ac.uk/ukbiobank/

36000:
Closes #226 , Closes #243


36000:
@arokem . I am not sure how to add this to the CLI file. Should it be a part of the "Bundles Group" or should I start a new "Registration Group"?

36000:
Should there also be a way to customize the bundle dict? I.e. ROIs, probability maps

36000:
Call it template_group.

36000:
Let the fetcher get the entire 1GB. Warn people that its so big. Unzip and keep only FA.

36000:
Let whether to mask the templates optional

36000:
The FA template has shape (182,218,182), while the mask has shape (193,229,193). I am just going to crop the mask, as they look like they are in the same space.

arokem:
You probably want to do something like this: https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/data.py#L104-L108, rather than cropping. Most viewers (e.g., MI-BRAIN) do something like that under the hood, to make things show up in the same space.

36000:
Should reg_template always be the same as static? Right now I am using the FA template for the syn registration but the reg_template everywhere else. In effect this means reg_template is being used as the codomain image when calling read_mapping. This should work as the MNI template and FA template are in the same space, but does it make sense?

36000:
rename moving / static to reg_template  and reg_subject.

36000:
As of this commit: https://github.com/yeatmanlab/pyAFQ/pull/283/commits/d3971a1e277a33fc3062e049dc46b4471c8a18d9
These tests pass locally. I wonder what is causing them to stop when remote.

arokem:
I'm guessing something is taking a long time to run and GitHub actions gives up after it sees nothing in standard out for a long time. This might be helpful: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes ?

36000:
Swapped some endpoint filtering, which Closes #235

36000:
Closes #234 by saving the registered template.

36000:
@arokem  this is rebased and ready to merge.

AFQ/_fixes.py
```diff
@@ -5,12 +5,28 @@
...
 
+class ConformedAffineMap(AffineMap):
+    """
+    Modifies AffineMap API to match functionality of DiffeomorphicMap.
+    Important for SLR maps to be indistinguishable from SYN maps
+    in terms of API.
+    """
+    def transform(self, *args, interpolation='linear', **kwargs):
+        kwargs['interp'] = interpolation
+        return super().transform_inverse(*args, **kwargs)
```
arokem, line 23: Should this really be `transform_inverse` here?

AFQ/_fixes.py
```diff
@@ -5,12 +5,28 @@
...
+
+    def transform_inverse(self, *args, interpolation='linear', **kwargs):
+        kwargs['interp'] = interpolation
+        return super().transform(*args, **kwargs)
```
arokem, line 27: And `transform` here?

AFQ/api.py
```diff
@@ -182,14 +188,48 @@ def __init__(self,
...
+            "subject_sls", "hcp_atlas", "mni_T1", or "power_map,"
+            image data will be loaded automatically.
+
+        reg_template : str or Nifti1Image, optional
+            The target image data for registration.
+            Can either be a Nifti1Image, a path to a Nifti1Image, or
+            If "mni_T2", "b0", "dti_fa_subject", "dti_fa_template",
+            "subject_sls", "hcp_atlas", "mni_T1", or "power_map,"
+            image data will be loaded automatically.
+            (defaults to the MNI T2)
```
arokem, line 223: I am a little bit confused here: `reg_subject` is in the subject native space and `reg_template` is in some (usually MNI for now) template space, right? 

So it doesn't make sense to use `reg_subject="mni_T2"` or `reg_subject="mni_T1"` or, conversely `reg_template="dti_fa_subject"` or `reg_template="power_map"`. Or am I missing something?

 This can never be `dti_fa_subject` or `power_map`, right? We only have these for the individual subject right now.

AFQ/api.py
```diff
@@ -182,14 +188,48 @@ def __init__(self,
...
+
+        mask_templ : bool, optional
```
arokem, line 225: Let's call this `mask_template`.

AFQ/api.py
```diff
@@ -182,14 +188,48 @@ def __init__(self,
...
+            Whether to mask the chosen template(s) with a brain-mask
+            Default: True
+
+        reg_algo : string, optional
+            Algorithm to use for registration.
+            Can be one of: {"syn", "slr"}. Default: "syn"
```
arokem, line 231: If you are using `sls` and `hcp_atlas` this has to be SLR, right? And vice versa, if these are images it has to be "syn". So maybe we don't need user input?

AFQ/tests/test_api.py
```diff
@@ -161,12 +161,44 @@ def test_AFQ_data():
...
         npt.assert_equal(nib.load(myafq.b0[0]).shape,
-                        nib.load(myafq['dwi_file'][0]).shape[:3])
+                         nib.load(myafq['dwi_file'][0]).shape[:3])
         npt.assert_equal(nib.load(myafq.b0[0]).shape,
-                        nib.load(myafq.dti[0]).shape[:3])
+                         nib.load(myafq.dti[0]).shape[:3])
         myafq.export_rois()
 
 
+# def test_AFQ_slr():
```
arokem, line 170: Maybe instead of commenting these out we can mark them for skipping as slow and implement the following pattern: https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

AFQ/_fixes.py
36000: For these two its really weird. As far as I can tell the DiffeomorphicMap and AffineMap have the opposite definitions of what is transform inverse and what is transform. Maybe I should just swap what is domain and codomain when I pass in, instead of swapping these transforms?

AFQ/api.py
36000: I will update the documentation so that only sensible options are listed here.

AFQ/api.py
36000: I now assume reg_algo and logger.error if the registration choices mix slr/syn. Should I throw an error instead of using logger.error?

AFQ/tests/test_api.py
36000: I copied the code for contest.py from that website. I added a link to the website in conftest.py . Is this a good way to document where the code came from?

AFQ/_fixes.py
arokem: Hmm. Could we raise this as an issue in DIPY? I saw that you already posted an issue about the API differences, but I think that it would be worth asking about this issue as well. It seems odd that this has the opposite directionality, and I don't know why that is. I want to make sure that we understand this correctly.

arokem: For now, let's leave it as-is. It's being in the fixes module will (hopefully) help us remember that we need to eventually reconcile this.

AFQ/api.py
```diff
@@ -139,18 +139,23 @@ def make_bundle_dict(bundle_names=BUNDLES, seg_algo="afq", resample_to=False):
...
     """
+
     def __init__(self,
                  bids_path,
                  dmriprep='dmriprep',
                  segmentation='dmriprep',
                  seg_suffix='seg',
                  b0_threshold=0,
+                 target_b_val=None,
+                 target_b_sensitivity=50.0,
```
arokem, line 150: I thought about this a bit more: instead of these, we could have `max_bval` and `min_bval` for this selection? Also, am I wrong to think that the selection here: https://github.com/yeatmanlab/pyAFQ/pull/283/files#diff-ae57f5dda2497f45a9b9955970ea8bddR466, throws out the b0? I don't think we want to do that. So we need to include all the volumes that have a b-value smaller than b0_threshold and all the volumes between max_bval and min_bval.

AFQ/api.py
```diff
@@ -233,8 +280,31 @@ def __init__(self,
...
-        self.use_prealign = use_prealign
+        self.reg_subject = reg_subject
+        self.reg_template = reg_template
+        self.mask_template = mask_template
+        if reg_subject == 'subject_sls' or reg_template == 'hcp_atlas':
+            if reg_template != 'hcp_atlas':
+                self.logger.error(
+                    "If reg_subject is 'subject_sls',"
+                    + " reg_template must be 'hcp_atlas'")
+            if reg_subject != 'b0':
```
arokem, line 299: Shouldn't this be as follows?

```suggestion
            if reg_subject != 'subject_sls':
```

AFQ/data.py
```diff
@@ -926,10 +930,46 @@ def s3fs_json_write(data, fname, fs=None):
...
+                         mask_img.affine,
+                         template_img.affine),
+            template_img.affine)
+        mask_data = mask_img.get_fdata()
+
+    out_data = template_data * mask_data
+    return nib.Nifti1Image(out_data, template_img.affine)
+
+
+def read_mni_template(resolution=1, mask=True, weight=2):
```
arokem, line 969: Let's change this to a string to match the DIPY API: `"T2w"` or `"T1w"`

```suggestion
def read_mni_template(resolution=1, mask=True, weight=2):
```

AFQ/data.py
```diff
@@ -940,25 +980,145 @@ def read_mni_template(resolution=1, mask=True):
...
+                'biobank_templates'),
+        "http://biobank.ctsu.ox.ac.uk/showcase/showcase/docs/",
+        ["bmri_group_means.zip"],
+        ["bmri_group_means.zip"],
+        data_size="1.1 GB",
+        doc="Download UK Biobank templates",
+        unzip=True)
+
+
+def read_fa_template(mask=True):
```
arokem, line 1018: Let's clearly denote that this is from the UKBB:

```suggestion
def read_ukbb_fa_template(mask=True):
```

AFQ/data.py
```diff
@@ -940,25 +980,145 @@ def read_mni_template(resolution=1, mask=True):
...
+    """
+    Reads the FA template
```
arokem, line 1020:
```suggestion
    Reads the UK Biobank FA template
```

AFQ/_fixes.py
36000: I will add a comment with an explanation, and also add the issue for dipy.

36000: OK. Looking further into dipy code, it seems like they should be doing the same transformation. Probably some problem on our end. I am going to double check everything again and run the code, see if I can isolate the problem.

AFQ/api.py
36000: good catch!

AFQ/_fixes.py
36000: I figured out what was happening. When we load the DiffeomorphicMap we set it to inverse: https://github.com/yeatmanlab/pyAFQ/blob/822af52d43e960a84fa5ed0b831d97fdea75c7a7/AFQ/registration.py#L197 I am not really sure why. It makes all of our transforms transform_inverses and vice versa from then on out. That's why I had to do the same for the AffineMap to make it work. It seems to be necessary for the tests to make sense. @arokem, do you remember why we need to do this?

arokem: No - I don't exactly remember why I set it this way. 

Let's see if  I understand, though: if we set this to `mapping.is_inverse = False` there, and then drop the code here, so that it does the same that DIPY currently does, would the tests then pass?

36000: Nope. mapping.is_inverse is for the DiffeomorphicMap, while above is for AffineMap. If we dropped both, we might have to change all of the transforms to transform_inverse in our code or something like that. I will look into this further.

arokem: Maybe we can keep this fix for now, and then come back and clean this all up by moving to use [`nitransforms`](https://github.com/poldracklab/nitransforms) at some later point.

arokem: See https://github.com/yeatmanlab/pyAFQ/issues/264

36000: Ok. I figured it out. It's very tricky and may deserve an issue report on Dipy.
The optimize function from SymmetricDiffeomorphicRegistration, which we use to get our Diffeomorphic map, 
returns with is_inverse set to true. This means, internally, transform are transform_inverse and vice versa. I think this flag is used in some important calculation in sdr.optimize, however, it should almost certainly be set back to false before the result is returned. This means, I think, that the transforms and transform_inverses in our code all need to be switched. We might want to talk all of this through on Monday.

AFQ/tests/test_api.py
arokem: Yes - that works well. Are we using this in the CI to skip the slow tests?

AFQ/_fixes.py
arokem: OK - for now, since we're not going to have an opportunity to discuss next week, let's leave this as-is for now, and then post an issue pointing to this conversation, so that we can come back to this later.

AFQ/tests/test_api.py
36000: Yup! The two slow tests I added are now being skipped.

AFQ/api.py
```diff
@@ -392,12 +451,26 @@ def log_and_save_trk(self, sft, fname):
...
+            data = data[
+                ...,
+                (gtab.bvals >= self.min_bval)
+                or (gtab.bvals <= self.b0_threshold)]
+        if filter_b and (self.max_bval is not None):
+            data = data[
+                ...,
+                (gtab.bvals <= self.max_bval)
+                or (gtab.bvals <= self.b0_threshold)]
+        return data, gtab, img
```
arokem, line 468: Sorry: one more thing -- shouldn't we also create a new GradientTable object that has only the relevant b-values and corresponding b-vectors?

36000: Good catch. I just rewrote this function in a way which I think makes more sense.

PR #289, opened 2020-07-02
Viz module uses fury and plotly

### LINKED ISSUE #273: Viz module hangs inside of jupyterlab on aws
If you run visualize_bundles from viz.py with inline=true, or try to use create_gif, python hangs.

### LINKED ISSUE #274: Visualize gifs that are in S3 inline in a notebook
Something like `afd.s3fs_visualize_gif` that takes as string (full path to the S3 file) and then drops the animated gif into the notebook (I think using https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html --  same as the inline trick for visualize_bundles).

### LINKED ISSUE #276: Integrate plotly into visualization
Plotly may improve the interactiveness of our visualizations and sidestep some bugs we have been running into: https://plotly.com/chart-studio-help/make-a-streamtube-plot/

36000:
Adds plotly functionality. Changes viz module to allow choice of fury and plotly. Updates github workflows set up to use fury. Adds inline gif visualization. Fixes vtk version to remove warnings / transparency issues. Redesigns visualization api to allow more customization. Plotly works well in notebook and in cloud. Closes #276 , Closes #273, Closes #274 .


arokem:
Do you think the code would be very ugly if we tried to make it possible to choose a backend, "fury" or "plotly" and maintain both functionalities?

36000:
I don't think it would be too ugly. I think if we do that we should be careful to only require/import one of them in the default case though, because both involve their own dependency chains.

arokem:
I think that fury probably gives us better performance with large amounts of data. 

One way to do this would be to split the viz module in two, and make sure that imports are kept within sub-module, so: 

cloudknot/viz/__init__.py
cloudknot/viz/utils.py <= common things (e.g., color maps)
cloudknot/viz/fury.py <= fury-dependent functionality
cloudknot/viz/plotly.py <= plotly-dependent functionality

36000:
that's a good idea! I will get started on that. In that case, what should the requirements file look like? Should we keep it with only fury by default? It seems like requiring the user to install more than one visualization scheme would be extraneous.

36000:
> cloudknot/viz/**init**.py
> cloudknot/viz/utils.py <= common things (e.g., color maps)
> cloudknot/viz/fury.py <= fury-dependent functionality
> cloudknot/viz/plotly.py <= plotly-dependent functionality

I assume you mean AFQ, not cloudknot?

arokem:
Oh yeah - OFC. Sorry - my head is in cloudknot this morning  :-0

For the requirements, we can do what we do on cloudknot where there are optional requirements defined. See: https://github.com/nrdg/cloudknot/blob/master/cloudknot/version.py#L94. This way, when the package is installed with

```
pip install cloudknot[dev]
```
pytest and flake8 are installed. 

For us, we'd have extras for 

```
pip install cloudknot[fury]
```
and 
```
pip install cloudknot[plotly]
```
Or, if you want all the functionality

```
pip install cloudknot[plotly, fury]
```

Something like that

36000:
A lot of the documentation is similar across functions between viz_fury and viz_plotly. Is there a way to reduce this duplication?

36000:
Having the modules literally named fury.py and plotly.py leads to all sorts of import problems because it conflicts with the actual fury and plotly. Should they be called render_fury.py or afq_fury.py or viz_fury.py or something else?

36000:
> For the requirements, we can do what we do on cloudknot where there are optional requirements defined. See: https://github.com/nrdg/cloudknot/blob/master/cloudknot/version.py#L94. This way, when the package is installed with

I think this is a good idea but should be in a separate PR, because it involves replacing requirements files with version.py . For now, we can leave the requirements files as is.

36000:
Not sure what is causing some of these errors: https://github.com/yeatmanlab/pyAFQ/pull/289/checks?check_run_id=831846984#step:6:54

Maybe it is confused because I added a new folder? Some settings file I need to change?

36000:
@arokem this is ready to be reviewed / merged

36000:
make sure orca isnt imported unless in relevant function. Cath import error, explain what is needed for gif with plotly and why we cant pip install it. Recommend use of fury instead.

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,318 @@
...
+        plotly.io.orca.shutdown_server()
+    except ValueError:
+        _orca_err()
+
+
+def _orca_err():
+    print("pyAFQ is trying to generate gifs using plotly. "
+          + "This requires orca, which cannot be installed via pip. "
+          + "See: https://github.com/plotly/orca \n"
+          + "Or consider using fury to visualize with pyAFQ instead")
```
arokem, line 141:
```suggestion
    except 
        ValueError(_orca_err())


def _orca_err():
    return ("pyAFQ is trying to generate gifs using plotly. "
          + "This requires orca, which cannot be installed via pip. "
          + "See: https://github.com/plotly/orca \n"
          + "Or consider using fury to visualize with pyAFQ instead")
```

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,318 @@
...
+        camera = dict(
+            eye=dict(x=np.cos(theta) * zoom,
+                     y=np.sin(theta) * zoom, z=z_offset)
+        )
+        figure.update_layout(scene_camera=camera)
+        try:
+            figure.write_image(tdir + f"/tgif{i}.png")
+        except ValueError:
+            _orca_err()
+            return
```
arokem, line 165: I think that we do want the error raised, but with an explanation:

```suggestion
        except 
            ValueError(_orca_err())
```

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,318 @@
...
+        )
+    )
+
+
+def visualize_roi(roi, affine_or_mapping=None, static_img=None,
+                  roi_affine=None, static_affine=None, reg_template=None,
+                  figure=None, color=np.array([0.9999, 0, 0]), opacity=1.0,
+                  interact=False, inline=False):
+    """
+    Render a region of interest into a VTK viz as a volume
```
arokem, line 192: Needs a docstring

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,318 @@
...
+            lenmode='fraction',
+            len=0.6
+        ))  # TODO: these sliders won't become independent!
+
+
+def visualize_volume(volume, figure=None, show_x=True, show_y=True,
+                     show_z=True, interact=False, inline=False, opacity=0.3,
+                     slider_definition=0):
+    """
+    Visualize a volume
```
arokem, line 296: Needs a docstring.

AFQ/viz_libs/utils.py
```diff
@@ -0,0 +1,186 @@
+from collections import OrderedDict
```
arokem, line 1: Would be good to have some docstrings in this module, even for the helper functions.

.github/workflows/docbuild.yml
```diff
@@ -23,10 +23,10 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
-        pip install .
+        pip install -r requirements-fury.txt
     - name: Install software
       run: |
-        pip install .
+        pip install -e .
```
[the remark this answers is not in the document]
36000: For some reason, the standard install was running into errors. It could not find the new folder I added. However I will try this change and see if it runs into the same error.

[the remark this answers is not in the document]
arokem: Did you add  an `__init__.py` file in that folder? It has to have that to be considered a Python module.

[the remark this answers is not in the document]
36000: I think that may be the problem! I will do that.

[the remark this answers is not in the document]
36000: @arokem  Now I am experiencing a new issue.
`pip install '.[dev,fury]'` causes the following warnings:
```
WARNING: pyAFQ 0.1.dev119761714318 does not provide the extra 'dev'
WARNING: pyAFQ 0.1.dev119761714318 does not provide the extra 'fury'
```
However, pip install -e '.[dev,fury]' and pip install 'pyAFQ[dev,fury]' do not cause these warnings. 

These warnings are what is causing the PR to fail the test suite (its not installing dev or fury).

It is exactly the same as what they describe here, but this is 5 years old and marked resolved (my pip is updated):
https://github.com/pypa/pip/issues/1236

36000:
@arokem I had to mess with some of the dependencies of our dependencies to get the install to work correctly in both environments, but I think it is OK now. I also had to use -e (see new comment above). So there have been a few changes to installation. I also added documentation.

36000:
Consider removing axes in 3d plots

36000:
anti aliasing in FA map?

36000:
Output html files into folder likes gifs

36000:
Try to make the streamlines look more like the tubes in fury

36000:
Color by local direction or by local FA (this might be separate PR)
Make whole interactive webpage (separate PR)

36000:
Option to use different volume than FA

36000:
add optional argument for single bundle in visualize ROIs

36000:
modulating fiber thickness or opacity when you hover over

.github/workflows/docbuild.yml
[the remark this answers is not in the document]
arokem: Sorry - more questions here. I see that you wrote `pip install '. [dev, fury]'`. 

Do you really have those quotes in there? What about `pip install .[dev,  fury]` (no quotes)?

arokem:
Could we rename `viz_libs` simply to `viz`?

AFQ/api.py
```diff
@@ -211,6 +211,7 @@ def __init__(self,
                  wm_labels=[250, 251, 252, 253, 254, 255, 41, 2, 16, 77],
                  use_prealign=True,
                  virtual_frame_buffer=False,
+                 viz_library="fury",
```
arokem, line 214:
```suggestion
                 viz_backend="fury",
```

AFQ/api.py
```diff
@@ -256,6 +257,10 @@ def __init__(self,
             Whether to use a virtual fram buffer. This is neccessary if
             generating GIFs in a headless environment. Default: False
 
+        viz_library : str, optional
```
arokem, line 260:
```suggestion
        viz_backend : str, optional
```

AFQ/api.py
```diff
@@ -256,6 +257,10 @@ def __init__(self,
...
+            Should be either "fury" or "plotly".
```
arokem, line 261:
```suggestion
            Which visualization backend to us. One of {"fury", "plotly"}.
```

AFQ/api.py
```diff
@@ -281,6 +286,7 @@ def __init__(self,
             from xvfbwrapper import Xvfb
             self.vdisplay = Xvfb(width=1280, height=1280)
             self.vdisplay.start()
+        self.viz = viz.Viz(viz_library=viz_library)
```
arokem, line 289:
```suggestion
        self.viz = viz.Viz(backend=viz_backend)
```

bin/pyAFQ
```diff
@@ -187,6 +189,7 @@ myafq = api.AFQ(dmriprep_path,
                 reg_template=reg_template,
                 wm_labels=wm_labels,
                 virtual_frame_buffer=virtual_frame_buffer,
+                viz_library=viz_library,
```
arokem, line 192:
```suggestion
                viz_backend=viz_backend,
```

bin/pyAFQ
```diff
@@ -172,6 +173,7 @@ compute = config.get("compute", {})
 dask_it = compute.get("dask_it", False)
 force_recompute = compute.get("force_recompute", False)
 virtual_frame_buffer = compute.get("virtual_frame_buffer", False)
+viz_library = compute.get("viz_library", 'fury')
```
arokem, line 176:
```suggestion
viz_backend = compute.get("viz_backend", 'fury')
```

bin/pyAFQ
```diff
@@ -86,6 +86,7 @@ wm_labels = [250, 251, 252, 253, 254, 255, 41, 2, 16, 77]
 dask_it = false
 force_recompute = false
 virtual_frame_buffer = false
+viz_library = 'fury'
```
arokem, line 89:
```suggestion
viz_backend = 'fury'
```

arokem:
I see now. I think we can reorganize this. 

I would move everything that is currently in `AFQ/viz.py` to `AFQ/viz/utils.py` and then import the names you want available at the top level into the `__init__.py` file for that module. For example: 

```
from .utils import Viz
```

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,552 @@
...
+    if cbv is not None:
+        customdata = []
+        line_color = []
+        cbv_max = cbv.max()
+        color_max = color.max()
+    else:
+        customdata = None
+        line_color = _color_arr2str(color)
+
+    for sl in sls:
```
arokem, line 86: I think that this could be sped up by using the underlying data in a `StatefulTractogram`. 

Looking at one of these objects, if you do something like this:

```
sft = afd.load_tractogram('fname.trk', 'ref.nii.gz')
```
then, `sft.streamlines._data` is an `(N, 3)` array (with N the number of nodes is all the streamlines), and `sft.streamlines._offsets` is an `(M, )` array (M is the number of streamlinse) where each item is the index of the start of the next streamline (the first item is always `0`, for the start of the first streamline). There is also an attribute `sft.streamlines._lengths`, which provides complementary information. 

So -- I *think* that you could avoid this loop and the rather expensive calls to `extend` below, potentially using only preallocations and indexing operations.

AFQ/viz_libs/afq_plotly.py
```diff
@@ -0,0 +1,552 @@
...
+        Opacity of slices.
+        Default: 1.0
+
+    slider_definition : int, optional
+        How many discrete positions the slices can take.
+        If 0, slices are stationary.
+        Default: 50
+
+    which_plane : str, optional
+        Should be 'Coronal', 'Axial', 'Sagittal', or None.
```
arokem, line 494:
```suggestion
        Which plane can be moved with a slider.
        Should be 'Coronal', 'Axial', 'Sagittal', or None.
```

arokem:
One more renaming suggestion: maybe instead of `afq_plotly` and `afq_fury`: `plotly_backend` and `fury_backend`?

.github/workflows/docbuild.yml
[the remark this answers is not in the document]
36000: it returns bad pattern on zsh and invalid requirement on bash. If you do no quotes and no spaces, It works on bash but not zsh. With quotes works on both. I have not tested other shells.

AFQ/viz_libs/afq_plotly.py
36000: Making a StatefulTractogram involves getting a reference / space or limiting possible inputs to only trk filenames. Is that something we want to require here?

arokem: In the context it's currently used (as a helper function in `visualize_bundles`) these elements are already required, so maybe that's not too burdensome. 

OTOH, we might not even need to require that. `sft.streamlines` is an `ArraySequence` object. So, what if we require the `sls` input here be an `ArraySequence` object? Then it has these attributes (`_data` ,'_offsets' and `lengths`). 

I guess I don't know what `sls` is now. What kind of object can be used as input here?

36000: It comes in as:
trk : str, list, or Streamlines
        The streamline information
that gets passed to tract_loader which returns a tractogram, which then gets passed to tract_generator which returns a list of streamlines (the documentation should be more clear here, I will update it), which is used by either fury or plotly to plot.

Considering we only call this from the API, it may make sense to require a stateful tractogram or path to trk file or image instead of just streamlines. However, currently we only require streamlines, no image information required. The question is, would end users directly visualize_bundles find the image requirement annoying/unnecessary? 

If we wanted maximum flexibility, we could change which algorithm we use depending on the information given. IE, if a stateful tractogram is passed in, use its offsets, but if a normal tractogram is passed in, iterate through it.

arokem: I think that it would be OK to require `StatefulTractogram` as input. We use SFT objects everywhere else.

36000:
@arokem Not only did using numpy arrays / sft's make plotly rendering significantly faster, it also somehow fixed the trace rendering order (don't ask me how these are related, plotly can be very finicky). I have rebased. If this gets a green check mark, it should be ready to merge.

arokem:
Did you rebase on master?

36000:
Yes, it is rebased on master

Issue #255, opened 2020-04-30
Allow S3 paths as input to API


arokem:
If an S3 path is provided, we save and look for files on S3, instead of locally and use various s3fs functions to write/read.

PR #290, opened 2020-07-02
Add classes for fetching a BIDS-compliant study on S3

richford:
This PR adds classes to AFQ.data that can query and download subjects from a BIDS-compliant study hosted on Amazon S3. It is related to #255 but is much narrower in scope and does not resolve that issue. The main use case is to create a subset BIDS study for use on distributed computing elements.

Here's some example usage using the `HBN` class, which inherits from `S3BIDSStudy`:
```python
import AFQ.data as afqd
hbn = afqd.HBN(subjects=3, random_seed=42)

# List all subjects in the HBN study
print(hbn._all_subjects)

# List the three random subjects that we asked for. This is a list of `S3BIDSSubject` objects
print(hbn.subjects)

# Take a look at the first one
s0 = hbn.subjects[0]

# Print the s3 keys for this subject
print(s0.s3_keys)

# Download this subject
s0.download("./hbn-test", include_site=True, include_derivs=False)

# Print a dict which maps the s3 keys to local downloaded files
print(s0.files)

# Download the rest of the study all at once (using multithreading and skipping previously downloaded files for s0, unless `overwrite=True`)
hbn.download("./hbn-test", include_site=True, include_derivs=False)

# Maybe we want all of the mindboggle derivatives too, but not the freesurfer derivatives
# Unless `overwrite=True`, this will skip all the previously downloaded files
hbn.download("./hbn-test", include_site=True, include_derivs=["mindboggle"])

# Okay, maybe we want all the derivatives, after all
hbn.download("./hbn-test", include_site=True, include_derivs=True)

# Hmm, I'm concerned that we're missing some s3 keys that live at the higher levels of the directory structure and should be inherited, let's look at them
print(hbn.non_subject_s3_keys)
```

Remaining questions or todo items:
- [x] It'd be nice to give the user some way to easily download those "non-subject" s3 keys.
- [x] More sophisticated treatment of sites. Currently, if `multisite=True` in the S3BIDSStudy parameters, we assume that all of the top-level directories are Site IDs. But maybe there's some other top level data in there. It'd be nice to be able to differentiate between "Site-RU" and a top-level "participants.tsv," for example. 
- [x] Testing with [moto](http://docs.getmoto.org/en/latest/docs/getting_started.html)


arokem:
Looks great! 

Regarding this

>  It'd be nice to give the user some way to easily download those "non-subject" s3 keys

Once you have these keys in hand, it's a call to `fs.get` away, right? But the user would have to sort through them, to see which ones they want to download?

richford:
Agreed. So we probably don't need to include it in AFQ.data. They already have the list of s3 keys. So maybe I'll check that off the todo list and assume that we put the rest of that responsibility on the user.

richford:
Still a WIP. I did some significant restructuring after realizing that HBN is not BIDS compliant. We now have an `S3BIDSStudy` class and a separate `HBNSite` class. I think the remaining things to do are:
- [ ] Test with moto, this will require adding moto to the `[dev]` requirements in `extras_require`.
- [ ] Rewrite the `download` method for the HBN site and subject classes so that it BIDS-ifies the output on the local disk

We could also consider creating an `HBN` object that contains all of the `HBNSite` objects. But why? As tidy as it would seem, I can't think of an authentic use case.

richford:
@arokem Is the documentation build failing because the PR is coming from my fork and not from yeatmanlab?

arokem:
That looks like the return of https://github.com/numpy/numpydoc/issues/268. Is this branch rebased on current master?

richford:
Nope. Just rebased. Thanks for the tip on the 🧟 bug!

richford:
Hmm, no such luck. Anyway, low priority since this is still a WIP. We can come back to it once I implement the other things.

richford:
I tried updating `requirements-dev.txt` to require numpydoc==1.1.0 but that did not fix my local doc build. I doubt it will work on GitHub but we'll see.

richford:
Thanks @36000 for fixing the docbuild issue!

richford:
Added tests using moto. They pass locally. Let's see if they pass here. Then I think this is ready for review.

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+
+        results = [delayed(sub.download)(
+            directory=directory,
+            include_derivs=include_derivs,
+            overwrite=overwrite,
+            pbar=pbar,
+            pbar_idx=idx,
+        ) for idx, sub in enumerate(self.subjects)]
+
+        compute(*results, scheduler='threads')
```
arokem, line 1106: Wouldn't you rather use the dask progressbar construct here as well? Instead of the pbar input to `sub.download`?

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+            include_non_sub_raw_keys=include_non_sub_raw_keys,
+            include_derivs=include_derivs,
+            overwrite=overwrite,
+            pbar=pbar
+        )
+
+        to_bids_description(
+            directory,
+            **{"BIDSVersion": "1.0.0",
+               "Name": "HBN Study, " + self.site})
```
arokem, line 1290: We might want to add some additional information here. For example, what subjects were selected this time. I think you can add a list here. (and yes, we should probably also do this for the HCP data as well, if we aren't).

AFQ/tests/test_data.py
```diff
@@ -1,9 +1,154 @@
...
 
 import AFQ.data as afd
 
+DATA_PATH = op.join(op.abspath(op.dirname(__file__)), "data")
+TEST_BUCKET = "test-bucket"
+TEST_DATASET = "ds000102-mimic"
+
+
+@pytest.fixture
+def temp_data_dir():
```
arokem, line 23: Is this better in some way than https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/tests/test_api.py#L217 ?

requirements.txt
```diff
@@ -1,22 +1,22 @@
...
+scipy==1.2.0
 setuptools_scm[toml]==3.4.3
-popylar==0.2
-imageio==2.8.0
-ipython==7.13.0
 templateflow==0.6.2
-pybids==0.11.1
+toml==0.10.0
+toolz==0.9.0
+tqdm==4.32.2
```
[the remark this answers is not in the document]
36000: I think he alphabetized it.

36000:
There could also be an upload function similar to the download function, which uploads derivatives generated by AFQ back into the same folder. So the cloudknot workflow would look like:
```
study = afqd.S3BIDSStudy()
study.download()

myafq = api.AFQ(bids_folder, virtual_frame_buffer=True)
myafq.export_all()

study.upload()
```

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+        self._local_directories.append(directory)
+        self._local_directories = set(self._local_directories)
+
+        results = [delayed(sub.download)(
+            directory=directory,
+            include_derivs=include_derivs,
+            overwrite=overwrite,
+            pbar=pbar,
+            pbar_idx=idx,
+        ) for idx, sub in enumerate(self.subjects)]
```
36000, line 1104: This should also the bids descriptor for the study.

requirements.txt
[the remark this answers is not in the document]
richford: Indeed, I got carried away and alphabetized. Happy to change back if you want.

AFQ/tests/test_data.py
richford: Oh, cool. I didn't know about `nibabel.tmpdirs`. The main difference I see is that `nbtmp.InTemporaryDirectory()` also changes the current working directory. Also, using is as a `pytest.fixture` with a yield statement will make sure that we delete the directory even if the test fails. I'm not sure what happens to the temporary directory if the test you linked in `test_AFQ_data` fails in the middle. but maybe it also deletes the directory.

requirements.txt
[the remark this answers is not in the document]
richford: But It appears that this is all happening in setup.cfg now anyway.

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+        Whether to use anonymous connection (public buckets only).
+        If False, uses the key/secret given, or boto’s credential
+        resolver (client_kwargs, environment, variables, config files,
+        EC2 IAM server, in that order). Default: True
+
+    Returns
+    -------
+    subjects : dict
+    """
+    fs = s3fs.S3FileSystem(anon=anon)
```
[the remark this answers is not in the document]
richford: Yeah! I added to `s3fs_nifti_read` and `s3fs_json_read` and assumed that the write functions do not need an `anon` parameter.

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+        # The S3 API is paginated, returning up to 1000 keys at a time.
+        # Pass the continuation token into the next response, until we
+        # reach the final page (when this field is missing).
+        try:
+            kwargs['ContinuationToken'] = resp['NextContinuationToken']
+        except KeyError:
+            break
+
+
+def _download_from_s3(fname, bucket, key, overwrite=False, anon=True):
```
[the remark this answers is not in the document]
richford: Yep, that's better. I pushed a commit to that effect. Thanks!

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+        study : AFQ.data.S3BIDSStudy
+            The S3BIDSStudy for which this subject was a participant
+
+        site : str, optional
+            Site-ID for the site from which this subject's data was collected
+        """
+        super().__init__(
+            subject_id=subject_id,
+            study=study,
+            site=site
```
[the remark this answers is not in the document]
richford: Yeah, that's a better idea. See updated commits. Thanks!

AFQ/data.py
richford: Thanks. Updated in latest commits.

AFQ/data.py
richford: Added a parameter to download: `include_modality_agnostic` which can be `True`, `"all"`, `False`, or a subset of the [modality agnostic files described in the BIDS spec](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html). To achieve what you're talking about here, you'd use `include_modality_agnostic=["dataset_description.json"]`.

AFQ/data.py
richford: I had trouble getting the nesting to work with the Dask progress bar since we download multiple subjects all at once. So I went with this instead. I admit it's weird to mix the dask progress bar with the tqdm progress bar but I didn't want to spend too much time on that part. But by all means, commits to this PR are welcome.

AFQ/data.py
```diff
@@ -286,6 +293,1008 @@ def read_templates(resample_to=False):
...
+
+        Parameters
+        ----------
+        subject_id : str
+            Subject-ID for this subject
+
+        study : AFQ.data.S3BIDSStudy
+            The S3BIDSStudy for which this subject was a participant
+
+        site : str, optional
```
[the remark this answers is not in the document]
richford: I don't think this particular part is an HBN idiosyncrasy. Rather, I'm just trying to record some notion of the parent study object. On the other hand, the parameter `_subject_class` described on line 814 is an idiosyncrasy designed to support HBN, but I think I document that reasonably well there. What do you think?

richford:
Thanks @arokem and @36000 for the review! I rebased and responded to most of your suggestions, resolving the comments that I believe I responded to and leaving open the ones that we might want to keep talking about.

richford:
@36000, regarding the upload function. This would presuppose that the s3 bucket where we got the input data is the same one where we want to put the results. But for public datasets, this might not be true. Two options: **(1)** we could assume that it will work and then throw an error if it doesn't (or rather, let `s3fs` throw the permissions error). **(2)** we could put the upload functionality inside of the `myafq` object in your example. In either case, I think we should delay the upload functionality to another issue/PR so that this one can be concerned only with download.

arokem:
Indeed. This has moved into setup.cfg.

Could you please rebase onto master? You can sort the dependencies
alphabetically in there, if you want :-)

setup.cfg
```diff
@@ -73,3 +74,6 @@ plotly =
     plotly==4.9.0
     psutil==5.7.0
     kaleido
+
+[options.package_data]
+* = tests/data/*
```
arokem, line 79: Following the example in nibabel: https://github.com/nipy/nibabel/blob/master/setup.cfg#L82
```suggestion
AFQ = tests/data/*
```

arokem:
Well. This is annoying. Not as elegant, perhaps, but you can still add a `package_data` input to the `setup.py` call to `setup`: https://github.com/yeatmanlab/pyAFQ/blob/master/setup.py#L25

AFQ/data.py
```diff
@@ -288,6 +295,1039 @@ def read_templates(resample_to=False):
...
+
+        See Also
+        --------
+        AFQ.data.S3BIDSSubject.download
+        """
+        self._local_directories.append(directory)
+        self._local_directories = set(self._local_directories)
+
+        if include_modality_agnostic is True or include_modality_agnostic == "all":
+            _download_non_sub_keys(directory, select="all")
```
36000, line 1101: This should be self._download_non_sub_keys I believe, as well as the one below.

AFQ/data.py
```diff
@@ -288,6 +295,1039 @@ def read_templates(resample_to=False):
...
+                 include_dataset_description=True,
+                 include_derivs=False, overwrite=False, pbar=True):
+        """Download files for each subject in the study
+
+        Parameters
+        ----------
+        directory : str
+            Directory to which to download subject files
+
+        include_modality_agnostic : bool, "all" or any subset of ["dataset_description.json", "CHANGES", "README", "LICENSE"]
```
36000, line 1074: You should consider making include_modality_agnostic=["dataset_description.json"] the default, 
considering pyAFQ's API requires it.

AFQ/data.py
richford: Yep, thanks.

AFQ/data.py
richford: Done.

richford:
> Well. This is annoying. Not as elegant, perhaps, but you can still add a package_data input to the setup.py call to setup: https://github.com/yeatmanlab/pyAFQ/blob/master/setup.py#L25

I went with the `MANIFEST.in` solution, instead.

richford:
Yay! Tests finally pass. `MANIFEST.in` FTW.

arokem:
Great! I believe this is ready to go? Maybe we can merge this and follow up with issues/PRs on things that came up but were left unresolved?

richford:
Agreed. I think it's ready to go. I believe the only outstanding issue is the upload functionality that @36000 mentioned. However, I may have missed something else in the back-and-forth.

PR #291, opened 2020-07-02
Extra requirements feature added, consistent with current requirement files

36000:
setup.py now looks for files of the form `'requirements-*.txt' `and uses them for the EXTRAS_REQUIRE argument.
Currently allows: `pip install 'pyAFQ[dev]'`
Automatically finds more extra setups when new files are added.
For example, if `requirements-fury.txt` were added, this would find that file and make a `'pyAFQ[fury]'` extra setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ from setuptools import find_packages
 import string
 import os.path as op
 from setuptools_scm import get_version
+import glob
 
 try:
     from setuptools import setup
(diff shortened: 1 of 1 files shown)


36000:
Should there be an extra_requirements folder to put potential new requirements in?

36000:
@arokem what do you think of this setup? Should we do this or just replicate cloudknot i.e. get rid of requirement files.

setup.py
```diff
@@ -22,6 +23,20 @@
         REQUIRES.append(ll)
         ll = f.readline()[:-1]
 
+EXTRAS_REQUIRE = {}
+for extra_req_file in glob.glob(op.join(here, 'requirements-*.txt')):
+    extra_name = extra_req_file[
```
arokem, line 28:
```suggestion
    extra_name = extra_req_file.split('-')[1].split('.')[0]
```

arokem:
To answer your question: we can merge this for now. Let's keep in mind that we might want to eventually switch to a `setup.py`/`setup.cfg` format (see pybids, as an example of that).

PR #295, opened 2020-07-06
Remove six

36000:
Six is already in requirements.txt, with a different required version. Let's just remove the dev requirement since it is already in requirements.txt


arokem:
Could you please just remove it in both places? This is not supposed to work on Python 2

PR #302, opened 2020-07-14
RF: Speed up testing by moving viz test into API run.

arokem:
Currently, testing on the CI takes a long time (about 50 minutes, I believe). 

I believe that this will substantially shorten this, by testing the visualization in tandem with a run of the full processing pipeline.


arokem:
Any objections on your end, @36000?

PR #310, opened 2020-07-22
New CLI / Config

36000:
The AFQ.api.AFQ.__init__ docstring is now scraped to make a configuration file with all of the defaults and descriptions.
The scraping is done with: https://github.com/Chilipp/funcargparse

usage.rst can also be generated using updatedocs.py . If there were some way to call this automatically on server side for each commit, that could be cool.


36000:
Looks like Cli2Gui, which primarily uses pysimplegui, is getting pyqt installed, not because pysimplegui requires it but because getostheme requires it.

36000:
Add back in "The default values of the configuration are" thats in master.

36000:
Use toml instead of ini

36000:
Remove GUI comments

36000:
@arokem if these tests pass, this is ready for review

36000:
- [ ] Write a unit tests that makes sure the warnings of missing docstrings don't happen.
- [ ] Make sure lack of square brackets is handled as above.
- [ ]

arokem:
OK - this works great locally for me. 

Could you please: 

1. Rebase on master
2. Add `docs/source/usage.rst` to the .gitignore file? Since that one's generated automatically, we'd like to avoid tracking it.

arokem:
This is ready to go, I believe. I'll merge as soon as the CI comes back 🟢

PR #312, opened 2020-07-22
std_export function

### LINKED ISSUE #301: Standard export function?
Almost every time I use the pyAFQ api I have a block of function calls that looks like this:
```
    myafq.get_template_xform()
    myafq.export_bundles()
    myafq.get_tract_profiles()
    myafq.export_bundle_gif()
    myafq.export_ROI_gifs()
```
Maybe there should be one function, something like `myafq.std_exports()`, that just calls all of these? What should be in it?

36000:
This is a convenient function that should reduce boiler plate code.

Closes #301

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1541,3 +1541,14 @@ class AFQ(object):
         df = pd.concat(dfs)
         df.to_csv(op.join(self.afq_path, 'tract_profiles.csv'), index=False)
         return df
+
+    def std_export(self):
+        self.export_registered_b0()
+        self.get_template_xform()
+        self.export_bundles()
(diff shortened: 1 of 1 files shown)


AFQ/api.py
```diff
@@ -1543,3 +1543,14 @@ def combine_profiles(self):
         df = pd.concat(dfs)
         df.to_csv(op.join(self.afq_path, 'tract_profiles.csv'), index=False)
         return df
+
+    def std_export(self):
```
arokem, line 1547:
```suggestion
    def export_all(self):
        """ Exports all the possible outputs"""
```

PR #317, opened 2020-07-24
Compare profiles from CSVs

### LINKED ISSUE #313: Compare CSVs
There should be a viz function whose inputs are two csv's and which generates an image similar to that created by plot_tract_profiles but with two lines in each plot. It should also output the percent differences, possibly beneath each plot.

36000:
Closes #313 .

Also makes a combine_profiles function outside of the API. You can simply give it a folder, and it will search that folder / subdirectories for csv files that follow our format for profiles. It combine those files to a single profile. It will be useful when using cloudknot where you process each subject separately then merge at the end.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -3,7 +3,6 @@ import pandas as pd
 import dask.dataframe as ddf
 import os
 import os.path as op
-from pathlib import Path
 import json
 
 import numpy as np
@@ -1749,9 +1748,16 @@ class AFQ(object):
         self.data_frame.apply(self._export_registered_b0, axis=1)
 
     def combine_profiles(self):
-        return combine_profiles_in_folder(
-            self.afq_path,
-            op.join(self.afq_path, 'tract_profiles_combined.csv'))
+        dfs = []
+        for ii, fname in enumerate(self.tract_profiles):
+            profiles = pd.read_csv(fname)
+            profiles['sub'] = self.data_frame['subject'].iloc[ii]
+            profiles['ses'] = op.split(self.data_frame['ses'].iloc[ii])[-1]
+            dfs.append(profiles)
+
+        df = pd.concat(dfs)
+        df.to_csv(op.join(self.afq_path, 'tract_profiles.csv'), index=False)
+        return df
 
     def export_all(self):
         """ Exports all the possible outputs"""
@@ -1764,40 +1770,3 @@ class AFQ(object):
             self.viz_ROIs(export_as_gif=True)
             self.export_rois()
         self.combine_profiles()
-
-
-def combine_profiles_in_folder(tract_profile_folder, outfile):
-    """
-    Combine tract profiles from different subjects / sessions
-    into one CSV.
-
-    Parameters
-    ----------
-    tract_profile_folder : path to folder
-        Folder with all tract profiles you wish to merge.
-        Will also search subdirectories recursively.
-        It will select all files of the form '*profiles.csv'.
-    outfile : filename
-        Filename for the output CSV.
-
-    Returns
-    -------
-    Ouput CSV's pandas dataframe.
-    """
-    dfs = []
-    for path in Path(tract_profile_folder).rglob('*profiles.csv'):
-        profiles = pd.read_csv(path)
-
-        subject_name = path.name.split('sub-')[1].split('_')[0]
-        if 'ses-' in path.name:
-            session_name = path.name.split('ses-')[1].split('_')[0]
-        else:
-            session_name = 'unknown'
-
-        profiles['subjectID'] = subject_name
-        profiles['sessionID'] = session_name
-        dfs.append(profiles)
-
-    df = pd.concat(dfs)
-    df.to_csv(outfile, index=False)
-    return df
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1753,8 +1753,8 @@ class AFQ(object):
         dfs = []
         for ii, fname in enumerate(self.tract_profiles):
             profiles = pd.read_csv(fname)
-            profiles['sub'] = self.data_frame['subject'].iloc[ii]
-            profiles['ses'] = op.split(self.data_frame['ses'].iloc[ii])[-1]
+            profiles['subjectID'] = self.data_frame['subject'].iloc[ii]
+            profiles['sessionID'] = op.split(self.data_frame['ses'].iloc[ii])[-1]
             dfs.append(profiles)
 
         df = pd.concat(dfs)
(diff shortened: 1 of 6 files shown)


36000:
@arokem  this should be ready to review, but its not time prescient so you can wait until after neurohackademy .

36000:
Use std error for uncertainties. Use subject std errs for uncertainties on bottom plot.

AFQ/api.py
```diff
@@ -1734,3 +1726,40 @@ def export_all(self):
...
+        It will select all files of the form '*profiles.csv'.
+    outfile : filename
+        Filename for the output CSV.
+
+    Returns
+    -------
+    Ouput CSV's pandas dataframe.
+    """
+    dfs = []
+    for path in Path(tract_profile_folder).rglob('*profiles.csv'):
```
arokem, line 1750: Maybe instead of globing, we could query the BIDS dataset here and infer the well-defined file-names that should be there?

36000:
Rename to LongitudinalCSVComparison

36000:
Add contrast index profiles

36000:
lateral contrast indices (ie ATR_L vs ATR_R)

36000:
@arokem I remove the combine profiles outside of API feature that this PR had, see explanation in comment above. The rest is ready for review!

AFQ/api.py
36000: I decided to remove this all together, see below.

AFQ/viz/utils.py
```diff
@@ -411,6 +434,615 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+        message = (
+            'NaNs found in scalar ' + scalar
+            + ' for subject ' + str(subject)
+            + ' for bundle ' + bundle
+            + ' for CSV ' + name)
+        if repl_nan:
+            message = message + '. These Nans were replaced with 0.'
+        self.logger.info(message)
+
+    def _warn_many_nans(self, scalar, subject, bundle, name):
```
arokem, line 568: Wouldn't it make sense to have one logger that tells you how many nodes have nans?

AFQ/viz/utils.py
```diff
@@ -411,6 +434,615 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+            return None
+
+        if np.sum(nans) > 10:
+            self._warn_many_nans(scalar, subject, bundle, name)
+            return None
+
+        if np.sum(nans) > 0:
+            self._warn_nans(scalar, subject, bundle, name, repl_nan)
+            if repl_nan:
+                single_profile[nans] = 0
```
arokem, line 600: IIUC, this means that we impute to 0. Another (maybe more careful?) approach would be to interpolate from the closest nodes, if this is in the middle of a profile, or extrapolate the last value out, if this is at the end of a profile.

arokem: We can't use nan-agnostic functions like nanmean or nanstd here, right? Because you need a non-nan value to calculate correlations.

AFQ/viz/utils.py
```diff
@@ -411,6 +434,615 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+            for bundle in bundles:
+                profiles = [None] * 2
+                both_found = True
+                for i, name in enumerate(names):
+                    profiles[i] = self._get_profile(
+                        name, bundle, subject, scalar)
+                    if profiles[i] is None:
+                        both_found = False
+                if both_found:
+                    this_contrast_index = \
```
arokem, line 743: Maybe we should make the calculation of a contrast index a utility function that can be reused elsewhere? Maybe put it in the utils module, rather than in here?

AFQ/viz/utils.py
```diff
@@ -411,6 +434,615 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+            f"{names[0]}_vs_{names[1]}"))
+        if not show_plots:
+            plt.ion()
+        return contrast_index
+
+    def lateral_contrast_index(self, names=None, scalar="dti_fa",
+                               bundles=list(POSITIONS.keys()),
+                               show_plots=False):
+        """
+        Calculate the contrast index for each bundle in two datasets.
```
arokem, line 770: Not clear from the docstring how this differs from the previous function

AFQ/viz/utils.py
36000: The distinction for me is that a few NaNs is sent to logger.info while many NaNs is sent to logger.warning and the profile is ignored. I could make that distinction clearer in the error message. Some profiles are all / nearly all NaNs, and we should ignore those, while some have a few NaNs at the edges, which is less serious. Most of this has to do with the profiles coming out of matlab AFQ.

AFQ/viz/utils.py
36000: If the profiles were only NaN's, they would not have made it this far. NaN agnostic functions calculate the mean using non-NaN values (ignore the NaNs) while the normal numpy functions simply propogates the NaNs.

AFQ/viz/utils.py
arokem: OK - that makes sense. Maybe we should make this a percentage and allow users to control that percentage? I don't like the hard-coding here. And we can still report what the percentage of nan nodes was.

AFQ/viz/utils.py
36000: Where does the contrast index come from? I am thinking some function like this should already exist in numpy or scipy, but I cannot find much information about it online. Is there another name for it?

AFQ/utils/stats.py
```diff
@@ -0,0 +1,7 @@
+def contrast_index(x1, x2):
+    """
+    Calculate the contrast index between two arrays.
+    The contrast index is symmetrical with respect to
+    the choice of the baseline for comparison
+    """
+    return (x1 - x2) / (x1 + x2)
```
arokem, line 7: To answer your question about this: I have not seen this implemented elsewhere. This is a good place to put it! But please add a docstring and some unit testing, please.

AFQ/viz/utils.py
```diff
@@ -411,6 +435,610 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+        """
+        Load in csv files, converting from matlab if necessary.
+
+        Parameters
+        ----------
+        out_folder : path, optional
+            Folder where outputs of this class's methods will be saved.
+
+        csv_fnames : list of filenames
+            Filenames for the two CSVs containing tract porfiles to compare.
```
arokem, line 462:
```suggestion
            Filenames for the two CSVs containing tract profiles to compare.
```

PR #322, opened 2020-07-30
b value selection fix and test, power map test, models moved to own submodule

36000:
I found a bug in the b value selection code that is now fixed and tested for.


AFQ/tests/test_api.py
```diff
@@ -168,6 +168,25 @@ def test_AFQ_data():
         myafq.export_rois()
 
 
+def test_AFQ_anisotropic():
+    """
+    Test if API can run without prealign
```
arokem, line 173: Docstring needs an update

AFQ/tests/test_api.py
```diff
@@ -168,6 +168,25 @@ def test_AFQ_data():
...
+    myafq = api.AFQ(
+        bids_path=bids_path,
+        dmriprep='vistasoft',
+        segmentation='freesurfer',
+        min_bval=1990,
+        max_bval=2010,
+        b0_threshold=50,
+        reg_template="mni_T1",
+        reg_subject="power_map")
+    myafq.export_rois()
```
arokem, line 187: Can we do something more specific here to check that the selection works as expected? I guess we can call individual (private) methods of this object once it’s initialized.

arokem, line 187: Can we do something more specific here to check that the selection works as expected? I guess we can call individual (private) methods of this object once it’s initialized.

36000: Is this similar to what you had in mind?

arokem: Can we maybe check that the APM is where we'd expect it to be at this point?

36000: I am now checking that the APM map is roughly similar (rounded to the ten's place) to the map in file_dict. Is there a better way to check the mapping?

36000:
Our tests officially take longer than an hour now!  :tada:

36000:
move apm calculations into apm.py

arokem:
Great. I think this is now ready to merge. I'll just wait for the CI to finish and then go ahead.

PR #320, opened 2020-07-26
Adds a setup.cfg file and cleans up metadata and other hacks

arokem:
Hopefully addresses #306 along the way.

PR #323, opened 2020-07-31
Fix broken import of version in the API module

arokem:
Attempting to fix the breakage introduced into master with the merging of #320


arokem:
Made obsolete by #324

PR #324, opened 2020-07-31
Revert "Adds a setup.cfg file and cleans up metadata and other hacks"

arokem:
Reverts yeatmanlab/pyAFQ#320


arokem:
This will revert the merge of #320, because that seems to have broken master.

PR #326, opened 2020-07-31
Setup with config redux ... redux

arokem:
OK  - one more time. Rebasing on master pulls in #324, so that's not a good idea


arokem:
Oh halleluya! I don't know why this didn't work in the previous PR, but now it seems to work. 

Anyone want to take a look, just to see that I am not doing anything silly here?

.github/workflows/docbuild.yml
```diff
@@ -18,10 +18,10 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install 
+    - name: Install
       run: |
         python -m pip install --upgrade pip
-        pip install -e '.[dev, fury]'
```
36000, line 24: The '-e' used to be necessary but I am glad it works without it now.

setup.cfg
```diff
@@ -0,0 +1,75 @@
...
+install_requires =
+    scikit_image==0.16.2
+    nibabel==3.0.2
+    dipy==1.1.1
+    scipy==1.2.0
+    numpy==1.17.5
+    pandas==1.0.3
+    joblib==0.12.4
+    dask==1.1.0
+    palettable==3.1.1
```
36000, line 37: Does this mean we should remove requirements.txt and requirements-*.txt, or that we need to update both when we want to change our requirements?

arokem: Let's see. I just pushed a commit (c46b4c5) where I remove all the requirements files. I've seen other projects (e.g., nibabel) that have both the requirements files and a setup.cfg with all the requirements spelled out, so I wonder.

arokem: Well, strangely, this installs just fine, but now can't import. I'll roll that change back. So, yes: I think that we need to remember to update in both places. At least until we can figure out why this duplication is needed.

arokem:
So strange. I don't understand why this worked in b38cf0c and now fails in 9423d4f

I am going to do an interactive rebase to get rid of these last two commits, and see whether order is restored.

arokem:
Oh! Looks like this https://github.com/yeatmanlab/pyAFQ/pull/326/commits/c46b4c5c490108bdbd06a73422752ee47875f759#diff-fe8421955fd596131bb6f1b78984b2fbR24 also snuck into the commit where I removed the requirements files. Looks like even with this setup we might still need to install with `-e` -- looks like some kind of interaction with the scm_setuptools versioning. Back to the drawing board...

arokem:
OK - I think I (finally!) understand what happened here. I'll write it down so that I can look it up when I inevitably run into this again... When we install with the `-e` flag, a 'version.py' file is created within the source tree and placed in 'AFQ/version.py', but when we install the regular way the file is created but moved away from the source tree to the installation location. This interacted with the fact that when py.test is called in the source tree, it imports the code that is in the source tree, not the code that is installed in system (this is just a quirk of how python imports things). Taken together, the version was not found unless a `-e` installation was done, or unless (the solution I eventually implemented), testing was done in a completely different place in the file system.

PR #329, opened 2020-08-05
Make power maps the default for registration

36000:
It's possible this will find more bundles, requiring tests to be updated.


arokem:
Super. Thank you!

PR #330, opened 2020-08-05
Generating a wm mask out of multiple scalars

36000:


AFQ/api.py
```diff
@@ -157,7 +157,7 @@ def __init__(self,
                  dask_it=False,
                  force_recompute=False,
                  scalars=["dti_fa", "dti_md"],
-                 wm_criterion=0.1,
+                 wm_criterion={"dti_fa": ">0.1", "dti_md": "<0.003"},
```
arokem, line 160: Usually, when I find myself trying to write code in strings, to be parsed by something else, it means that something is overly (or underly) complicated. 

I think that there are two (possibly) more robust ways to do this: 

1. Expand the dictionary a bit to have: `{'lb':{'dti_fa': 0.1}, 'ub': {'dti_md':0.0003}}`. One advantage of that is that you can choose ranges. 

2. Use two key-word arguments: 

```
wm_criterion_lb={'dti_fa':0.1}
wm_criterion_ub={'dti_md':0.003} 
```

36000: I agree that code in strings should be avoided. I like the first option because it keeps everything in the same argument, which could also be a list if you want to use labels.

AFQ/api.py
```diff
@@ -854,12 +861,43 @@ def _wm_mask(self, row):
...
+                        if wm_mask is None:
+                            wm_mask = new_wm_mask
+                        else:
+                            wm_mask = np.logical_and(wm_mask, new_wm_mask)
+                        bounds.append(
+                            scalar_fname
+                            + for_meta_bound_type
+                            + str(threshold))
+
+                meta = dict(bounds=bounds)
```
arokem, line 900: Instead of creating strings and the `for_meta_bound_type` variable, let's simplify this by using the `self.wm_criterion` as the metadata dict. Or am I missing something?

36000: We previously outputted a filename and a threshold, so I was trying to preserve that. but I think outputting self.wm_criterion and a list of filenames might make the most sense.

arokem: Yep - this looks good!

36000:
I think the documentation error here was just a connection problem.

arokem:
I have a button that allows me to rerun builds. Maybe you see that one too?

arokem:
Is this ready? The PR still says "WIP", so I figured I'd ask before merging.

36000:
Nope! Right now I am using it on the prek data and I am finding that there is still overlap between the resulting wm mask and high values in MD. I marked this as WIP when I realized this was happening! I am investigating this bug and I will let you know when I think this works as intended.

36000:
I think I have located the problem but I would like it if @arokem could make sure my understanding of this is correct.
We use this line on our wm_masks:
`wm_mask = binary_dilation(wm_mask) > 0`
When we were getting wm masks from segmentation files / an arbitrary FA cutoff this made sure we were not being too strict. But now that we are trying to make specific masks we should not be doing this. 
Should we do this for seg files but not wm masks based off of scalars? Or should it be based on user input?

arokem:
Oh - good spot! I think that this was a precaution against overly conservative masks, but I think that's really not an issue anymore.  Given everything we've seen in the last couple of weeks, I am starting to think that maybe there is never a reason to use binary dilation here. And as you've demonstrated, it can lead to some puzzling outcomes. Maybe we should remove it altogether?

36000:
Because of this interpolation trickiness, I may just revert to using an FA mask for tracking to replicate mAFQ, but then mask out any nodes with high MD after profiling.

PR #331, opened 2020-08-05
b0_threshold default updated

36000:


arokem:
Looks good, but let's let the CI run, in case something comes up.

Issue #334, opened 2020-08-07
Add particle filtering tractography as an option


arokem:
See: https://www.sciencedirect.com/science/article/abs/pii/S1053811914003541

I think this should be a matter of importing the `ParticleFilteringTracking` object into the tractography module and then 
allowing to switch between `LocalTracking` and `ParticleFilteringTracking`

https://dipy.org/documentation/1.1.1./examples_built/tracking_pft/#example-tracking-pft

arokem:
Closed through #412 and #335

PR #335, opened 2020-08-07
Adds ParticleFilteringTractography.

arokem:
Addresses #334. For now, I have done the naive thing of assuming that these objects are duck-typing. From my local testing, it looks like it's going to be a bit more complicated than that. We might need to do some more work in the `_track` function to be able to use both objects.

### DIFF
--- a/AFQ/tractography.py
+++ b/AFQ/tractography.py
@@ -11,12 +11,13 @@ from dipy.io.stateful_tractogram import StatefulTractogram, Space
 from dipy.tracking.stopping_criterion import ThresholdStoppingCriterion
 
 from AFQ._fixes import VerboseLocalTracking, tensor_odf
-
+from dipy.tracking.local_tracking import ParticleFilteringTracking
 
 def track(params_file, directions="det", max_angle=30., sphere=None,
           seed_mask=None, seed_threshold=0, n_seeds=1, random_seeds=False,
           rng_seed=None, stop_mask=None, stop_threshold=0, step_size=0.5,
-          min_length=10, max_length=1000, odf_model="DTI"):
+          min_length=10, max_length=1000, odf_model="DTI",
+          tracker="local"):
     """
     Tractography
 
@@ -48,7 +49,7 @@ def track(params_file, directions="det", max_angle=30., sphere=None,
         to generate within the mask.
     random_seeds : bool
         Whether to generate a total of n_seeds random seeds in the mask.
-        Default: XXX.
+        Default: False.
     rng_seed : int
         random seed used to generate random seeds if random_seeds is
         set to True. Default: None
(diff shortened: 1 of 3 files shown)


arokem:
This does not currently work, because PFT requires using an `AnatomicalStoppingCriterion` class as its stopping criterion. Because this interacts with work happening in #330, I will wait to merge that one and then revisit. For now, I am marking this one for 0.5, so we don't have to worry about it for the 0.4 milestone.

arokem:
I have removed the WIP label. I think that this is ready for review. 

Related to #370 : for PFT, this now requires a string as the `stop_mask`: either "ACT" or "CMC" for the strategy to use in determining where to stop and a tuple as the `stop_threshold` that is to be interpreted as three volumes: `(pve_wm, pve_gm, pve_csf)`. 

And a `tracker` input is added, which defaults to `"local"`. When `tracker="pft"`, `stop_mask` needs to be a string and `stop_threshold` needs to be a tuple.

arokem:
In describing this, I can already see that better error-handling would be good.

36000:
does this mean we should get rid of the wm_mask parameter altogether?

36000:
Where does the user get pve_wm, pve_gm, pve_csf data from? It seems to me like there should be a special PFTmask where you give it the bids suffix for each of these files, and get_mask returns these in its tuple instead of what it currently returns. 
And then maybe stop_threshold should be "ACT" or "CMC" instead.

36000:
So I've been thinking about this, and this is what I think the ui should look like (for example):
seed_mask = (afm.MaskFile(f"WM_probseg"), afm.MaskFile(f"GM_probseg"), afm.MaskFile(f"CSF_probseg"))
stop_threshold = "ACT"

Then the api finds the relevant mask for each subject.

AFQ/tractography.py
```diff
@@ -117,38 +141,58 @@ def track(params_file, directions="det", max_angle=30., sphere=None,
...
+        pve_wm_data, pve_gm_data, pve_csf_data = stop_threshold
+        my_tracker = VerboseParticleFilteringTracking        
+        if stop_mask == "CMC":
+            average_voxel_size = np.mean(params_img.header.get_zooms()[:3])
+            stopping_criterion = CmcStoppingCriterion.from_pve(
+                pve_wm_data,
+                pve_gm_data,
+                pve_csf_data,
+                step_size=step_size,
+                average_voxel_size=average_voxel_size)
```
arokem, line 166: I now realize that this needs to be the voxel size of the PVE maps. Need to think how we get that in there.

36000: Maybe instead of a tuple, it's its own class:

seed_mask = PFTInfo(afm.MaskFile(f"WM_probseg"), afm.MaskFile(f"GM_probseg"), afm.MaskFile(f"CSF_probseg"), "ACT")

The API would interact with it normally, but the track function would recognize it and access parameters like average_voxel_size which would have been calculated when the masks were loaded.

arokem:
Would love to have #380 merged before moving on here, so I am switching to that one for now.

36000:
I have been thinking about this, and I think all of my suggestions should be implemented and handled in the api. the track function should not use mask objects, only the API. This can be another PR.

So I think the only outstanding issue on this PR is how to handle average_voxel_size

arokem:
Agreed! I'll try to get that done today.

arokem:
Looks like I actually did that at some point on my other laptop, but forgot to push :-)

I think this is done?

36000:
LGTM!

PR #337, opened 2020-08-08
Remove this section of the pyproject.

arokem:
The hope is that the local version will be picked up from setup.py instead.

This should resolve the current breakage on master.


arokem:
Only one way to see whether this solves the problem... merging.

If this doesn't work, we could consider creating a git tag with every merge, and using that to create a valid version to push into test pypi, instead of the current hack.

PR #342, opened 2020-08-10
Update versions of scipy and vtk.

arokem:


arokem:
Looks like this is OK. I will go ahead and merge it, because I want to try to run #310 locally, and can't do that without first merging this...

PR #352, opened 2020-08-12
Allow for selecting subject(s) by position after randomization

36000:
This will be useful with cloudknot. For example, imagine I want to use cloudknout to process 3 random subjects. Each process can use the same random seed, but be told to use either the first, second, or third subject out of the randomized list of subjects generated from the same seed.

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -792,8 +792,9 @@ class S3BIDSStudy:
         subjects : str, sequence(str), int, or None
             If int, retrieve S3 keys for the first `subjects` subjects.
             If "all", retrieve all subjects. If str or sequence of
-            strings, retrieve S3 keys for the specified subjects. If
-            None, retrieve S3 keys for the first subject. Default: None
+            strings, retrieve S3 keys for the specified subjects. If sequence
+            of ints, then for each int n retrieve S3 keys for the nth subject.
+            If None, retrieve S3 keys for the first subject. Default: None
 
         anon : bool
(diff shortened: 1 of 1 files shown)


36000:
@richford what do you think of this? I don't want to overcomplicate the function so I can put this elsewhere if you like.

AFQ/data.py
```diff
@@ -862,13 +864,21 @@ class that quacks like AFQ.data.S3BIDSSubject. Default:
...
+        if subjects is None or isinstance(subjects, int) \
+                or (isinstance(subjects, list)
+                    and isinstance(subjects[0], int)):
             # if subjects is an int, get that many random subjects
-            n_subs = subjects if subjects is not None else 1
             prng = np.random.RandomState(random_seed)
-            subjects = prng.choice(sorted(self._all_subjects),
-                                   size=subjects,
-                                   replace=False)
+            randomized_subjects = self._all_subjects.copy()
```
richford, line 872:
```suggestion
            randomized_subjects = sorted(self._all_subjects.copy())
```

Suggest sorting this since I'm not positive that the matching S3 keys will always be returned in the same order.

36000: Ah I see. good catch!

richford:
Thanks @36000 for sorting `_all_subjects`. IMO, this is ready to go now...you or @arokem can merge after tests pass! 🚀

arokem:
Now you can too! :shipit:

Issue #353, opened 2020-08-12
CLI with default_config gets multiple values for `params_file`


arokem:
I edited the default config file only to point to the BIDS directory where the data is living (this is still the data from #261). I am getting:

```
(afq) ➜  pyAFQ git:(master) ✗ pyAFQ default_config.toml
Loading AFQ libraries...
INFO:pyAFQ:Your use of pyAFQ is being recorded using Google Analytics. For more information, please refer to the pyAFQ documentation: https://yeatmanlab.github.io/pyAFQ/usage.html#usage-tracking-with-google-analytics. To turn this off, use the `--notrack` flag when using the pyAFQ CLI
INFO:dipy.data.fetcher:Dataset is already in place. If you want to fetch it again please first remove the folder /Users/arokem/AFQ_data/templates 
INFO:dipy.data.fetcher:Dataset is already in place. If you want to fetch it again please first remove the folder /Users/arokem/AFQ_data/callosum_templates 
/Users/arokem/miniconda3/envs/afq/lib/python3.8/site-packages/bids/layout/models.py:98: FutureWarning: The 'extension' entity currently excludes the leading dot ('.'). As of version 0.14.0, it will include the leading dot. To suppress this warning and include the leading dot, use `bids.config.set_option('extension_initial_dot', True)`.
  warnings.warn("The 'extension' entity currently excludes the leading dot ('.'). "
INFO:AFQ.api:Saving /Users/arokem/AFQ_data/duanliangtao2/derivatives/afq/sub-01/ses-01/sub-01_ses-01_dwi_b0.nii.gz
INFO:AFQ.api:Saving /Users/arokem/AFQ_data/duanliangtao2/derivatives/afq/sub-01/ses-01/sub-01_ses-01_dwi_brain_mask.nii.gz
INFO:AFQ.api:Saving /Users/arokem/AFQ_data/duanliangtao2/derivatives/afq/sub-01/ses-01/sub-01_ses-01_dwi_model-DTI_diffmodel.nii.gz
INFO:AFQ.api:Saving /Users/arokem/AFQ_data/duanliangtao2/derivatives/afq/sub-01/ses-01/sub-01_ses-01_dwi_model-DTI_desc-DEC_FA.nii.gz
INFO:AFQ.api:Saving /Users/arokem/AFQ_data/duanliangtao2/derivatives/afq/sub-01/ses-01/sub-01_ses-01_dwi_model-DTI_PDD.nii.gz
[...]

PR #354, opened 2020-08-13
Ignore positional arguments in non AFQ functions for docstring parser, add test

36000:


arokem:
I'll merge this and give the example in #353 another run here.

Issue #355, opened 2020-08-13
Speed up unit tests.


36000:
We could reduce the number of seeds / ROIs in some key tests. Maybe mark other tests as slow. Any ideas are welcome.

36000:
I wonder if we could further parallelize the tests, like how python versions 3.6 and 3.7 are already done in parallel.

PR #356, opened 2020-08-13
Added test that runs full cli pipeline on default config file. Tell CI to not run the tests marked slow.

36000:
Even with only 500 sls it takes ~20 min to complete. Marked slow. Still useful to run if we are messing with the automatic CLI. Passes locally, ready to merge.


36000:
@arokem thoughts on this test? It's marked as slow so it won't normally run. It adds 20 minutes. Ran locally and passed.

36000:
Oh- I see know why our test time is ballooning. It does not seem like the pytest on the CI recognizes the slow marks, and is running all tests (even the slow ones). Good news is we have been passing even the slow tests! I will fix this on this PR.

arokem:
Are you sure the slow tests are getting skipped? This has been running for almost 2 hours, and I don't see any skips here, for example: https://github.com/yeatmanlab/pyAFQ/runs/981999945. Should be an `s` instead of those `.` for skipped tests.

In general, I think that our strategy should be to run only fast tests on PRs and such and then set up a [cron job](https://jasonet.co/posts/scheduled-actions/) to run all of the tests every night at 2 AM, or some-such. That way, we'll at least be alerted whenever something funky makes it into the `master` branch.

arokem:
Jinx!

36000:
@arokem I am not quite sure I understand the directory structure on these test suites. It is having trouble finding the ini file I made.

arokem:
Could you please try putting this in the `pyproject.toml` file instead? https://docs.pytest.org/en/stable/customize.html#pyproject-toml 

We have a proliferation of configuration files 😄 . This might also resolve the issue of where to put this.

36000:
I tried in previous commits to put it there, but it doesn’t seem to recognize that either.

arokem:
I wonder whether we should use the information on [this page](https://docs.pytest.org/en/stable/example/markers.html) to add something to the the CLI call to pytest [here](https://github.com/yeatmanlab/pyAFQ/blob/master/.github/workflows/test.yml#L31)

arokem:
Judging from how long this is taking, doesn't seem to have done the trick 😞

36000:
The weird thing is, everything I try works locally. I wonder if the configuration file needs to be somewhere weird because of where we are running pytest from? Like it is not picking up on the configuration file?

36000:
Hopefully one of these three will do the trick.

36000:
@arokem  this should work in ~1h10m and does not run tests marked slow. I think we can reduce this time in other PRs like #364 as described in #355 . For now, this should be good to go.

Issue #349, opened 2020-08-12
Visualization with plotly hangs


arokem:
I am reanalyzing the data from #261. 

The first comment I have is that the issues raised there seem to be resolved on master, which is great!

This issue has to do with the observation I made along the way that plotly seems to hang or get stuck while visualizing. I am not sure what the cause is, but it doesn't seem to be random: I tried running this a few times and it always seems to hang while visualizing the second bundle it encounters (ATR_L).

Hard-quitting the program suggests that this is stuck while executing this line: https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/viz/plotly_backend.py#L245

36000:
Interesting. I would guess this has something to do with kaleido. I wonder if a recent version broke it: https://pypi.org/project/kaleido/#history

Or maybe it has something to do with this: https://github.com/plotly/Kaleido/issues/23
Try printing the working directory each loop and see if it changes.

arokem:
Still haven't figured this one out, but noting that pinning kaleido to 0.0.2 does not resolve this issue.

36000:
I am going to try printing the working directory each loop right now.

36000:
Working directory change does not seem to be the problem. I hung in the 5th frame of the first gif it tried to make.

36000:
So I did it a little test. So, for each gif, write_image is called 60 times (once per frame). It looks like the time this call takes (time to export ) starts off high, but bearable. However, that time steadily increases untill the program stalls. I think I should setup a contained example of this and post an issue to kaleido. For this I would need a saved figure that could be posted publicly.
```
Time to export: 4.486781587824225
Time to export: 3.5240664649754763
Time to export: 3.313969453331083
Time to export: 3.2749502579681575
Time to export: 3.324021595995873
Time to export: 3.2209196309559047
Time to export: 3.303899036720395
Time to export: 3.442790257744491
Time to export: 3.4185144901275635
Time to export: 3.4022287311963737
Time to export: 3.4128747819922864
Time to export: 3.3995601208880544
Time to export: 3.518615571781993
Time to export: 3.4847019966691732
Time to export: 3.548639959190041
Time to export: 3.70137710403651
Time to export: 3.611623720731586
Time to export: 3.740179928019643
Time to export: 3.74419933790341
Time to export: 3.8939710282720625
Time to export: 3.7193696573376656
Time to export: 3.8059598891995847
Time to export: 3.9091362422332168
[...]

PR #361, opened 2020-08-13
Export what is natural to the viz library

36000:
Making gifs using plotly instead of fury has always been slow. Plotly is better at making interactive html. That said, it should be possible. For 0.4, maybe we should warn users to not use plotly to make gifs (like in this PR) until we get to the bottom of why kaleido is taking so long (see: https://github.com/yeatmanlab/pyAFQ/issues/349#issuecomment-673738530). We could also throw an error.


arokem:
LGTM!

PR #363, opened 2020-08-14
Throw error when backend is misnamed

### LINKED ISSUE #362: Add error handling for viz backend typos
#361 reminded me to check this. Maybe we should add a final else here: https://github.com/yeatmanlab/pyAFQ/blob/1259fb1c511d649bb7e7700363abd530bdde247c/AFQ/viz/utils.py#L371 that raises a nice explanatory error if you typo'd the name of the backend.

36000:
Closes #362

### DIFF
--- a/AFQ/tests/test_viz.py
+++ b/AFQ/tests/test_viz.py
@@ -1,12 +1,14 @@
+import pytest
+
 from AFQ.viz.utils import Viz
 
 
 def test_viz_name_errors():
     Viz("fury")
 
-    try:
+    with pytest.raises(
+        TypeError,
+        match="Visualization backend should be"
+        + " either 'plotly' or 'fury'. "
+            + "It is currently set to plotlyy"):
         Viz("plotlyy")
-    except TypeError:
-        pass
-    else:
-        AssertionError("Misnamed viz backend did not throw type error")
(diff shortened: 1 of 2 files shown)


AFQ/viz/utils.py
```diff
@@ -368,6 +368,9 @@ def __init__(self,
             self.visualize_roi = AFQ.viz.plotly_backend.visualize_roi
             self.visualize_volume = AFQ.viz.plotly_backend.visualize_volume
             self.create_gif = AFQ.viz.plotly_backend.create_gif
+        else:
+            TypeError("Visualization backend should be"
+                      + " either 'plotly' or 'fury'")
```
arokem, line 373: Always a good idea to tell them what the value was at that moment. In case the user wasn't the one setting the value themselves.

```suggestion
                      + " either 'plotly' or 'fury'. It is currently set to %s"%self.viz.backend)
```

arokem:
Also, adding a test that `__init__` indeed raises an error when you enter "plotlyy" or somesuch would be good.

AFQ/tests/test_viz.py
```diff
@@ -0,0 +1,12 @@
+from AFQ.viz.utils import Viz
+
+
+def test_viz_name_errors():
+    Viz("fury")
+
+    try:
```
arokem, line 7: Pytest has a construct for this: https://stackoverflow.com/a/29855337

PR #364, opened 2020-08-17
Allow user to seed tractography with ROIs via api. Use this to reduce test times. Mark some tests as nightly.

### LINKED ISSUE #231: Slow tests
Maybe mark some tests as slow for fast runs of the tests? (using `@pytest.mark.slow`). 

And set up a CRON job to run the slow tests nightly.

### LINKED ISSUE #355: Speed up unit tests.

### LINKED ISSUE #358: API should be to make "ROI mask"
Currently the api can detect when you pass in, for example, "dti_fa" as your seed mask, and then replace that string with the actual dti_fa. It should do the same when "roi" is passed in, i.e., it should replace that string with a mask which is all the ROIs or'd together. Seeding by ROI could reduce the number of seeds needed in tests, reducing test time.

36000:
Closes #358 . Adds nightly test and marks some tests as slow. Closes #355 . Closes #231 .


36000:
@arokem ready to review / merge. Tests running at ~15 min!

arokem:
Alright! Looks good.

Issue #346, opened 2020-08-12
TOML does not handle dictionaries well


arokem:
So, this line in the default config: 

```
wm_criterion = {'lb': {'dti_fa': 0.2}, 'ub': {'dti_md': 0.002}}
```

Throws: 

```
(afq) pyAFQ (master)$pyAFQ AFQ/tests/test_config.toml 
Loading AFQ libraries...
INFO:pyAFQ:Your use of pyAFQ is being recorded using Google Analytics. For more information, please refer to the pyAFQ documentation: https://yeatmanlab.github.io/pyAFQ/usage.html#usage-tracking-with-google-analytics. To turn this off, use the `--notrack` flag when using the pyAFQ CLI
Traceback (most recent call last):
  File "/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/toml/decoder.py", line 607, in load_inline_object
    _, value = candidate_group.split('=', 1)
ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/toml/decoder.py", line 456, in loads
    multibackslash)
  File "/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/toml/decoder.py", line 725, in load_line
    value, vtype = self.load_value(pair[1], strictly_valid)
  File "/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/toml/decoder.py", line 805, in load_value
    self.load_inline_object(v, inline_object)
  File "/Users/arokem/.virtualenvs/afq/lib/python3.7/site-packages/toml/decoder.py", line 609, in load_inline_object
    raise ValueError("Invalid inline table encountered")
ValueError: Invalid inline table encountered

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/arokem/.virtualenvs/afq/bin/pyAFQ", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/arokem/source/pyAFQ/bin/pyAFQ", line 186, in <module>
    parse_config_run_afq(config_file, arg_dict, overwrite)
  File "/Users/arokem/source/pyAFQ/bin/pyAFQ", line 104, in parse_config_run_afq
    f_arg_dict = toml.load(toml_file)
[...]

PR #370, opened 2020-08-18
[WIP] Overhaul Mask File UI

36000:
TODO:

- [ ] Add get_mask for ThresholdedMaskFile
- [ ] Add ThresholdedScalarMask, may have similar functionality with ThresholdedMaskFile
- [ ] Integrate these classes with existing API, updating docs
- [ ] Add docs to all new classes
- [ ] Find a way to save classes to toml (probably should be handled the same way we handle dictionaries and None)


36000:
@arokem let me know what you think of this, it is a big change to the ui surrounding masks. It definitely makes things more flexible but it may be a little much.

AFQ/mask.py
```diff
@@ -0,0 +1,530 @@
...
+            raise TypeError(f"find_path method not found in {mask_name}")
+        else:
+            return False
+    elif not hasattr(mask, 'get_mask'):
+        if error:
+            raise TypeError(f"get_mask method not found in {mask_name}")
+        else:
+            return False
+    else:
+        return True
```
arokem, line 29:
```suggestion
    if not hasattr(mask, 'find_path'):
        if mask_name:
            raise TypeError(f"find_path method not found in {mask_name}")
        else:
            return False
    elif not hasattr(mask, 'get_mask'):
        if mask_name:
            raise TypeError(f"get_mask method not found in {mask_name}")
        else:
            return False
    else:
        return True
```

AFQ/mask.py
```diff
@@ -0,0 +1,530 @@
...
+        if self.combine == "or":
+            self.mask = np.zeros(shape, dtype=bool)
+        elif self.combine == "and":
+            self.mask = np.ones(shape, dtype=bool)
+        else:
+            self.combine_illdefined()
+
+    def combine_mask(self, other_mask):
+        if self.combine == "or":
+            self.mask = np.logical_or(self.mask, other_mask)
```
arokem, line 66: I wonder whether instead of implementing this combiner object you could implement the `__or__` and `__and__` operator for these objects. That way you can do something like: `combined = mask1 or mask2` or `combined mask1 and mask2` to express this idea, returning a new mask object as output. Does that do what you want?

AFQ/api.py
```diff
@@ -226,20 +235,16 @@ def __init__(self,
...
-            upper bounds respectively (see the format in the default). Each
-            scalar mask will 'logical and' to make the white-matter mask.
-            For an example of the former case,
-            the white matter values for the segmentation provided
-            with the HCP data including labels for midbrain are:
-            [250, 251, 252, 253, 254, 255, 41, 2, 16, 77].
-            Default: {'lb': {'dti_fa': 0.1}, 'ub': {'dti_md': 0.003}}
+        wm_mask : mask from AFQ.mask, optional
+            [REGISTRATION] This is the mask used to make
+            the white matter mask. Currently it is not used for anything.
```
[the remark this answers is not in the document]
36000: We don't use the variable wm_mask for anything (it could be removed, but you said it could useful for PFT?). 
The brain_mask, seed_mask, and stop_mask are all specified with their own mask.

36000:
> I like the idea of having an object-oriented interface to masks. I wonder if we don't want to go even a step further and have these objects define some of their own operations, such as `or` and `and` -- see comment about that. This does seem to simplify the code in api.py, but I still worry that we are digging ourselves deeper and deeper into an approach where we are parsing and generating the toml ourselves, rather than using the `toml` library from/to native Python objects (see #346 and related).

I hadn't realized the toml library could encode objects. Let me try that.

arokem:
> I hadn't realized the toml library could encode objects. Let me try that.

Oh no - that's not what I meant. I mean core Python objects. I.e., dictionaries.

arokem:
I don't think it can serialize and deserialize just any object. Would we need that?

36000:
We don't need that, but it could be cool. I see what you mean. Right now, what this PR does, is let you define the class in the toml. That Mixin class turns existing objects back to their definitions based on their __init__ function and assuming the arguments to the __init__ function are stored in the object. For example, our default toml's wm_mask looks like this:
```wm_mask = 'CombinedMask([ThresholdedScalarMask("dti_fa", 0.2, None, "and"), ThresholdedScalarMask("dti_md", None, 0.002, "and")], "and")'```
And our brain_mask looks like this:
```brain_mask = 'B0Mask()'```

**Comment by 36000** [comment]
We are valid toml, we just put complicated things like dicts and classes into strings.

**Comment by 36000** [comment]
Also, did you see this comment:
> We don't use the variable wm_mask for anything (it could be removed, but you said it could useful for PFT?).
The brain_mask, seed_mask, and stop_mask are all specified with their own mask.

**Comment by arokem** [comment]
Yeah - that worries me a bit, because the string parsing stuff is pretty
brittle. For example, extraneous whitespace can break it.

I think it would be better if dicts were handled in the proper toml way
(using nested config blocks) and maybe if objects were handled as dicts too.

**Comment by arokem** [comment]
> Also, did you see this comment:

>> We don't use the variable wm_mask for anything (it could be removed, but you said it could useful for PFT?).
>> The brain_mask, seed_mask, and stop_mask are all specified with their own mask.

Yes - still chewing on it. Need to go back and look exactly what we need for PFT. I think it requires several different anatomical segments: white matter, but also gray matter and CSF. So, complicated.

**Comment by 36000** [comment]
> Yeah - that worries me a bit, because the string parsing stuff is pretty brittle. For example, extraneous whitespace can break it. I think it would be better if dicts were handled in the proper toml way (using nested config blocks) and maybe if objects were handled as dicts too.

Not necessarily that brittle as we are using literal_eval and eval to turn the strings to dicts and classes respectively. I see your point though.

File: AFQ/mask.py (hunk 2 above)
Comment (36000) [reply AFQ/mask.py:70]: The main purpose of this class is to reduce the amount of times I have to write:
```
        if self.combine == "or":
            ...
        elif self.combine == "and":
            ...
        else:
            self.combine_illdefined()
```
Ie it determines whether to use 'and' or 'or'. If we implemented __or__, that might reduce the amount of times we have to write out np.logical_or, but we would still have to use the switch statement above. There may be a better way to name these variables.

arokem:
Following up on this conversation: 
> Also, did you see this comment:

>> We don't use the variable wm_mask for anything (it could be removed, but you said it could useful for PFT?).
>> The brain_mask, seed_mask, and stop_mask are all specified with their own mask.

>Yes - still chewing on it. Need to go back and look exactly what we need for PFT. I think it requires several different anatomical segments: white matter, but also gray matter and CSF. So, complicated.

Maybe we can call this input variable `tracking_mask` and use it as suggested in https://github.com/yeatmanlab/pyAFQ/pull/335#issuecomment-677929804?

36000:
We already have seed_mask, it might be confusing to have two masks, one of which overrides the other.

36000:
I think we should just remove wm_mask. Tractography already has seed_mask and stop_mask to work with.

PR #373, opened 2020-08-20
Return a unique set of subjects in S3BIDSStudy._all_subjects

richford:
This PR forces `AFQ.data.S3BIDSStudy` instances to return only unique subject IDs for it's `._all_subjects` attribute. Previously, the presence of qsiprep report html files would cause duplicate entries in `._all_subjects`. This did not affect the `.subjects` attribute but was still annoying. This fixes that issue and returns a unique subject list.

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -1029,7 +1029,8 @@ class S3BIDSStudy:
             list of participant_ids
         """
         if self._use_participants_tsv:
-            tsv_key = '/'.join([self.s3_prefix, 'participants.tsv']).lstrip("/")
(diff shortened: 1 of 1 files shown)


richford:
@36000, this should resolve the issues you were running into while running pyAFQ on the Cam-CAN dataset on S3. I've tested it locally and think it works now but please let me know if you run it using cloudknot and still run into the same issues.

36000:
It works for me. Ready to merge?

36000:
I am going to merge this and rebase the mask PR on it so I can use that PR on the cam can data.

PR #376, opened 2020-08-20
Add upload_to_s3, download_and_combine_afq_profiles

36000:
- [x] upload_to_s3
- [x] download_and_combine_afq_profiles

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -558,7 +558,7 @@ class S3BIDSSubject:
         self._s3_keys = s3_keys
 
     def download(self, directory, include_derivs=False, overwrite=False,
-                 pbar=True, pbar_idx=0):
+                 suffix=None, pbar=True, pbar_idx=0):
         """Download files from S3
 
         Parameters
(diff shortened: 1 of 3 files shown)


AFQ/data.py
```diff
@@ -605,24 +612,38 @@ def split_key(key):
             else:
                 return key
 
+        s3_keys_raw = self.s3_keys['raw']
```
36000, line 615: deep copy here

AFQ/data.py
```diff
@@ -605,24 +612,38 @@ def split_key(key):
...
+        s3_keys_deriv = self.s3_keys['derivatives']
```
36000, line 616: these lines are unnecessary; use the ['raw'] directly below

richford:
LGTM after [0f76930](https://github.com/yeatmanlab/pyAFQ/pull/376/commits/0f76930847f0d481856fa05ea9774f36e205fb1b) and discussion on our zoom chat today?

Issue #372, opened 2020-08-20
CI: add a GitHub action that runs a realistic dataset and uploads artifacts that can be examined locally


arokem:
This is something we do informally, and in the examples: run through a dataset at realistic scale and see that the artifacts produced (e.g., visualizations of bundles) look OK. We could have GitHub Actions do that for us, either on a regular basis as a cron job, or maybe even better, we could trigger this on a particular commit message, so that we can do this on candidate branches, before merging them into master.

PR #380, opened 2020-08-23
Update example to serve as a CI target

arokem:
The column names were updated in #279, and this follows that update.

The note that used to be here is no longer correct 🎉 

Also, we might want to always run this in the CI. Let's see how long this takes.

### DIFF
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install
       run: |
         python -m pip install --upgrade pip
-        pip install .[dev,fury]
+        pip install .[dev,fury,plotly]
     - name: Build docs
       run: |
         cd docs
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1319,6 +1319,7 @@ class AFQ(object):
                     figure.write_html(fname)
         row['timing']['Visualization'] =\
             row['timing']['Visualization'] + time() - start_time
+        return figure
 
     def _plot_tract_profiles(self, row):
         tract_profiles = pd.read_csv(self.get_tract_profiles()[0])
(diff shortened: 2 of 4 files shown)


arokem:
With this implemented, documentation builds take about 50 minutes. But, this has the advantage of satisfying some of the requirements of #372. I would like to also add generation of the GIFs and embedding these into the examples page. That would be a think to quickly check when we are making changes to default processing or API.

36000:
A few ideas:
- [ ] Not running the other tests would save 10 minutes.
- [ ] We could only generate gifs for a few key tracts.
- [ ] We could also reduce the number of tracts we end up plotting by setting n_seeds to a very small number ~(1-10k) and random_seeds to true in tractography.

arokem:
[ ] Experiment with embedding plotly htmls into the example webpage.

arokem:
Not sure I understand this suggestion: 

> Not running the other tests would save 10 minutes.

The tests are not running in the documentation build CI workflow. Or do you mean running the other examples?

36000:
I meant not running the other examples. But that is more of a last resort.

arokem:
Need to understand how the viewers are embedded in this example: https://brainsprite.github.io/auto_examples/plot_anat.html#sphx-glr-auto-examples-plot-anat-py

arokem:
Or maybe better: https://github.com/plotly/plotly-sphinx-gallery/blob/master/examples/plot_plotly.py

arokem:
This is very neat: the resulting example now includes an interactive visualization of all the segmented bundles. 

To avoid running this every time, we could use something like: https://github.com/Khan/pull-request-comment-trigger.

arokem:
@deploy

arokem:
build

arokem:
build

arokem:
I gave up on triggering the builds with comments. I think that this is ready to merge, but happy for more feedback.

36000:
I fixed the bug this is catching in #395 which was introduced in #393 . May want to wait and rebase this on that, to make sure you are passing.

36000:
Otherwise, LGTM!

arokem:
OK. I think that this is done for now. I will merge when it goes green (in about 40 minutes :-D)

PR #383, opened 2020-08-24
Change nighty tests to happen at midnight, PST

36000:


arokem:
Let's see what happens tonight🤞

Issue #263, opened 2020-05-15
error for run pyAFQ on qsiprep-preprocessed data


weiwei-wch:
Hi all！
    recently I run pyAFQ on some qsiprep-preprocessed data and followed error occured. before this error I ran pyAFQ on data only preprocessed by eddy successfully. is there any way to solve this problem?
Here is the error code:
INFO:AFQ.tractography:Tracking...
100%
INFO:AFQ.api:Saving /home/weiwei/Desktop/AFQtest/afq/sub-N5110WWEIWEI/sess01/sub-N5110WWEIWEI_sess-01_dwi_space-RASMM_model-DTI_desc-det_tractography.trk
Traceback (most recent call last):
  File "/usr/local/bin/pyAFQ", line 198, in <module>
    myafq.export_ROI_gifs()
  File "/usr/local/lib/python3.6/dist-packages/AFQ/api.py", line 1360, in export_ROI_gifs
    self.data_frame.apply(self._export_ROI_gifs, axis=1)
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/frame.py", line 6014, in apply
    return op.get_result()
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/apply.py", line 142, in get_result
    return self.apply_standard()
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/apply.py", line 248, in apply_standard
    self.apply_series_generator()
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/apply.py", line 277, in apply_series_generator
    results[i] = self.f(v)
  File "/usr/local/lib/python3.6/dist-packages/AFQ/api.py", line 1055, in _export_ROI_gifs
    bundles_file = self._clean_bundles(row)
  File "/usr/local/lib/python3.6/dist-packages/AFQ/api.py", line 817, in _clean_bundles
    bundles_file = self._segment(row)
  File "/usr/local/lib/python3.6/dist-packages/AFQ/api.py", line 774, in _segment
    tg = load_tractogram(streamlines_file, img, Space.VOX)
  File "/usr/local/lib/python3.6/dist-packages/dipy/io/streamline.py", line 135, in load_tractogram
    if not is_header_compatible(filename, reference):
  File "/usr/local/lib/python3.6/dist-packages/dipy/io/utils.py", line 321, in is_header_compatible
    reference_2)
[...]

PR #384, opened 2020-08-25
Update to Dipy 1.2.0

36000:
The load_tractogram in dipy v1.1.1 uses srow to get the affine information from a reference. However, some files have affine stored in the q info (q offset, pixdim), not s. This is fixed in the current dipy master, where they find the best source for affine information, but not dipy 1.1.1 .

Until a new version of dipy is released, we can use this PR with some datasets. This PR will probably not need to be merged, especially if a new version of dipy is releasing soon.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -387,17 +387,9 @@ class Segmentation:
         dist = []
         for roi in include_rois:
             # Use squared Euclidean distance, because it's faster:
-            try:
-                dist.append(cdist(sl, roi, 'sqeuclidean'))
-                if np.min(dist[-1]) > tol:
-                    # Too far from one of them:
-                    return False, []
-            except ValueError:
-                self.logger.warning(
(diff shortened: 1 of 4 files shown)


36000:
Fixes #263

arokem:
Heads up that DIPY 1.2 is due to be released any day now. We can probably require that as a dependency and then merge this PR?

AFQ/segmentation.py
```diff
@@ -387,9 +387,17 @@ def _check_sl_with_inclusion(self, sl, include_rois, tol):
         dist = []
         for roi in include_rois:
             # Use squared Euclidean distance, because it's faster:
-            dist.append(cdist(sl, roi, 'sqeuclidean'))
-            if np.min(dist[-1]) > tol:
-                # Too far from one of them:
+            try:
```
arokem, line 390: Maybe move this into a different PR? I think we should catch the case where the ROI is empty before we are here.

PR #385, opened 2020-08-26
[WIP] Bids layout get expects no session filter if there is no session, not None

36000:
This might not be necessary. Needs to be tested.


36000:
I don't think this is necessary.

PR #386, opened 2020-08-27
Add ability to remove edges of profiles

36000:


36000:
@arokem  this is ready for review / merge

arokem:
Sorry about this: after some thinking, it occurs to me that this might better be handled in post-processing of the CSV files by the user. The information is all there at that point. WDYT?

arokem:
Oh wait - this is all at the level of the visualization. That is, in a sense, postprocessing. Merging.

PR #387, opened 2020-08-28
make decision to combine profile automatic in export_all

36000:


36000:
@arokem  this is ready for review.

Issue #392, opened 2020-08-28
pyAFQ timing info (possibly in export_all?) 


36000:
pyAFQ should output the amount of time it spends on each operation to a file so we don't have to rely on my vague intuitions from occasionally checking cloudwatch logs.

PR #393, opened 2020-08-31
Export timing information

36000:
@arokem I would be interested your advice on where to put the start and end times.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -324,15 +324,6 @@ class AFQ(object):
                                             seg_algo=self.seg_algo,
                                             resample_to=self.reg_template_img)
 
-        # Store relevant timing information
-        self.timing = {
-            "Tracking": 0,
-            "Registration": 0,
-            "Segmentation": 0,
-            "Cleaning": 0,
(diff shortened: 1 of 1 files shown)


36000:
Closes #392

arokem:
I think what you are doing here makes sense. This creates a separate CSV file for each subject, right? And time in seconds in each row?

36000:
Yes.

arokem:
Seems good to me!

arokem:
Looks like this broke our default: https://github.com/yeatmanlab/pyAFQ/pull/380/checks?check_run_id=1053390003#step:5:557

AFQ/api.py
```diff
@@ -839,6 +857,8 @@ def _streamlines(self, row):
                 include_track=True)
             afd.write_json(meta_fname, meta)
             self.log_and_save_trk(sft, streamlines_file)
+            row['timing']['Tractography'] =\
```
arokem, line 860: The key here should be "Tracking" not "Tractography"

36000: #395

PR #399, opened 2020-09-02
[WIP] Make pyAFQ into a brainlife App

36000:


arokem:
We don't need to keep doing this because it looks like the brainlife team is already on it: https://github.com/brainlife/app-pyafq-segment

soichih:
@arokem We recently noticed that AFQ.segment function will *modify* the input streamline's space. For example, we feed tg.space set to rasmm and pass it to segment(). After the segmentation runs, the tg.space is now set to VOX. We basically had to reload the whole streamline from the disk to revert it back to the rasmm space. Is this an expected behavior?

@giulia-berto made this PR to correct this
> https://github.com/brainlife/app-pyafq-segment/pull/1

PR #409, opened 2020-09-08
Many minor bug fixes

### LINKED ISSUE #391: Y-axis range for MD tract profile plots is too large
sub-01_ses-01_dwi_space-RASMM_model-DTI_desc-det-AFQ_dti_md_profile_plots

### LINKED ISSUE #403: Case invariance
In tractography the odf model is not case invariant. It should be, and we should double check that everywhere we input strings that are not paths that we are case invariant.

### LINKED ISSUE #404: This row sometimes flips the `b0_in_MNI` file
https://github.com/yeatmanlab/pyAFQ/blob/b1f86b6ad0223bce155f1b4c4f15554845e27056/AFQ/api.py#L746

I believe the affine for this file should be the MNI affine, not the DWI affine.

36000:
Closes #404 , Closes #403, Closes #391

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -808,7 +808,7 @@ class AFQ(object):
         return mapping_file
 
     def _streamlines(self, row):
-        odf_model = self.tracking_params["odf_model"]
+        odf_model = self.tracking_params["odf_model"].upper()
 
         streamlines_file = self._get_fname(
             row,
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -145,6 +145,7 @@ class AFQ(object):
 
     def __init__(self,
                  bids_path,
+                 bids_filters={'suffix': 'dwi'},
                  dmriprep='all',
                  b0_threshold=50,
                  min_bval=None,
@@ -179,6 +180,9 @@ class AFQ(object):
             [BIDS] The path to preprocessed diffusion data organized in a BIDS
             dataset. This should contain a BIDS derivative dataset with
             preprocessed dwi/bvals/bvecs.
+        bids_filters : dict
+            [BIDS] Filter to pass to bids_layou.get when finding DWI files.
+            Default: {'suffix': 'dwi'}
         dmriprep : str, optional.
             [BIDS] The name of the pipeline used to preprocess the DWI data.
             Default: "dmriprep".
@@ -380,21 +384,22 @@ class AFQ(object):
                 os.makedirs(results_dir_list[-1], exist_ok=True)
                 dwi_file_list.append(
                     bids_layout.get(subject=subject, session=session,
-                                    extension='nii.gz', suffix='dwi',
+                                    extension='nii.gz',
                                     return_type='filename',
-                                    scope=dmriprep)[0])
+                                    scope=dmriprep,
+                                    **bids_filters)[0])
                 bvec_file_list.append(
                     bids_layout.get(subject=subject, session=session,
                                     extension=['bvec', 'bvecs'],
-                                    suffix='dwi',
                                     return_type='filename',
-                                    scope=dmriprep)[0])
+                                    scope=dmriprep,
+                                    **bids_filters)[0])
                 bval_file_list.append(
                     bids_layout.get(subject=subject, session=session,
                                     extension=['bval', 'bvals'],
-                                    suffix='dwi',
                                     return_type='filename',
-                                    scope=dmriprep)[0])
+                                    scope=dmriprep,
+                                    **bids_filters)[0])
 
                 if check_mask_methods(self.tracking_params["seed_mask"]):
                     self.tracking_params["seed_mask"].find_path(
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -309,6 +309,8 @@ class AFQ(object):
                 default_tracking_params[k] = tracking_params[k]
 
         self.tracking_params = default_tracking_params
+        self.tracking_params["odf_model"] =\
+            self.tracking_params["odf_model"].upper()
 
         default_seg_params = get_default_args(seg.Segmentation.__init__)
         if segmentation_params is not None:
@@ -808,7 +810,7 @@ class AFQ(object):
         return mapping_file
 
     def _streamlines(self, row):
-        odf_model = self.tracking_params["odf_model"].upper()
+        odf_model = self.tracking_params["odf_model"]
 
         streamlines_file = self._get_fname(
             row,
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -145,8 +145,8 @@ class AFQ(object):
 
     def __init__(self,
                  bids_path,
-                 bids_filters={'suffix': 'dwi'},
-                 dmriprep='all',
+                 bids_filters={"suffix": "dwi"},
+                 dmriprep="all",
                  b0_threshold=50,
                  min_bval=None,
                  max_bval=None,
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1170,7 +1170,7 @@ class AFQ(object):
 
             mapping = reg.read_mapping(self._mapping(row),
                                        row['dwi_file'],
-                                       self.reg_template,
+                                       self.reg_template_img,
                                        prealign=reg_prealign_inv)
 
         if xform_volume:
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -181,7 +181,7 @@ class AFQ(object):
             dataset. This should contain a BIDS derivative dataset with
             preprocessed dwi/bvals/bvecs.
         bids_filters : dict
-            [BIDS] Filter to pass to bids_layou.get when finding DWI files.
+            [BIDS] Filter to pass to bids_layout.get when finding DWI files.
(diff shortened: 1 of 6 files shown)


36000:
Closes #403

arokem:
This is still WIP, right? I changed the title to reflect that.

36000:
@arokem ready for review. If this passes, we can merge.

AFQ/api.py
```diff
@@ -179,6 +180,9 @@ def __init__(self,
             [BIDS] The path to preprocessed diffusion data organized in a BIDS
             dataset. This should contain a BIDS derivative dataset with
             preprocessed dwi/bvals/bvecs.
+        bids_filters : dict
+            [BIDS] Filter to pass to bids_layou.get when finding DWI files.
```
arokem, line 184:
```suggestion
            [BIDS] Filter to pass to bids_layout.get when finding DWI files.
```

AFQ/api.py
```diff
@@ -179,6 +180,9 @@ def __init__(self,
...
+            Default: {'suffix': 'dwi'}
```
arokem, line 185:
```suggestion
            Default: {"suffix": "dwi"}
```

AFQ/api.py
arokem: I think this is still not addressed? This is a typo currently, right?

AFQ/api.py
```diff
@@ -282,17 +289,17 @@ def __init__(self,
...
             from xvfbwrapper import Xvfb
             self.vdisplay = Xvfb(width=1280, height=1280)
             self.vdisplay.start()
-        self.viz = Viz(backend=viz_backend)
+        self.viz = Viz(backend=viz_backend.lower())
 
         default_tracking_params = get_default_args(aft.track)
-        default_tracking_params["seed_mask"] = ScalarMask("dti_fa")
-        default_tracking_params["stop_mask"] = ScalarMask("dti_fa")
+        default_tracking_params["seed_mask"] = ScalarMask(self.scalars[0])
```
arokem, line 301: Ooh. This can be rather brittle. What if a user asks for `["dti_md", "dti_fa"]` as scalars?

AFQ/api.py
```diff
@@ -800,7 +808,7 @@ def _mapping(self, row):
         return mapping_file
 
     def _streamlines(self, row):
-        odf_model = self.tracking_params["odf_model"]
+        odf_model = self.tracking_params["odf_model"].upper()
```
arokem, line 811: Maybe we should make `tractography.track` robust to different input cases?

AFQ/mask.py
```diff
@@ -238,12 +238,23 @@ def find_path(self, bids_layout, subject, session):
         pass
 
     def get_mask(self, api, row):
```
arokem, line 240: This one doesn't have a docstring. What kind of thing is the `api` argument here?

arokem: Is this an api.AFQ class instance? If so, we might want to call this input something else. Maybe `afq_object`?

AFQ/tests/test_api.py
```diff
@@ -349,7 +349,7 @@ def test_AFQ_data_waypoint():
     tgram = load_tractogram(myafq.bundles[0], myafq.dwi_img[0])
 
     bundles = aus.tgram_to_bundles(tgram, myafq.bundle_dict, myafq.dwi_img[0])
-    npt.assert_(len(bundles['CST_R']) > 0)
+    npt.assert_(len(bundles['CST_L']) > 0)
```
arokem, line 352: Do you understand why this changed? That is, which of the other changes in this PR required that this be changed?

arokem: I will say that I took a look at the documentation example, and the segmentation, as well as tract profile does look god, but I am always a bit worried when the segmentation changes, even if it's for this rather skeletal test case.

AFQ/tests/test_api.py
```diff
@@ -363,24 +363,24 @@ def test_AFQ_data_waypoint():
...
+        'sub-01_ses-01_dwi_space-RASMM_model-dti_desc-det-AFQ-clean_tractography_idx.json'))  # noqa
 
     tract_profiles = pd.read_csv(myafq.tract_profiles[0])
     assert tract_profiles.shape == (500, 5)
 
     myafq.plot_tract_profiles()
     assert op.exists(op.join(
         myafq.data_frame['results_dir'][0],
-        'sub-01_ses-01_dwi_space-RASMM_model-DTI_desc-det-AFQ_dti_fa_profile_plots.png'))  # noqa
+        'sub-01_ses-01_dwi_space-RASMM_model-dti_desc-det-AFQ_dti_fa_profile_plots.png'))  # noqa
```
arokem, line 379: These file-names no longer comply with the draft of the BIDS diffusion derivatives: https://bids-specification.readthedocs.io/en/derivatives/05-derivatives/06-diffusion-derivatives.html, which has the model label for DTI with upper-case.

AFQ/viz/utils.py
```diff
@@ -429,10 +443,15 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
             (tract_profiles["bundle"] == bundle)
         ][scalar].values
         ax.plot(fa, 'o-', color=COLOR_DICT[bundle])
-        ax.set_ylim([min_fa, max_fa])
-        ax.set_yticks([0.2, 0.4, 0.6])
-        ax.set_yticklabels([0.2, 0.4, 0.6])
-        ax.set_xticklabels([])
+
+        if (min_scalar is not None) and (max_scalar is not None):
+            ax.set_ylim([min_scalar, max_scalar])
```
arokem, line 448: Maybe instead of `min_scalar` and `max_scalar`, get an input `ylim`? You can take `[None, 0.5]` for cases where the user only wants to set the upper limit.

AFQ/viz/utils.py
```diff
@@ -429,10 +443,15 @@ def visualize_tract_profiles(tract_profiles, scalar="dti_fa", min_fa=0.0,
...
+
+        if use_fa_ticks:
```
arokem, line 450: Maybe clarify in the docstring that `use_fa_ticks` takes precedent over `min_scalar`/`max_scalar`?

AFQ/api.py
36000: It is. This is for these checks a few lines below:
```
            if odf_model == "DTI":
                params_file = self._dti(row)
            elif odf_model == "CSD":
                params_file = self._csd(row)
            elif odf_model == "DKI":
                params_file = self._dki(row)

```

AFQ/api.py
36000: Then "dti_md" would be the seed mask. I think its more brittle, but I don't know what the smart default is for this. Maybe look in scalars for a string containing fa, and if its not there, then use the first one? I want this to be more automatic, such that when the user inputs ["dki_fa", "dki_md"] she doesnt have to change a bunch more parameters.

AFQ/tests/test_api.py
36000: This only uses 100 streamlines to be fast. I think the update to dipy might have changed it? I could up the sls to 200 to see if that finds the CST_R.

AFQ/tests/test_api.py
36000: I suppose I could .upper() this when its inputted, and avoid some of these problems.

AFQ/tests/test_api.py
arokem: OK - if it's the DIPY upgrade, then that's probably fine.

AFQ/tests/test_api.py
arokem: Yes. That's probably a good way to go.

AFQ/tests/test_api.py
36000: Actually, I think its because I wasn't using the subsampled tractography anymore because it was saved with the uppercase model-DTI.
Due to changes recommended in another comment, everything uses uppercase again. But I am going to remove the subsampled tractography because I want to use the ROIMask for tractography instead.

36000:
@arokem this is ready for another review I believe

AFQ/tests/test_api.py
arokem: Oh good! This one test is one that has proven to be a really useful canary in the coal-mine in the past, so whenever it fails, it raises questions. But I think that changing as you suggest should be fine.

AFQ/tests/test_api.py
```diff
@@ -349,7 +343,8 @@ def test_AFQ_data_waypoint():
     tgram = load_tractogram(myafq.bundles[0], myafq.dwi_img[0])
 
     bundles = aus.tgram_to_bundles(tgram, myafq.bundle_dict, myafq.dwi_img[0])
-    npt.assert_(len(bundles['CST_R']) > 0)
+    print(bundles)
```
arokem, line 346: I think we'll want to remove this print statement before merging. 

Other than that - I think this is ready to go?

36000: Oh - nice catch! must have left that in there from debugging.

Issue #397, opened 2020-09-01
Possible empty bundle in Cam-CAN subject


richford:
Using PR #384 (i.e. cloudknot with `image_github_installs=https://github.com/36000/pyAFQ.git@avoid_dipy_load_tractogram_bug#egg=pyAFQ[fury]`), I get the [following error](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/afq_cam_can-arh-0-cloudknot-job-definition$252Fdefault$252F6163d18d-3d82-4b48-9af2-4b5d4e8444aa):

```
INFO:AFQ.Segmentation:Finding Streamlines for SLF_L
/usr/local/lib/python3.7/site-packages/AFQ/utils/volume.py:31: UserWarning: Input image is entirely zero, no valid convex hull. Returning empty image
  return convex_hull_image(hole_filled)
INFO:AFQ.Segmentation:97050 streamlines exceed the probability threshold.
Traceback (most recent call last):
  File "/home/cloudknot-user/afq-cam-can.py", line 169, in <module>
    pickle_to_s3(args.sse, args.arrayjob)(afq_cam_can)(input_)
  File "/home/cloudknot-user/afq-cam-can.py", line 35, in wrapper
    result = f(*args, **kwargs)
  File "/home/cloudknot-user/afq-cam-can.py", line 108, in afq_cam_can
    myafq.export_all()
  File "/usr/local/lib/python3.7/site-packages/AFQ/api.py", line 1603, in export_all
    self.export_bundles()
  File "/usr/local/lib/python3.7/site-packages/AFQ/api.py", line 1547, in export_bundles
    self.data_frame.apply(self._export_bundles, axis=1)
  File "/usr/local/lib/python3.7/site-packages/pandas/core/frame.py", line 6878, in apply
    return op.get_result()
  File "/usr/local/lib/python3.7/site-packages/pandas/core/apply.py", line 186, in get_result
    return self.apply_standard()
  File "/usr/local/lib/python3.7/site-packages/pandas/core/apply.py", line 296, in apply_standard
    values, self.f, axis=self.axis, dummy=dummy, labels=labels
  File "pandas/_libs/reduction.pyx", line 620, in pandas._libs.reduction.compute_reduction
  File "pandas/_libs/reduction.pyx", line 128, in pandas._libs.reduction.Reducer.get_result
[...]

PR #415, opened 2020-09-09
Raise error earlier for empty ROI

36000:
In response to #397 . Does not fix issue but makes a more clear error.

- [x] Add test

### DIFF
--- a/AFQ/utils/volume.py
+++ b/AFQ/utils/volume.py
@@ -1,10 +1,13 @@
+import logging
+import numpy as np
+
 import scipy.ndimage as ndim
 from skimage.filters import gaussian
 from skimage.morphology import convex_hull_image
 from scipy.spatial.qhull import QhullError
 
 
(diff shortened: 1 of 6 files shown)


36000:
@arokem does this look like a smart way to handle this?

AFQ/utils/volume.py
```diff
@@ -21,12 +24,20 @@ def patch_up_roi(roi):
...
+        Name of bundle, which may be useful for error messages.
+        Default: None
+
     Returns
     -------
     ROI after dilation and hole-filling
     """
 
     hole_filled = ndim.binary_fill_holes(roi > 0)
+    if np.sum(hole_filled) < 1:
```
arokem, line 37:
```suggestion
    if not np.any(hole_filled):
```

arokem: Maybe we should a small test for this function?

PR #417, opened 2020-09-09
Divide nightly test into 2; have nightly tests only do nightly tests

36000:
This should help with OOM error or at least binary search for the test causing it.

### DIFF
--- a/.github/workflows/nightly2_test.yml
+++ b/.github/workflows/nightly2_test.yml
@@ -2,7 +2,7 @@ name: Nightly test suite
 
 on:
   schedule:
-    - cron: '0 7 * * *'  # every day at midnight, PST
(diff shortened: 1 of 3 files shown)


arokem:
Could you do the same resampling for the visualization you did for the updated api example? In case the oom is because of the visualization tests

arokem:
Could you do the same resampling for the visualization you did for the updated api example? In case the oom is because of the visualization tests

36000:
I checked, and I dont think any of these tests make visualizations. Only test_AFQ_data_waypoint does, which is a part of the normal tests.

arokem:
Oh - I just realized that we'd need to merge this to master to get this to work. Merging.

PR #420, opened 2020-09-11
Automatically choose bundle names for recobundles

36000:
- [x] Mark reco bundles test as nightly once it is passing


36000:
@arokem I think this is ready to go

arokem:
Yep. Looks good!

PR #422, opened 2020-09-14
Allow user to input custom tractography

36000:
- [x] Consider moving test to nightly

### DIFF
--- a/AFQ/tests/test_api.py
+++ b/AFQ/tests/test_api.py
@@ -173,7 +173,8 @@ def test_AFQ_custom_tract():
         dmriprep='vistasoft',
         bundle_names=bundle_names,
         custom_tractography_bids_filters={
-            "suffix": "tractography"
+            "suffix": "tractography",
+            "scope": "vistasoft"
         })
     my_afq.export_rois()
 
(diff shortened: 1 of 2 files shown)


36000:
@arokem I think this is ready to go

AFQ/tests/test_api.py
```diff
@@ -138,6 +138,46 @@ def create_dummy_bids_path(n_subjects, n_sessions):
...
+        op.join(
+            op.expanduser('~'),
+            'AFQ_data',
+            'stanford_hardi_tractography',
+            'tractography_subsampled.trk'),
+        op.join(
+            tmpdir.name,
+            'stanford_hardi',
+            'derivatives',
+            'vistasoft',
```
arokem, line 165: Does the tractography have to reside in the 'dmriprep' derivatives folder?  Can it be in another folder (i.e., generated by some other pipeline)?

36000: Good point. Scope should be specified by the user filters instead.

36000: I made this change in "make scope a user choice" commit. Does this look like a good solution?

36000:
@arokem I think I addressed the earlier review. This is ready for another review or merge.

arokem:
Yep. I think that we might want to provide better documentation and maybe even an example in the gallery for how this is done. I'll add an issue to address this on a separate PR.

PR #423, opened 2020-09-14
BF: Register the PVE to the parameters.

arokem:
It's a bit hard to test and demonstrate this as a bug, but this produces nonsense streamlines in practice. It also makes sense when you think about it.


36000:
LGTM. Is PVE tested anywhere? I could test it when I make the masks PR.

arokem:
There are some pretty simple-minded tests [here](https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/tests/test_tractography.py#L78). As you can see, they're not really designed to catch subtle things :-)

Issue #421, opened 2020-09-11
Extend recobundles to use the more comprehensive atlas


arokem:
This one has 80 bundles: https://figshare.com/articles/Advanced_Atlas_of_80_Bundles_in_MNI_space/7375883

36000:
Done in #431

PR #431, opened 2020-09-15
Adds 80-bundle atlas for RB

arokem:
Addresses #421, but we need to think about how this integrates with work already done in #420.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -45,6 +45,7 @@ class Segmentation:
     def __init__(self,
                  nb_points=False,
                  seg_algo='AFQ',
+                 slr_algo="linear",
                  clip_edges=False,
                  progressive=True,
                  greater_than=50,
@@ -75,6 +76,10 @@ class Segmentation:
             'Reco': Segment streamlines using the RecoBundles algorithm
             [Garyfallidis2017].
             Default: 'AFQ'
+        slr_algo : string
+            Algorithm for streamline registration (case-insensitive):
+            'linear' : Use Streamlinear Registration [Garyfallidis2015]_
+            'syn' : Use image-based nonlinear registration
         clip_edges : bool
             Whether to clip the streamlines to be only in between the ROIs.
             Default: False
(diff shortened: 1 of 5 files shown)


AFQ/segmentation.py
```diff
@@ -642,6 +648,33 @@ def segment_afq(self, tg=None):
...
+        if slr_algo == "linear":
+            self.logger.info("Registering tractogram with linear SLR")
+            atlas = self.bundle_dict['whole_brain']
+            self.moved_sl, _, _, _ = whole_brain_slr(
+                atlas, self.tg.streamlines, x0='affine', verbose=False,
+                progressive=self.progressive,
+                greater_than=self.greater_than,
+                rm_small_clusters=self.rm_small_clusters,
+                rng=self.rng)
+        elif slr_algo == "syn":
```
36000, line 672: following the convention in the api.AFQ object, the variable name should be reg_algo and the values should be "slr" or "syn". Also, there should be a way for the api to pass in its own mapping which it already created.

36000: We might want to just remove SLR from the segmentation class and ask for a mapping. Just like syn, if the user wants to use slr, they should use: https://github.com/yeatmanlab/pyAFQ/blob/9dda145ee06314a45862c07a0137c523276dd5f2/AFQ/registration.py#L476
This reduces redundant code and makes it so all mapping have the same interface.

arokem:
Thanks for the review @36000! I am done with a first pass here, but still need to address your comments.

arokem:
Looks like the change in defaults from 'slr' to 'syn' is causing a regression in one of the tests: https://github.com/yeatmanlab/pyAFQ/pull/431/checks?check_run_id=1146298668#step:6:56, so I am going to investigate a bit more.

arokem:
For now, I suggest that we back off making 'syn' the default registration, until we've had an opportunity to experiment with this more. WDYT @36000 ? Other than that, I think it's ready to merge, once the tests pass.

36000:
LGTM

Issue #394, opened 2020-08-31
Libgl xvfb docker image folder


36000:
To run fury on the cloud using cloudknot, you need to use a docker image that installs libgl and xvfb. We should document this and add a dockerfile somewhere in pyAFQ that can be used to make the image.

arokem:
Maybe add that to #439?

PR #439, opened 2020-09-16
Try plotly as default, re-organize usage documentation

36000:
- [x] make b0 default background
- [x] make plotly default viz lib
- [x] make plotly installs non optional
- [x] fix kaleido version
- [x]  make gifs with plotly
- [x] resample to 40 by default
- [x] run plotly on cloudknot successfully
- [x] plotly_no_gif as default
- [x] documentation page

Note: It takes ~2 minutes to create a gif out of the plotly figures, and for ~20 gifs, this adds to ~40 minutes.


36000:
@arokem what do you think of making plotly the default?

arokem:
I can see pros and cons. For example, those gifs are really easy to drop into a GitHub issue. It’s so easy to switch between them, so I’m not even sure it matters all that much.

36000:
It might help with memory issues on the nightly tests. I don't know which backend uses more memory.

36000:
Really want I want is both GIFs and HTML. I should revisit the kaleido problems with exporting GIFs.

36000:
It looks like plotly is taking around 1-2 seconds to export each frame in the gif. Unfortunately, our tests produce ~40 gifs (one batch of ~20 for testing afq.api module, another batch of ~20 for testing the cli) with 60 frames each. This leads to the test taking ~1.5 hours to finish. We could cut the amount of gifs the tests produce, but is 1-2 minutes per gif an acceptable pace for the end user as our default? I will see what I can do to reduce this time.

36000:
@arokem this is ready to go but might warrant an in person discussion about the pros/cons of the various features. Basically plotly can make html and gifs and is ready to go after `pip install pyafq`, but making the gifs is super slow and will add about an hour in processing time.

arokem:
This might merit more in-person conversation. I think that adding about 40 minutes to processing for these gifs is definitely not something we want to apply as the default setting. Generally, I think that these gifs may actually be less important in the use-case where the one big HTML is produced. That's because in the HTML file you can toggle individual bundles on and off. So maybe in the plotly case, we produce only the one big HTML and not the rotating gifs? Alternatively, we could add some complexity to the whole thing by exposing different options for what gets visualized in the API.

arokem:
OK - I think that what we discussed is to make `'plotly_no_gif'` the default behavior and adding a documentation page that explains all the different viz options. Future PRs will automate the creation of websites out of the plotly viz.

arokem:
I think that now that #482 is merged you can rebase and move ahead with this one.

36000:
break up usage into directory structure

36000:
Closes #394

AFQ/api.py
```diff
@@ -271,9 +271,11 @@ def __init__(self,
             [VIZ] Whether to use a virtual fram buffer. This is neccessary if
             generating GIFs in a headless environment. Default: False
         viz_backend : str, optional
-            [VIZ] Which visualization backend to us.
-            One of {"fury", "plotly"}.
-            Default: "fury"
+            [VIZ] Which visualization backend to use.
+            See Visualization Backends page in documentation
```
arokem, line 275: I believe that you can link directly to the docs from here using rst link syntax. If the link doesn't exist yet, it's pretty straightforward to infer what it will be once this is released.

arokem (commented):
Looks good! I had one small comment about the docstring, suggesting to add a link to the online docs. But other than that, I think this is ready to go?

36000:
@arokem I am tweaking the docs pages a little bit as I learn the rst format. I made the fix you suggested. I will merge it when done!

PR #445, opened 2020-09-18
Rename profile columns back to standard column names

### LINKED ISSUE #369: Make outputs that can be directly opened in AFQ-Browser
If there is participant data: a participants.tsv or participants.json file in the BIDS dataset, we can use that to generate subjects.csv. Otherwise, this will just be subject (and session?) ids. 

And we need to take our CSV file and make an AFQ-browser compliant nodes.csv.

36000:
Resolves #369


36000:
@arokem when these tests pass this should be ready to merge.

PR #456, opened 2020-09-21
Type check AFQ object inputs

36000:


36000:
@arokem does this look good?

arokem:
Maybe add some tests that trigger these errors?

36000:
A way to do this automatically from the docstring?

36000:
> A way to do this automatically from the docstring?

Let's worry about this later.

Issue #448, opened 2020-09-18
Only load part of the Callosum Template


36000:
Callosum Template takes ~45 seconds to load, and is loaded whenever we make the default bundle dict, even though we only use the mid sagital.

PR #463, opened 2020-09-22
Catch no subject error early

36000:


arokem:
Any chance to add a test of this?

AFQ/tests/test_api.py
```diff
@@ -180,6 +180,23 @@ def test_AFQ_custom_tract():
 
 
 @pytest.mark.nightly2
+def test_AFQ_no_derivs():
```
arokem, line 183: Should this test really be marked as nightly? It probably runs pretty quickly (?)

36000: test_AFQ_init is also marked nightly. Initializing an afq object takes about a minute I think, and if we have a lot of these tests, they can add up. To reduce time to make afq object, see: https://github.com/yeatmanlab/pyAFQ/issues/448

arokem: OK - that makes sense. Merging!

PR #468, opened 2020-09-22
Streamline counts table

arokem:
As we are exporting the bundles, count the streamlines in each bundle, and store that in a CSV.


36000:
LGTM

PR #469, opened 2020-09-23
Adds total number of streamlines, to complete the sl counts table.

### LINKED ISSUE #464: New output: number of streamlines in each bundle
Generate a csv file that has two columns: bundle name and number of streamlines.

arokem:
Closes #464.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1237,7 +1237,7 @@ class AFQ(object):
 
             sl_count_file = self._get_fname(
                 row,
-                '_sl_count.csv',
+                f'{func}_sl_count.csv',
                 include_track=True,
                 include_seg=True)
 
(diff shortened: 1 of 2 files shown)


arokem:
I've overhauled this a bit, so it's a separate function. Instead of two separate files, it now saves a single table with two columns: one for the bundles and the other for the cleaned bundles, also giving a sense of how many sl are removed in cleaning.

36000:
consider Integrating changes in #471 and make sure it passes test_AFQ_reco.

arokem:
Ah yes: that makes sense. I was testing this only with `seg_algo='afq'`

arokem:
OK - I will finish this before proceeding to rebase/fix #470.

arokem:
Looks like it's working. Merging.

PR #470, opened 2020-09-23
Follow up on #462, to fix failing nightly test.

arokem:


arokem:
This will be ready to merge, once I confirm that the nightly test that was failing passes the CI.

36000:
Rebase this once I merge #471 and then it should pass.

arokem:
I believe this is now gtg. Merging so we can get this running with the nightly tests.

PR #471, opened 2020-09-23
FIX: remove whole brain from bundle list for sl count when using recobundles

36000:
- [x] wait for reco api test to pass


arokem:
Looks like the tests are passing. Moving right along.

PR #472, opened 2020-09-23
Reg algo automatically chosen based on whether mapping is provided, syn mapping for sls fixed, recobundle defaults updated

36000:
- [x] passes reco api test

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -78,7 +78,8 @@ class Segmentation:
             Algorithm for streamline registration (case-insensitive):
             'slr' : Use Streamlinear Registration [Garyfallidis2015]_
             'syn' : Use image-based nonlinear registration
-            If None, will use syn if a mapping is provided, slr otherwise.
+            If None, will use SyN if a mapping is provided, slr otherwise.
+            If  seg_algo="AFQ", SyN is always used.
             Default: None
         clip_edges : bool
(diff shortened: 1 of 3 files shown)


36000:
@arokem this is ready for review.

AFQ/segmentation.py
```diff
@@ -73,11 +74,12 @@ def __init__(self,
             'Reco': Segment streamlines using the RecoBundles algorithm
             [Garyfallidis2017].
             Default: 'AFQ'
-        reg_algo : string, optional
+        reg_algo : string or None, optional
             Algorithm for streamline registration (case-insensitive):
             'slr' : Use Streamlinear Registration [Garyfallidis2015]_
             'syn' : Use image-based nonlinear registration
-            Default: 'syn'
+            If None, will use syn if a mapping is provided, slr otherwise.
```
arokem, line 81:
```suggestion
            If None, will use SyN if a mapping is provided, slr otherwise.
            If  seg_algo="AFQ", SyN is always used.
```

arokem (commented):
Looks great. I assume the change to defaults is so that the tests for RB pass? 

I just had a small suggestion for the documentation. If that makes sense, integrate that and then feel free to merge.

36000:
I am not sure why we set those defaults to such high values in the first place, now they are closer to recobundle's defaults.

Issue #473, opened 2020-09-23
Update CSD to new DIPY API


arokem:
To get rid of this warning: 

```
/Users/arokem/source/pyAFQ/AFQ/models/csd.py:34: DeprecationWarning: dipy.reconst.csdeconv.auto_response is deprecated, Please use dipy.reconst.csdeconv.auto_response_ssst instead
```

PR #474, opened 2020-09-24
Multi-shell, multi-tissue model

### LINKED ISSUE #367: Add support for MSMT
Need https://github.com/dipy/dipy/pull/2191 to land first

arokem:
Based on [this](https://dipy.org/documentation/1.2.0./examples_built/reconst_mcsd/#example-reconst-mcsd) example, adds a new model to the CSD module.  

Closes #367 and along the way also addresses #473.


arokem:
OK - I think that this is ready to go. @36000 : would you mind reviewing this when you get a chance?

arokem:
Looks like cvxpy has some issues with Python 3.8. I will investigate.

36000:
This looks great! Is everything squared away on Dipy's end?

arokem:
Yes. On further inspection, I don't think that this is a Python 3.8 issue specifically. Rather, I think we might be hitting something related to this PR: https://github.com/dipy/dipy/pull/2221 (short version: cvxpy sometimes fails in hard-to-predict ways) so it's resolved on DIPY and we should get that fix here once DIPY 1.3 is released. The question is whether we should tolerate it for now and hope it doesn't bite us too often, or patch this into _fixes.

36000:
I think we should patch this into fixes if we want to add this to 0.5 , so 0.5 is stable.

arokem:
OK - done in https://github.com/yeatmanlab/pyAFQ/pull/474/commits/955433bd876ae81a6fe9a2ccdfb38b59c0bfeb00.

36000:
This is good to merge?

arokem:
Yes

PR #475, opened 2020-09-24
Removes extraneous underscore.

arokem:
Should address failing nightly test, but the PR currently also includes testing for this, which should happen before merging.


arokem:
OK - seems to work. Merging.

Issue #443, opened 2020-09-18
Add an example with callosal groups


arokem:
Use the callosal templates stored in AFQ_data/callosum_templates, to demonstrate how the `plot_tract_profiles` example can be reworked to generate new bundles from other templates.

PR #476, opened 2020-09-25
ENH #443 Callosal Group Example

bloomdt-uw:
Baseline commit. There are several other enhancements requests and issues that will follow from work on this pull request. I think the main concern I have at the moment is whether the descriptive text is sufficiently accurate and any additional recommended text that would enhance the interpretive value the anatomical images, ROIs, or streamline density maps. I will add few comments in the pull request to address specific areas of interest and concern. From this conversation, I will open separate issues and enhancement requests where needed.

### DIFF
--- a/examples/plot_callosal_tract_profile.py
+++ b/examples/plot_callosal_tract_profile.py
@@ -1,0 +1,558 @@
+"""
+==============================
+Plotting Novel Tract Profiles:
+==============================
+
+The following is an example of tractometry for a novel bundle and plotting the
+resulting FA tract profile. We will run tractometry for the *anterior forceps*
+using waypoint ROIs.
+
+**AFQ Waypoint ROI Tractometry:**
+
+.. note::
+
+  This example uses the Yeatman et al. waypoint ROI approach, first
+  described in [Yeatman2012]_ and further elaborated in [Yeatman2014]_.
+
+  The waypoint ROIs used in this example are from the anterior frontal lobe of
+  the corpus callosum (AntFrontal). The waypoint ROIs are from the human corpus
+  callosum templates:
+
+    https://figshare.com/articles/Templates_for_Automated_Fiber_Quantification_of_corpus_callosum_from_Diffusion_MRI_data/3381523
+
+"""
+import os.path as op
+import matplotlib.pyplot as plt
+import numpy as np
+import nibabel as nib
+import dipy.data as dpd
+from dipy.data import fetcher
+from dipy.io.streamline import save_tractogram, load_tractogram
+from dipy.stats.analysis import afq_profile, gaussian_weights
+from dipy.io.stateful_tractogram import StatefulTractogram
+from dipy.io.stateful_tractogram import Space
+
+import AFQ.data as afd
+import AFQ.tractography as aft
+import AFQ.registration as reg
+import AFQ.models.dti as dti
+import AFQ.segmentation as seg
+from AFQ.utils.volume import patch_up_roi
+
+import logging
+import sys
+
+# Ensure segmentation logging information is included in this example's output
+root = logging.getLogger()
+root.setLevel(logging.ERROR)
+logging.getLogger('AFQ.Segmentation').setLevel(logging.INFO)
+logging.getLogger('AFQ.tractography').setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+root.addHandler(handler)
+
+# Monkey patch the progress bars so don't pollute this example's output
+import AFQ._fixes as fixes
+
+def notqdm(iterable, *args, **kwargs):
+    """
+    replacement for tqdm that just passes back the iterable to silence `tqdm`
+    """
+    return iterable
+
+
+afd.tqdm = notqdm
+fixes.tqdm = notqdm
+seg.tqdm = notqdm
+
+
+def update_progressbar(progress, total_length):
+    return
+
+
+fetcher.update_progressbar = update_progressbar
+
+
+# Target directory for this example's output files
+working_dir = "./callosal_tract_profile"
+
+##########################################################################
+# Helper functions:
+# -----------------
+
+
+def show_anatomical_slices(img_data, title):
+    """
+    display anatomical slices from midpoint
+
+    based on:
+    https://nipy.org/nibabel/coordinate_systems.html
+    """
+
+    axial_slice = img_data[:, :, int(img_data.shape[2] / 2)]
+    coronal_slice = img_data[:, int(img_data.shape[1] / 2), :]
+    sagittal_slice = img_data[int(img_data.shape[0] / 2), :, :]
+
+    fig = plt.figure(constrained_layout=False)
+    gs = fig.add_gridspec(nrows=3, ncols=2, wspace=0.01, hspace=0.01)
+    ax1 = fig.add_subplot(gs[:-1, :])
+    ax1.imshow(axial_slice.T, cmap="gray", origin="lower")
+    ax1.axis('off')
+    ax2 = fig.add_subplot(gs[2, 0])
+    ax2.imshow(coronal_slice.T, cmap="gray", origin="lower")
+    ax2.axis('off')
+    ax3 = fig.add_subplot(gs[2, 1])
+    ax3.imshow(sagittal_slice.T, cmap="gray", origin="lower")
+    ax3.axis('off')
+
+    plt.suptitle(title)
+    plt.show()
+
+# Imports for create_density_map
+
+
+from dipy.io.utils import (create_nifti_header, get_reference_info)
+from dipy.tracking.streamline import select_random_set_of_streamlines
+from dipy.tracking.utils import density_map
+
+
+def create_density_map(tractogram, fname):
+    """
+    Write streamline density maps.
+
+    based on:
+    https://dipy.org/documentation/1.1.1./examples_built/streamline_formats/
+    """
+
+    tractogram.to_vox()
+    streamlines = select_random_set_of_streamlines(tractogram.streamlines,
+                                                   1000)
+    affine, vol_dims, voxel_sizes, voxel_order = get_reference_info(tractogram)
+    tractogram_density = density_map(streamlines, np.eye(4), vol_dims)
+    nifti_header = create_nifti_header(affine, vol_dims, voxel_sizes)
+    density_map_img = nib.Nifti1Image(tractogram_density, affine, nifti_header)
+    nib.save(density_map_img, op.join(working_dir, fname))
+
+##########################################################################
+# Get example data:
+# -----------------
+# **Diffusion dataset**
+#
+# .. note::
+#   The diffusion data used in this example are from the Stanford High Angular
+#   Resolution Diffusion Imaging (HARDI) dataset:
+#
+#     https://purl.stanford.edu/ng782rw8378
+
+
+print("Fetching data...")
+
+# If does not already exist `fetch_stanford_hardi` will download the first
+# subject's session from the HARDI data into fetcher.dipy_home:
+# `~/.dipy/stanford_hardi`
+dpd.fetch_stanford_hardi()
+
+# Reference to data file locations
+hardi_dir = op.join(fetcher.dipy_home, "stanford_hardi")
+hardi_fdata = op.join(hardi_dir, "HARDI150.nii.gz")
+hardi_fbval = op.join(hardi_dir, "HARDI150.bval")
+hardi_fbvec = op.join(hardi_dir, "HARDI150.bvec")
+
+print(f'Loading data file: {hardi_fdata}')
+img = nib.load(hardi_fdata)
+
+# Display output from this step
+img_data = img.get_fdata()
+
+# working with space and time data
+img_data = img_data[..., int(img_data.shape[3] / 2)]
+
+show_anatomical_slices(img_data, 'HARDI 150 DWI')
+print(f'bvals: f{np.loadtxt(hardi_fbval)}')
+print(f'bvec: f{np.loadtxt(hardi_fbvec)}')
+
+
+##########################################################################
+# Calculate DTI:
+# --------------
+# Fit the DTI model using default settings, save files with derived maps.
+#
+# The by default following DTI measurements are calculated:
+#
+# - Fractional anisotropy (FA),
+#
+# - Mean diffusivity (MD),
+#
+# - Axial diffusivity (AD),
+#
+# - and Radial diffusivity (RD)
+#
+# In this example we will only use FA.
+#
+# .. note::
+#   By default:
+#
+#   - All b-values less than or equal to 50 are considered to be
+#     without diffusion weighting.
+#
+#   - No binary masks are applied; therefore all voxels are processed.
+#
+# .. note::
+#   The diffusion tensor imaging parameters contain the associated eigenvalues
+#   and eigenvectors from eigen decomposition on the diffusion tensor.
+
+
+print("Calculating DTI...")
+
+if not op.exists(op.join(working_dir, 'dti_FA.nii.gz')):
+    dti_params = dti.fit_dti(hardi_fdata, hardi_fbval, hardi_fbvec,
+                             out_dir=working_dir)
+else:
+    dti_params = {'FA': op.join(working_dir, 'dti_FA.nii.gz'),
+                  'params': op.join(working_dir, 'dti_params.nii.gz')}
+
+print(f"Loading {dti_params['FA']}")
+FA_img = nib.load(dti_params['FA'])
+FA_data = FA_img.get_fdata()
+
+show_anatomical_slices(FA_data, 'Fractional Anisotropy (FA)')
+
+##########################################################################
+# Register the individual data to a template:
+# -------------------------------------------
+# For the purpose of bundle segmentation, the individual brain is registered
+# to the MNI T2 template. The waypoint ROIs used in segmentation are then each
+# brought into each subject's native space to test streamlines for whether they
+# fulfill the segmentation criteria.
+#
+# .. note::
+#    Here, for simplicity we only perform a non-linear transformation between
+#    the individual's brain and the MNI T2 template. This is likely why the
+#    occipital/parietal junction appears concave. In practice, it's a good idea
+#    to also perform a pre-alignment using an affine transformation. Moreover,
+#    by default, registration uses APM (Anisotropic Power Map) which also tends
+#    to provide better results.
+
+
+print("Registering to template...")
+
+MNI_T2_img = afd.read_mni_template()
+nib.save(MNI_T2_img, op.join(working_dir, 'MNI_T2.nii.gz'))
+show_anatomical_slices(MNI_T2_img.get_fdata(), 'MNI T2 Template')
+
+# create a diffeomorphic map from subject to MNI space
+if not op.exists(op.join(working_dir, 'mapping.nii.gz')):
+    import dipy.core.gradients as dpg
+    gtab = dpg.gradient_table(hardi_fbval, hardi_fbvec)
+
+    warped_hardi, mapping = reg.syn_register_dwi(hardi_fdata, gtab)
+    reg.write_mapping(mapping, op.join(working_dir, 'mapping.nii.gz'))
+else:
+    mapping = reg.read_mapping(op.join(working_dir, 'mapping.nii.gz'),
+                               img,
+                               MNI_T2_img)
+
+mapping_img = nib.load(op.join(working_dir, 'mapping.nii.gz'))
+mapping_img_data = mapping_img.get_fdata()
+
+# Working with diffeomorphic map data
+mapping_img_data = mapping_img_data[..., 0, 0]
+show_anatomical_slices(mapping_img_data, 'Registration Displacement Mapping')
+
+# plot the transformational map of MNI T2 onto subject space
+warped_MNI_T2_data = mapping.transform_inverse(MNI_T2_img.get_fdata())
+warped_MNI_T2_img = nib.Nifti1Image(warped_MNI_T2_data.astype(float),
+                                    img.affine)
+
+nib.save(warped_MNI_T2_img, op.join(working_dir, 'warped_MNI_T2.nii.gz'))
+
+show_anatomical_slices(warped_MNI_T2_img.get_fdata(), 'Warped MNI T2')
+
+##########################################################################
+# Create novel bundle specification:
+# ----------------------------------
+# The bundles specify meta-data for the segmentation.
+#
+# The following keys are required for a `bundle` entry:
+#
+# - `ROIs`
+#
+#   lists the ROI templates.
+#
+# .. note::
+#   Order of the `ROIs` matters and may result in different tract profiles
+#
+# - `rules`
+#
+#   label each ROI as inclusionary `True` or exclusionary `False`.
+#
+# - `cross_midline`
+#
+#   whether or not the streamline crosses the midline
+#
+# .. note::
+#
+#   It is also possible to utilize probablity maps to further refine ROIs. If
+#   `prob_map` key is not specified the probablities will all be ones and same
+#   shape as the ROI.
+#
+# .. note::
+#
+#   If using multiple bundles it is recommended to add an optional identifier
+#   key such as `uid`. This key is not used by segmentation, but can be helpful
+#   for debugging or quality control reports.
+#
+# To identify candidate streamlines for the anterior forceps we will use three
+# waypoint ROI templates:
+#
+# 1. Left anterior frontal,
+#
+# 2. Right anterior frontal,
+#
+# 3. and midsaggital.
+#
+# The templates are first resampled into the MNI space, before
+# they are brought into the subject's individual native space.
+
+
+print("Fetching callosum ROI templates...")
+
+callosal_templates = afd.read_callosum_templates(resample_to=MNI_T2_img)
+
+show_anatomical_slices(callosal_templates["L_AntFrontal"].get_fdata(),
+                       'Left anterior frontal ROI')
+show_anatomical_slices(callosal_templates["R_AntFrontal"].get_fdata(),
+                       'Right anterior frontal ROI')
+show_anatomical_slices(callosal_templates["Callosum_midsag"].get_fdata(),
+                       'Midsagittal ROI')
+
+print("Creating callosal bundle specification...")
+
+# bundle dict
+bundles = {}
+
+# anterior frontal ROIs
+bundles["AntFrontal"] = {
+    'ROIs': [callosal_templates["L_AntFrontal"],
+             callosal_templates["R_AntFrontal"],
+             callosal_templates["Callosum_midsag"]],
+    'rules': [True, True, True],
+    'cross_midline': True
+}
+
+##########################################################################
+# Tracking:
+# ---------
+# Streamlines are generate using DTI and a deterministic tractography
+# algorithm. For speed, we seed only within the waypoint ROIs for each bundle.
+#
+# .. note::
+#   By default tractography:
+#
+#   - Will identify streamlines with lengths between 10 mm and 1 m, with
+#     turning angles of less than 30.
+#
+#   - Is seeded with a single seed in each voxel on each dimension
+#
+#   - Each step is 0.5 mm
+#
+# .. note::
+#   In this example tractography results in a large number of candidate
+#   streamlines for the anterior forceps.
+
+
+print("Tracking...")
+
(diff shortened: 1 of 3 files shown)


examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+
+
+fetcher.update_progressbar = update_progressbar
+
+
+# Target directory for this example's output files
+working_dir = "./callosal_tract_profile"
+
+##########################################################################
+# Helper functions:
```
bloomdt-uw, line 80: consider incorporating both helper functions show_anatomical_slices and create_density_map into viz package

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+# -----------------
+
+
+def show_anatomical_slices(img_data, title):
```
bloomdt-uw, line 84: Replace matplotlib usage with alternative plotting library:

1. use plotly to be consistent with our move towards single plotting library
2. use papaya (suggestion from @richford)
3. consider for brainsprite interactive visualization (suggestion from @richford)

I think brainsprite is probably the right way to go for the examples

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+else:
+    mapping = reg.read_mapping(op.join(working_dir, 'mapping.nii.gz'),
+                               img,
+                               MNI_T2_img)
+
+mapping_img = nib.load(op.join(working_dir, 'mapping.nii.gz'))
+mapping_img_data = mapping_img.get_fdata()
+
+# Working with diffeomorphic map data
+mapping_img_data = mapping_img_data[..., 0, 0]
```
bloomdt-uw, line 259: Only plotting one dimension from the displacement map. It contains a 3D vector and could modify vizualization to map xyz to RGB (suggestion from @arokem)

examples/plot_callosal_tract_profile.py
bloomdt-uw, line 80: This may already exist in other packages but consider additional helper function display statistics about image data.

Motivation here was the different Nifti files in this example have different ranges for data values:
- (0, 1000)            hardi150.nii.gz
- (0, 1)                  dti_FA.nii.gz, MNI_T2.nii.gz
- (0.001, 0.003)   dti_MD.nii.gz, dti_AD.nii.gz, dti_RD.nii.gz
- (-7, 7)                 mapping.nii.gz

Providing this and other statistics may be useful for end users when using external viewers like Mango and MI-Brain

bloomdt-uw, line 80: Quality Control Report
- Consider incorporating anatomical slices and density map functionality as part of quality control report (from @36000)
- Consider adding a visualization that builds on the streamline density maps, which plots contours for each individual in MNI space (from @jyeatman)
- Consider placing ROIs and streamline density maps on anatomical image (from @jyeatman and @arokem)
- Consider plotting ODF (orbital distribution function) as part of QC report (from @richford)

NOTE as far as I know the quality control report functionality does not exist yet, but perhaps there is an existing enhancement request which can attach these additional enhancements.

bloomdt-uw: There is also some overlap with whether these graphics should be integrated with the pyAFQ pipeline.

NOTE In my mind this might be something to consider in tandem with the work on nipype integration

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+
+# Imports for create_density_map
+
+
+from dipy.io.utils import (create_nifti_header, get_reference_info)
+from dipy.tracking.streamline import select_random_set_of_streamlines
+from dipy.tracking.utils import density_map
+
+
+def create_density_map(tractogram, fname):
```
bloomdt-uw, line 119: Consider streamline density map diff. This was one of the first considerations I had when review output. In discussion with the group seems like we could go multiple directions. Here's summary:

from @arokem 

> The main use for visitation maps or streamline density maps is in measuring the volume of certain tracts (i.e., the number of voxels visited) and in comparing two segmentations of the same brain to each other.

One approach is to use Dice's coefficient
> We would use the two visitation maps to do calculate a Dice coefficient between the two measurements/segmentations, which gives as an indication of the test-retest reliability of the segmentation in this case.

Other approaches worth mentioning that address spirit of streamline differences
- leveraging return idx and comparing (from @36000)
- comparing tract profiles (from @36000)
- utilizing convex hull comparison (from @richford)
- alpha shapes in plotly (from @richford)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+    """
+    Write streamline density maps.
+
+    based on:
+    https://dipy.org/documentation/1.1.1./examples_built/streamline_formats/
+    """
+
+    tractogram.to_vox()
+    streamlines = select_random_set_of_streamlines(tractogram.streamlines,
+                                                   1000)
```
bloomdt-uw, line 129: Use all stream lines rather than a random sample (from @jyeatman)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+    affine, vol_dims, voxel_sizes, voxel_order = get_reference_info(tractogram)
+    tractogram_density = density_map(streamlines, np.eye(4), vol_dims)
```
bloomdt-uw, line 131: Normalize maximum values to 1 (from @jyeatman)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+#
+# .. note::
+#   The diffusion tensor imaging parameters contain the associated eigenvalues
+#   and eigenvectors from eigen decomposition on the diffusion tensor.
+
+
+print("Calculating DTI...")
+
+if not op.exists(op.join(working_dir, 'dti_FA.nii.gz')):
+    dti_params = dti.fit_dti(hardi_fdata, hardi_fbval, hardi_fbvec,
```
bloomdt-uw, line 208: fit_dti calculates all the metrics and generates Nifty files for each one. While the examples are not the main use cases might want to add capability for end users to specify which metrics are calculated.

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+# brought into each subject's native space to test streamlines for whether they
+# fulfill the segmentation criteria.
+#
+# .. note::
+#    Here, for simplicity we only perform a non-linear transformation between
+#    the individual's brain and the MNI T2 template. This is likely why the
+#    occipital/parietal junction appears concave. In practice, it's a good idea
+#    to also perform a pre-alignment using an affine transformation. Moreover,
+#    by default, registration uses APM (Anisotropic Power Map) which also tends
+#    to provide better results.
```
bloomdt-uw, line 234: Add a registration example to demonstrate effect of different registrations on process

1. using alternative (non-MNI T2) registration template
2. including/excluding pre-alignment and affine transformation
3. using APM for registration (see models/csd.py fit_anisotropic_power_map())

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+    mapping = reg.read_mapping(op.join(working_dir, 'mapping.nii.gz'),
+                               img,
+                               MNI_T2_img)
+
+mapping_img = nib.load(op.join(working_dir, 'mapping.nii.gz'))
+mapping_img_data = mapping_img.get_fdata()
+
+# Working with diffeomorphic map data
+mapping_img_data = mapping_img_data[..., 0, 0]
+show_anatomical_slices(mapping_img_data, 'Registration Displacement Mapping')
```
bloomdt-uw, line 260: @arokem Is there a better more meaningful name for mapping file?

MNI T2 to Subject B0 Registration Symmetric Normalization Displacement Mapping

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+# The bundles specify meta-data for the segmentation.
+#
+# The following keys are required for a `bundle` entry:
+#
+# - `ROIs`
+#
+#   lists the ROI templates.
+#
+# .. note::
+#   Order of the `ROIs` matters and may result in different tract profiles
```
bloomdt-uw, line 283: @arokem I know we discussed why order matters, but I didn't follow closely enough to provide details here

I tested flipping the left and right ROIs and returned similar but different streamlines (was motivation for the diff request above)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+                       'Right anterior frontal ROI')
+show_anatomical_slices(callosal_templates["Callosum_midsag"].get_fdata(),
+                       'Midsagittal ROI')
+
+print("Creating callosal bundle specification...")
+
+# bundle dict
+bundles = {}
+
+# anterior frontal ROIs
```
bloomdt-uw, line 334: Incorporate bundle specification definitions for all callosum templates into api.

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+#    streamlines may be included that would otherwise be excluded.
+
+
+tractogram_img = nib.load(op.join(working_dir, 'afq_dti_density_map.nii.gz'))
+show_anatomical_slices(tractogram_img.get_fdata(), 'DTI Density Map')
+
+print("Segmenting fiber groups...")
+
+segmentation = seg.Segmentation(return_idx=True,
+                                filter_by_endpoints=False)
```
bloomdt-uw, line 448: Currently bundles_to_all hard codes the endpoints. Should be able to have users define as necessary for custom ROIs.

bloomdt-uw, line 448: If define custom ROI and do not set filter_by_endpoints=False then segmentation crashes. Could fail gracefully or ideally should warn that segmentation can't find and proceed with processing without endpoints (suggestion from @jyeatman).

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+
+                # Add voxels that aren't there yet:
+                seed_roi = np.logical_or(seed_roi, warped_roi)
+
+    seed_roi_img = nib.Nifti1Image(seed_roi.astype(float), img.affine)
+    nib.save(seed_roi_img, op.join(working_dir, 'seed_roi.nii.gz'))
+
+    show_anatomical_slices(seed_roi_img.get_fdata(), 'Seed ROI')
+
+    tractogram = aft.track(dti_params['params'], seed_mask=seed_roi,
```
bloomdt-uw, line 393: I have noticed that the docstring are out of date for default parameters for both track and segmentation (perhaps more wide spread)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+# - Axial diffusivity (AD),
+#
+# - and Radial diffusivity (RD)
+#
+# In this example we will only use FA.
+#
+# .. note::
+#   By default:
+#
+#   - All b-values less than or equal to 50 are considered to be
```
bloomdt-uw, line 195: The default b0-threshold is 0 for fit_dti and 50 for segmentation. Should set to 50 everywhere (from discussion with @36000 and @arokem)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+
+def update_progressbar(progress, total_length):
+    return
+
+
+fetcher.update_progressbar = update_progressbar
+
+
+# Target directory for this example's output files
+working_dir = "./callosal_tract_profile"
```
bloomdt-uw, line 77: All examples should have their own output directories to avoid file name collisions.

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
+"""
```
bloomdt-uw, line 1: sphinx build make clean should clean up after any sphinx-gallery examples files and folders:

- multiple files are placed directly into the `examples` folder (this are ignored by git)
- application data folders including `.dipy` and `AFQ_data` (these are not part of source)

Depending on what files are present this may impact local builds

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+print(f'Loading data file: {hardi_fdata}')
+img = nib.load(hardi_fdata)
+
+# Display output from this step
+img_data = img.get_fdata()
+
+# working with space and time data
+img_data = img_data[..., int(img_data.shape[3] / 2)]
+
+show_anatomical_slices(img_data, 'HARDI 150 DWI')
```
bloomdt-uw, line 170: I am wondering what if any interpretive text can be added for the subjects anatomical image

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+                             out_dir=working_dir)
+else:
+    dti_params = {'FA': op.join(working_dir, 'dti_FA.nii.gz'),
+                  'params': op.join(working_dir, 'dti_params.nii.gz')}
+
+print(f"Loading {dti_params['FA']}")
+FA_img = nib.load(dti_params['FA'])
+FA_data = FA_img.get_fdata()
+
+show_anatomical_slices(FA_data, 'Fractional Anisotropy (FA)')
```
bloomdt-uw, line 218: What interpretive text can be added for FA?

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+show_anatomical_slices(mapping_img_data, 'Registration Displacement Mapping')
+
+# plot the transformational map of MNI T2 onto subject space
+warped_MNI_T2_data = mapping.transform_inverse(MNI_T2_img.get_fdata())
+warped_MNI_T2_img = nib.Nifti1Image(warped_MNI_T2_data.astype(float),
+                                    img.affine)
+
+nib.save(warped_MNI_T2_img, op.join(working_dir, 'warped_MNI_T2.nii.gz'))
+
+show_anatomical_slices(warped_MNI_T2_img.get_fdata(), 'Warped MNI T2')
```
bloomdt-uw, line 269: Any additional interpretive comments for MNI T2 image space or the individual subjects space

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+                warped_roi_img = nib.load(op.join(working_dir,
+                                                  f"{bundle}_{idx+1}.nii.gz"))
+                show_anatomical_slices(warped_roi_img.get_fdata(),
+                                       f'warped {bundle}_{idx+1} ROI')
+
+                # Add voxels that aren't there yet:
+                seed_roi = np.logical_or(seed_roi, warped_roi)
+
+    seed_roi_img = nib.Nifti1Image(seed_roi.astype(float), img.affine)
+    nib.save(seed_roi_img, op.join(working_dir, 'seed_roi.nii.gz'))
```
bloomdt-uw, line 389: Do we want to say anything about the seed?

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+                                    img,
+                                    Space.VOX)
+    tractogram.to_rasmm()
+    save_tractogram(tractogram, op.join(working_dir, f'afq_{bundle}_seg.trk'),
+                    bbox_valid_check=False)
+
+    create_density_map(tractogram, f'afq_{bundle}_seg_density_map.nii.gz')
+    tractogram_img = nib.load(op.join(working_dir,
+                                      f'afq_{bundle}_seg_density_map.nii.gz'))
+    show_anatomical_slices(tractogram_img.get_fdata(),
```
bloomdt-uw, line 471: Any final thoughts for the various streamline density maps (tracking, segmentation, and cleaned)

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+#   - No binary masks are applied; therefore all voxels are processed.
+#
+# .. note::
+#   The diffusion tensor imaging parameters contain the associated eigenvalues
+#   and eigenvectors from eigen decomposition on the diffusion tensor.
+
+
+print("Calculating DTI...")
+
+if not op.exists(op.join(working_dir, 'dti_FA.nii.gz')):
```
bloomdt-uw, line 207: One additional suggestion is to use BIDS metadata, and not file name, to determine whether to regenerate output

examples/plot_callosal_tract_profile.py
bloomdt-uw, line 1: another enhancement regarding local sphinx-build would be to allow single example as target (since some of the example execution times are long)

To build only one file I locally updated `conf.py` by adding `filename_pattern`

```
sphinx_gallery_conf = {
     # path to your examples scripts
     'examples_dirs': '../../examples',
     **'filename_pattern': '/plot_callosal_tract_profile.py',**
     # path where to save gallery generated examples
     'gallery_dirs': 'auto_examples',
     'image_scrapers': image_scrapers,
}
```

bloomdt-uw, line 1: @arokem I believe that you mentioned in passing that this example might be relevant to baby-AFQ wondering if you could elaborate and whether we would want to share with Mareike

examples/plot_callosal_tract_profile.py
bloomdt-uw: @jyeatman I may be misstating this, but believe you mentioned testing out using APM in bundle classification instead of FA. If this is accurate I was wondering if could elaborate.

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+print(f'bvals: f{np.loadtxt(hardi_fbval)}')
+print(f'bvec: f{np.loadtxt(hardi_fbvec)}')
+
+
+##########################################################################
+# Calculate DTI:
+# --------------
+# Fit the DTI model using default settings, save files with derived maps.
+#
+# The by default following DTI measurements are calculated:
```
richford, line 180:
```suggestion
# By default following DTI measurements are calculated:
```

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+#
+# In this example we will only use FA.
+#
+# .. note::
+#   By default:
+#
+#   - All b-values less than or equal to 50 are considered to be
+#     without diffusion weighting.
+#
+#   - No binary masks are applied; therefore all voxels are processed.
```
richford, line 198: I see this note here addresses the issue that I mentioned on the Zoom chat this morning. Great!

examples/plot_callosal_tract_profile.py
36000: Both these functions are going to be super useful. I think show_anatomical_slices belongs in viz/utils.py . For create_density_map, I could see it going into tractography.py if its used for more things than just visualization (like comparisons based on the dice coefficient). It  might also belong in its own file.

examples/plot_callosal_tract_profile.py
36000: For non-interactive plots we just switched to using seaborn. For interactive plots, if they don't require any installs beyond pip install, then its good. brainsprite looks good to me!

examples/plot_callosal_tract_profile.py
36000: For integration with the AFQ object, the general layout would look like this, with create_density_map as an example:

// creates a density map for this row if it does not already exist
// remember a row is a specific session for a specific subject
// should be a wrapper around the create_density_map function that handles filenames / meta json file / checking if the file exists
def _create_density_map(self, row): 

def get_create_density_map(): // calls _create_density_map  for every row and returns the resulting files as a list

def set_create_density_map(): // calls _create_density_map  for every row

Most AFQ object functions follow this format. Some have some functionality that should probably be extracted out into their own functions. Also I bet we could reduce a lot of code by making decorators for these types of functions. And, if the get / set functions are not documented somewhere, they should be.

examples/plot_callosal_tract_profile.py
arokem: Yes. See #485

arokem: OK - this is now fleshed out more fully in #487

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,558 @@
...
+logging.getLogger('AFQ.Segmentation').setLevel(logging.INFO)
+logging.getLogger('AFQ.tractography').setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+root.addHandler(handler)
+
+# Monkey patch the progress bars so don't pollute this example's output
+import AFQ._fixes as fixes
+
+def notqdm(iterable, *args, **kwargs):
```
arokem, line 57: I wonder if there's a more elegant way to set up the sphinx config to do this for us, instead of very explicitly doing it here in the example. 

I imagine that if you are running this as a user (say, from the command line) you do want to see the progbar. It's only in the sphinx build context that we don't want it.

examples/plot_callosal_tract_profile.py
arokem: 👍

examples/plot_callosal_tract_profile.py
arokem: I believe that @36000 has integrated this into #484

arokem (commented):
This is far as I got through now, but will continue to review more later.

examples/plot_callosal_tract_profile.py
bloomdt-uw: addressed in #490

examples/plot_callosal_tract_profile.py
bloomdt-uw: Once #492 has been merged I will remove the monkey patch

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressed in #484

examples/plot_callosal_tract_profile.py
bloomdt-uw: I'll switch `create_density_map` calls to use `AFQ.utils.volume.density_map`. 

However #484 does not address moving `show_anatomical_slices`.

examples/plot_callosal_tract_profile.py
bloomdt-uw: opened #493

examples/plot_callosal_tract_profile.py
bloomdt-uw: Added as enhancement request #494

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressed in #497

bloomdt-uw: Addressed in #498

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressing in #500

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressing in #500

examples/plot_callosal_tract_profile.py
bloomdt-uw: For items 1 and potentially 2 see #487

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressed in #502

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,536 @@
...
+
+# Target directory for this example's output files
+working_dir = "./callosal_tract_profile"
+
+##########################################################################
+# Helper functions:
+# -----------------
+
+
+def show_anatomical_slices(img_data, title):
```
arokem, line 84: See #500

examples/plot_callosal_tract_profile.py
arokem: This is something we should think more thoroughly when considering how to refactor to a nipype workflow.

examples/plot_callosal_tract_profile.py
arokem: Addressed through issue: #505

examples/plot_callosal_tract_profile.py
arokem: You could say "This is an FA map" :-)

examples/plot_callosal_tract_profile.py
arokem: This is pretty good.

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,536 @@
...
+#
+#   label each ROI as inclusionary `True` or exclusionary `False`.
+#
+# - `cross_midline`
+#
+#   whether or not the streamline crosses the midline
+#
+# .. note::
+#
+#   It is also possible to utilize probablity maps to further refine ROIs. If
```
arokem, line 271:
```suggestion
#   It is also possible to utilize probablity maps to further refine the segmentation. If
```

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,536 @@
...
+# Tracking:
+# ---------
+# Streamlines are generate using DTI and a deterministic tractography
+# algorithm. For speed, we seed only within the waypoint ROIs for each bundle.
+#
+# .. note::
+#   By default tractography:
+#
+#   - Will identify streamlines with lengths between 10 mm and 1 m, with
+#     turning angles of less than 30.
```
arokem, line 329:
```suggestion
#     turning angles of less than 30 degrees.
```

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,536 @@
...
+#
+#   - Is seeded with a single seed in each voxel on each dimension
+#
+#   - Each step is 0.5 mm
+#
+# .. note::
+#   In this example tractography results in a large number of candidate
+#   streamlines for the anterior forceps.
```
arokem, line 337:
```suggestion
#   streamlines for the anterior forceps, but not many streamlines anywhere else.
```

examples/plot_callosal_tract_profile.py
```diff
@@ -0,0 +1,536 @@
...
+    tractogram.to_rasmm()
+    save_tractogram(tractogram, op.join(working_dir, f'afq_{bundle}.trk'),
+                    bbox_valid_check=False)
+
+    tractogram_img = density_map(tractogram, n_sls=1000, to_vox=True)
+    nib.save(tractogram_img, op.join(working_dir,
+                                     f'afq_{bundle}_density_map.nii.gz'))
+    show_anatomical_slices(tractogram_img.get_fdata(),
+                           f'Cleaned {bundle} Density Map')
+
```
arokem, line 497: Maybe add an html/plotly visualization with the resulting segmented bundle?

arokem: Use AFQ.viz.plotly_backend function for showing streamlines. This should generate an HTML file that you can then incorporate here using: 

    plotly.io.show(html_file_name)

examples/plot_callosal_tract_profile.py
bloomdt-uw: Why order sequence of ROIs in bundle matters:
Given sequence of ROIs: (e.g., A, B, C)
The waypoint ROIs endpoints should be first two (i.e., A, C, B)
Note that directionality doesn't matter, (i.e. A <-> C: A ->C or C -> A); however want to be consistent with with other streamlines

NOTE: aal endpoints interact as well, but we side step in this example since not using endpoints

arokem:
Looks like a few other commits got convolved in with your commits. Could you please rebase?

examples/plot_callosal_tract_profile.py
bloomdt-uw: Addressed, but depends on #513

examples/plot_callosal_tract_profile.py
bloomdt-uw: See #517

examples/plot_callosal_tract_profile.py
arokem: Looks like that's been merged. Could you please rebase on top of master one more time?

arokem: Oh - nevermind. Looks like it's working. I'll go ahead and merge.

PR #478, opened 2020-09-28
Apply brain mask to subject img before registration

36000:
Should we be doing this?


jyeatman:
applying the brain mask before registration is a good thing. Because you don't want the scull/scalp to be used in registration. You only care the the structures of the brain align and, in many cases, the scull/scalp is a high contrast feature of the image that can negatively affect the registration

arokem:
👍

PR #482, opened 2020-09-30
FIX: Downsample number of streamlines vizzed down to 200 when vizzing all bundles

36000:
- [x] Generate an example

### DIFF
--- a/AFQ/viz/utils.py
+++ b/AFQ/viz/utils.py
@@ -131,7 +131,8 @@ def bundle_selector(bundle_dict, colors, b):
     return color, b_name
 
 
-def tract_generator(sft, affine, bundle, bundle_dict, colors, n_points):
+def tract_generator(sft, affine, bundle, bundle_dict, colors, n_points,
+                    n_sls=200):
     """
     Generates bundles of streamlines from the tractogram.
     Only generates from relevant bundle if bundle is set.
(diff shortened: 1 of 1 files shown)


arokem:
This looks good to me. 

Did you generate an example? Does it look OK?

36000:
image
Looks good to me.

arokem:
The thing I worry about is that people might use this as a diagnostic for how many streamlines are generated for each track. And in the best case, there will be an equal number of streamlines (200) in each group. Could we do something a bit more subtle that takes a total number of streamlines, proportionally allocated to each bundle? Or a fixed proportion per bundle?

36000:
The flip side is when bundles are sparse I don't want to get rid of even more streamlines. So we might need to have a minimum amount of streamlines per bundle. And that number should be high enough to get a good idea for what the streamlines are doing. Maybe, for every bundle, at least 100 are displayed, then the rest are proportionally distributed?

I should also note that the html we make per bundle have all the streamlines in them.

36000:
image
This does look a little better. All the streamline counts are proportionally distributed from n_sls_per_bundle - n_sls_per_bundle/2 to n_sls_per_bundle + n_sls_per_bundle/2

AFQ/viz/utils.py
```diff
@@ -203,8 +212,15 @@ def tract_generator(sft, affine, bundle, bundle_dict, colors, n_points):
         if bundle is None:
             # No selection: visualize all of them:
 
-            for b in np.unique(sft.data_per_streamline['bundle']):
+            bundles = np.unique(sft.data_per_streamline['bundle'])
+            n_sls = len(sft.streamlines)
+            mean_n_sl_per_bundle_viz = n_sls_viz // len(bundles)
+            for b in bundles:
                 idx = np.where(sft.data_per_streamline['bundle'] == b)[0]
+                n_sl_viz = (len(idx) * mean_n_sl_per_bundle_viz) // n_sls\
```
36000, line 220: mean_n_sl_per_bundle_viz here should be something else

36000:

PR #483, opened 2020-09-30
ENH: Return contast index dataframe from contrast_index

36000:


arokem:
I need a little bit  more context to understand what the objective of this is.

36000:
we can discuss this during the meeting

36000:
Function should just return a df that the user can ask questions with.

PR #484, opened 2020-10-01
ENH: Dice coeff

36000:
Compute a dice coefficient using density maps from #476

### DIFF
--- a/AFQ/tractography.py
+++ b/AFQ/tractography.py
@@ -3,8 +3,6 @@ import nibabel as nib
 import dipy.reconst.shm as shm
 import logging
 
-from scipy.spatial.distance import dice
-
 import dipy.data as dpd
 from dipy.direction import (DeterministicMaximumDirectionGetter,
                             ProbabilisticDirectionGetter)
@@ -13,9 +11,6 @@ from dipy.io.stateful_tractogram import StatefulTractogram, Space
 from dipy.tracking.stopping_criterion import (ThresholdStoppingCriterion,
                                               CmcStoppingCriterion,
                                               ActStoppingCriterion)
-from dipy.io.utils import (create_nifti_header, get_reference_info)
-from dipy.tracking.streamline import select_random_set_of_streamlines
-from dipy.tracking.utils import density_map
 
 
 from AFQ._fixes import (VerboseLocalTracking, VerboseParticleFilteringTracking,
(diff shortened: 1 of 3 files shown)


AFQ/tractography.py
```diff
@@ -237,3 +242,62 @@ def _tracking(tracker, seeds, dg, stopping_criterion, params_img,
...
+    tractogram_density = density_map(sls, np.eye(4), vol_dims)
+    nifti_header = create_nifti_header(affine, vol_dims, voxel_sizes)
+    density_map_img = nib.Nifti1Image(tractogram_density, affine, nifti_header)
+
+    return density_map_img
+
+
+def compute_dice_similarity_coefficient(density_map_img1, density_map_img2):
+    """
+    Compute the Dice dissimilarity between two density maps
```
36000, line 287: Compute Dice's coefficient between two density maps using scipy's Dice.

AFQ/tractography.py
```diff
@@ -237,3 +242,62 @@ def _tracking(tracker, seeds, dg, stopping_criterion, params_img,
...
+    ----------
+    density_map_img1 : Nifti1Image
+        One density map to compare.
+    density_map_img2 : Nifti1Image
+        The other density map to compare.
+
+    Returns
+    -------
+    The dice similarity between the density maps.
+    """
```
36000, line 300: comment on why 1 -

AFQ/tractography.py
```diff
@@ -237,3 +242,62 @@ def _tracking(tracker, seeds, dg, stopping_criterion, params_img,
...
+    return 1 - dice(
```
arokem, line 301:
```suggestion
    # scipy's dice function returns the dice *dissimilarity*
    return 1 - dice(
```

arokem:
At least the Dice computation should be in utils/volume.py in my opinion. 

Maybe even both. The challenge with that is that then you're introducing non-trivial dependencies into utils/volume.py

arokem:
On further thought, I think that's fine. Please move to utils/volume.py

36000:
regression testing on create_density_map

36000:
@arokem ready for review / merge

AFQ/utils/tests/test_volume.py
```diff
@@ -1,7 +1,15 @@
...
 import pytest
+import os.path as op
+
+import nibabel as nib
+
+from dipy.io.stateful_tractogram import Space
+from dipy.io.streamline import StatefulTractogram
 
 import AFQ.utils.volume as AFV
+import AFQ.data as AFD
```
arokem, line 12: These should be  imported as lower-case (sorry for missing this before!)

```suggestion
import AFQ.utils.volume as afv
import AFQ.data as afd
```

AFQ/utils/volume.py
```diff
@@ -42,3 +49,62 @@ def patch_up_roi(roi, bundle_name="ROI"):
...
+    img2 : Nifti1Image
+        The other image to compare.
+
+    Returns
+    -------
+    The dice similarity between the images.
+    """
+    # scipy's dice function returns the dice *dissimilarity*
+    return 1 - dice(
+        img1.get_fdata().flatten().astype(bool),
```
arokem, line 109: This assumes that the input is a nibabel image, but I think that it would be nice to allow passing full path, nibabel image or array. You can check which you got on the input and then proceed as necessary: if it's an array, you go ahead and  calculate Dice. If it's a nibabel image, you need to call `get_fdata` and then calculate Dice. If it's a string, you read the  file, call  `get_fdata` and then calculate Dice.

arokem (commented):
A couple more comments  here.

36000:

AFQ/utils/volume.py
```diff
@@ -42,3 +49,74 @@ def patch_up_roi(roi, bundle_name="ROI"):
...
+    tractogram_density = dtu.density_map(sls, np.eye(4), vol_dims)
+    nifti_header = create_nifti_header(affine, vol_dims, voxel_sizes)
+    density_map_img = nib.Nifti1Image(tractogram_density, affine, nifti_header)
+
+    return density_map_img
+
+
+def dice_coeff(arr1, arr2):
+    """
+    Compute Dice's coefficient between two imagess.
```
arokem, line 94:
```suggestion
    Compute Dice's coefficient between two images.
```

PR #489, opened 2020-10-02
enh adding distclean and realclean targets for sphinx build

bloomdt-uw:
addressing enhancement request #485

PR #490, opened 2020-10-03
adding separate example output directories to avoid name collisions

bloomdt-uw:
This is related to [observation](https://github.com/yeatmanlab/pyAFQ/pull/476/files#r494815555 ) from #443 review. Updating RecoBundle and tract profile examples to place artifacts in corresponding subdirectory. Note that AFQ API does not produce any artifacts in the `examples` directory and therefore was not changed as part of this work.. For now just placed a `working_dir` variable in both examples. This should work for both build and local execution.

Figured this was sufficient for now, but I briefly considered passing the working directory variable as part of build process [see](https://sphinx-gallery.github.io/stable/configuration.html#passing-command-line-arguments-to-example-scripts).

### DIFF
--- a/examples/plot_recobundles.py
+++ b/examples/plot_recobundles.py
@@ -34,6 +34,9 @@ import AFQ.models.dti as dti
 import AFQ.segmentation as seg
 import AFQ.api as api
 
+# Target directory for this example's output files
+working_dir = "./recobundles"
+
 dpd.fetch_stanford_hardi()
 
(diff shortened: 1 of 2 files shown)


bloomdt-uw:
May want to adjust sphinx-gallery `distclean` target to remove these subfolders. see #489 at the moment it will remove the artifacts, but leave the empty folders.

arokem:
Thanks for doing this. I'll merge this now, and we can adjust around it.

PR #515, opened 2020-10-06
FIX: update reg_subject arg checking to include dicts

36000:
This should fix nightly test 4


arokem:
So this was real bug!

36000:
Theres also an OOM error in nightly 3, that was preventing us from seeing this real bug.

arokem:
In other words, even our bugs have bugs 😆

PR #520, opened 2020-10-07
BF: Reset the input tractogram space to what you originally got.

### LINKED ISSUE #518: Segmentation alters input data space
Moved from #399 : 

@soichih says: We recently noticed that AFQ.segment function will modify the input streamline's space. For example, we feed tg.space set to rasmm and pass it to segment(). After the segmentation runs, the tg.space is now set to VOX. We basically had to reload the whole streamline from the disk to revert it back to the rasmm space. Is this an expected behavior?

@giulia-berto made this PR to correct this

brainlife/app-pyafq-segment#1

arokem:
Closes #518.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -191,6 +191,7 @@ class Segmentation:
         else:
             self.tg = tg
         self._tg_orig_space = self.tg.space
+        return tg
 
     def segment(self, bundle_dict, tg, fdata=None, fbval=None,
                 fbvec=None, mapping=None, reg_prealign=None,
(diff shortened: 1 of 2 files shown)


36000:
It is inefficient to have the default behavior be to convert it back to rasmm here for the rest of the pipeline (I think most of our users are almost immediately going to put it back to VOX, like the API and CLI do). I think there should be a `preserve_space` flag that defaults to False. Brainlife could set it to True. Or, the flag could be set to True by default, and then our API/CLI set it to False.

Either way, this should be a flag.

AFQ/segmentation.py
```diff
@@ -215,6 +224,15 @@ def segment(self, bundle_dict, tg, fdata=None, fbval=None,
         img_affine : array, optional.
             The spatial transformation from the measurement to the scanner
             space.
+        reset_tg_space : bool, optional
```
36000, line 227: it is called reset_space elsewhere

36000:
Otherwise, LGTM!

36000:
I think this is ready to merge

PR #522, opened 2020-10-07
FIX: Remove xvfb from being needed in tests

36000:
@arokem I think this may help with the memory issues the nightly tests can run into.

- [x] add fury test


arokem:
Do we want to add a workflow that tests with fury as the backend and uses XVFB? Or is that already here somewhere?

36000:
I think we should, that does not exist.

arokem:
Do you want to add that here, or on a separate PR?

36000:
I can do that here. It will not take long to make this test 👍

Issue #519, opened 2020-10-07
Make bundle  input more flexible


arokem:
https://github.com/yeatmanlab/pyAFQ/blob/b60a25b860249b3dc38f11eba8f95012ee6d9bc0/AFQ/api.py#L180

Maybe we can change this to `bundles` instead of `bundle_names`. Then, users can provide a `bundle_dict` input (a dictionary) or a list (in which case the current behavior is retained).

36000:
I think that ROI and bundle_dict should be classes. Then the user wouldn't have to manage complicated dictionaries, they could look at the docstrings for these classes to see how to instantiate them.

arokem:
Good point. A BundleSpec object for each bundle perhaps and then a collection of these specs as inputs here

arokem:
Will also help us deal with some of the brittleness we currently have wrt order of ROIs endpoints and so forth

36000:
Other idea in #338 : use a path

36000:
It really comes down to what is least intrusive to existing workflows and what is least complicated for the end user to provide as input. We should see what Sandra thinks after she uses our bundle_dict interface.

36000:
This was solved with bundle_info at some point.

Issue #487, opened 2020-10-01
BabyAFQ


arokem:
@arokem I believe that you mentioned in passing that this example might be relevant to baby-AFQ wondering if you could elaborate and whether we would want to share with Mareike

_Originally posted by @bloomdt-uw in https://github.com/yeatmanlab/pyAFQ/pull/476#r495108048_

arokem:
Indeed! I think that this shows the way... 

I've already placed all the babyAFQ stuff in a figshare repository available here: https://figshare.com/articles/dataset/ROIs_probabilistic_maps_and_transform_data_for_pediatric_automated_fiber_quantification/13027487

The first thing to do would be to download these files into `~/AFQ_data` or somesuch, following the pattern of https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/data.py#L67-L117 

Then, you'll need to replace our MNI T2 template with the babyAFQ template (it's the file called 'UNCNeo-withCerebellum-for-babyAFQ.nii.gz'), and you'll need to replace the waypoint ROIs with the waypoint ROIs, in a similar fashion to the way we do in #476, and we're done.

Tagging in @grotheer, so she's in the loop as we implement this.

grotheer:
Thank you both! Let me know if I can help in any way

Issue #151, opened 2019-09-28
Make s3fs_nifti_read compatible with non-gzipped files


arokem:
This currently doesn't work raising something like: 

```
MNI_T2_img = afd.s3fs_nifti_read('hcp.recobundles/mni_icbm152_t2_tal_nlin_asym_09a.nii')
File "/usr/local/lib/python3.7/site-packages/AFQ/data.py", line 547, in s3fs_nifti_read
rr = zz.read()
File "/usr/local/lib/python3.7/gzip.py", line 276, in read
return self._buffer.read(size)
File "/usr/local/lib/python3.7/gzip.py", line 463, in read
if not self._read_gzip_header():
File "/usr/local/lib/python3.7/gzip.py", line 411, in _read_gzip_header
raise OSError('Not a gzipped file (%r)' % magic)
OSError: Not a gzipped file (b'\\\x01')
```

I tried to make it compatible [here](https://github.com/yeatmanlab/pyAFQ/pull/150/commits/4baced84ba4af61602a6e78121f2a08fa77808e9), but that didn't work (I think -- raising the same error). It's not central to what I am doing right now, so I am just leaving this note here, so we remember to come back to it at some point.

36000:
no longer relevant here

PR #524, opened 2020-10-08
pyBabyAFQ

bloomdt-uw:
#487 

Based on Mareike's recommendation, pausing work on dHCP dataset and generally abandoning approach to use FA. Instead will focus on replicating MATLAB AFQ results with Kalanit longitudinal data.

### [Kalanit longitudinal data](https://s3.console.aws.amazon.com/s3/buckets/dhcp-afq/Kalanit_Longitudinal_BabyData/)

Processed using [MATLAB BabyAFQ](https://github.com/yeatmanlab/AFQ/tree/babyAFQ)

Goal is to reproduce results in: 
- [Gold standard](https://s3.console.aws.amazon.com/s3/buckets/dhcp-afq/Kalanit_Longitudinal_BabyData/dti94trilin/fibers/afq/manual_tracts/)

- [MATLAB AFQ](https://s3.console.aws.amazon.com/s3/buckets/dhcp-afq/Kalanit_Longitudinal_BabyData/dti94trilin/fibers/afq/)

- [ ] use existing tractography data and feed into pyAFQ segmentation

    - [ ] convert dataset to BIDS compliant
    - [ ] may need to use [PyAFQ scripts](https://github.com/36000/pyAFQ-scripts) to convert MATLAB AFQ files
    - [ ] verify can convert `mif` files to `nii`
    - [ ] possible to use two step process and segment using CLI?

- [ ] if successful, then consider tractography using ACT within pyAFQ

### [dHCP data](https://s3.console.aws.amazon.com/s3/buckets/dhcp-afq/dhcp_dmri_pipeline/)

[Single Subject data](https://s3.console.aws.amazon.com/s3/buckets/dhcp-afq/dhcp_dmri_pipeline/sub-CC00060XX03/ses-12501/dwi/)

- [ ] if Kalanit pipeline sucessful, continue with dHCP

Revisit BIDS conversion

- using default dHCP dataset_description.json generates BIDS validation error.

   dHCP uses BIDS 1.0.2 ([dHCP Data Release Notes](https://dhcp-afq.s3.us-west-2.amazonaws.com/dhcp_dmri_pipeline/dHCP_Data_organization2019.pdf))

   - [ ] converter from BIDS 1.0.2 to BIDS 1.4.0

   - [ ] add BIDSVersion to original file

   - [ ] document and continue to use workaround: replace `"PipelineDescription.Name"` with `"PipelineDescription": { "Name"

- Error processing `sub-CC#####XX##_ses-#####_desc-preproc_dwi.json` file when present. Currently just excluded file

   - [ ] investigate error

- when making BIDS compliant pyAFQ required data in AFQ_data `derivatives` folder

   - [ ] add to documentation

- pyAFQ does not read `nii` files must be `nii.gz`, see #151, therefore needed to gzip nii files.

- [ ] Convert script to Jupyter notebook with [CloudKnot](https://github.com/nrdg/cloudknot) (see #533)

- [ ] if pyAFQ tractography doesn't work, consider two-step pipeline: use qsiprep/mrtrix for tractography then pyAFQ segmentation

- [ ] add segmentation end point filtering

- [ ] dependency: custom ROI monkey patch can be removed once #519 merged

### debugging FA tractography

However if decide to revisit dHCP FA:

Parameters explored thus far (all without endpoint filtering):

- default

- b0 1000

- b0 1000 with FA seed threshold 0.1

- b0 1000 with ROIMask

- b0 1000 with BrainMask and ROIMask

- b0 1000 with BrainMask, ROIMask, and  FA stop threshold 0.1

In all cases some bundles were entirely missing and tractography was highly irregular.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -122,7 +122,7 @@ class BundleDict(MutableMapping):
         else:
             expanded_bundle_names = []
             for bundle_name in bundle_info:
-                if self.seg_algo == "afq" or self.seg_algo == "pediatric":
+                if self.seg_algo == "afq":
                     if bundle_name in ["FA", "FP"]\
                             or bundle_name in CALLOSUM_BUNDLES:
                         expanded_bundle_names.append(bundle_name)
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -76,6 +76,9 @@ RECO_UNIQUE = [
     'CCMid', 'CC_ForcepsMajor', 'CC_ForcepsMinor', 'MCP', 'AC', 'PC', 'SCP',
     'V', 'CC', 'F_L_R']
 
+PEDIATRIC_BUNDLES = [
+    "ARC", "ATR", "CGC", "CST", "FA", "FP", "IFO", "ILF", "MdLF", "SLF", "UNC"]
+
 DIPY_GH = "https://github.com/dipy/dipy/blob/master/dipy/"
 
 
@@ -119,7 +122,7 @@ class BundleDict(MutableMapping):
         else:
             expanded_bundle_names = []
             for bundle_name in bundle_info:
-                if self.seg_algo == "afq":
+                if self.seg_algo == "afq" or self.seg_algo == "pediatric":
                     if bundle_name in ["FA", "FP"]\
                             or bundle_name in CALLOSUM_BUNDLES:
                         expanded_bundle_names.append(bundle_name)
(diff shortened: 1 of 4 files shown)


examples/plot_baby_afq.py
```diff
@@ -0,0 +1,496 @@
...
+    pediatric_templates['ARC_roi2_L'] = pediatric_templates['SLFt_roi2_L']
+    pediatric_templates['ARC_roi2_R'] = pediatric_templates['SLFt_roi2_R']
+    pediatric_templates['ARC_roi3_L'] = pediatric_templates['SLFt_roi3_L']
+    pediatric_templates['ARC_roi3_R'] = pediatric_templates['SLFt_roi3_R']
+
+    # For the middle longitudinal fasciculus (MdLF) reuse ILF ROI
+    pediatric_templates['MdLF_roi2_L'] = pediatric_templates['ILF_roi1_L']
+    pediatric_templates['MdLF_roi2_R'] = pediatric_templates['ILF_roi1_R']
+
+    return pediatric_templates
```
arokem, line 215: All of this stuff should move into the `data` module.

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,496 @@
...
+                    'ROIs': [pediatric_templates[name + '_roi1' + hemi],
+                             pediatric_templates[name + '_roi2' + hemi]],
+                    'rules': [True, True],
+                    'cross_midline': False,
+                    'prob_map': prob_map_data[...,
+                                              prob_map_order.index(name + hemi)],
+                    'uid': uid}
+                uid += 1
+
+    return pediatric_bundles
```
arokem, line 375: This probably also moves into `data` or possibly into `baby.data`

arokem:
One possible API here is that we create a new module `AFQ.baby` that includes all the baby-specific stuff. 

The question is what do we do about things that are implemented in `AFQ.api`? Should sub-class `AFQ.api.AFQ` into `AFQ.baby.api.AFQ`? We would then need to replace some of the stuff related to registration.

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,496 @@
...
+                # custom_tractography_bids_filters=None, # default
+                # ===== Registration parameters ====
+                # b0_threshold=50, # default
+                # min_bval=None, # default
+                min_bval=1000,  # override
+                # max_bval=None, # default
+                max_bval=1000,  # override
+                reg_template=pediatric_templates['UNCNeo-withCerebellum-for-babyAFQ'],
+                # reg_subject="power_map", # default
+                reg_subject="b0",  # override
```
arokem, line 428: Both of these (`reg_template` and `reg_subject`) should be skull-stripped using a brain-mask (for pilot data, we have that, for dHCP data it also comes with the anatomical pipeline derivatives). We should over-ride the pyAFQ default brain-mask (which is generated using the `median_otsu` algorithm) with the brain-mask that is provided with the respective datasets.

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,496 @@
...
+                # brain_mask=B0Mask(), # default
+                brain_mask=MaskFile("brainmask",
+                                    {"scope": "dHCP neonatal dMRI pipeline"}),
```
arokem, line 431: Yep - this is it!

Check if this is a binary mask or a continuous probability mask, and then threshold if needed (there is a class for that!).

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,496 @@
...
+                bids_path=dhcp_home,
+                dmriprep='dHCP neonatal dMRI pipeline',
+                # bids_filters={"suffix": "dwi"}, # default
+                # custom_tractography_bids_filters=None, # default
+                # ===== Registration parameters ====
+                # b0_threshold=50, # default
+                # min_bval=None, # default
+                min_bval=1000,  # override
+                # max_bval=None, # default
+                max_bval=1000,  # override
```
arokem, line 425: We might consider adding all the shells here. Proceed with caution.

Eventually, we might have some more denoising coming (through #625), so we could do DKI with multi-shell, and add some more scalars to the profiles (AWF, MK).

arokem:
> if successful, then consider tractography using ACT within pyAFQ

Let's punt on this one for now. We'll use the tractography that is already there for the pilot data and do qsiprep/mrtrix with the dHCP data.

arokem:
>  add segmentation end point filtering

Might not need this for this application. The MATLAB babyAFQ does not use endpoints for filtering.

arokem:
> add segmentation end point filtering

Furthermore, the template does have a version of the AAL atlas, so if need be, we can come back to that (with the caveat that we'd need to do some massaging of the template to add some volumes that we have in the template we use with grownups).

We'll also need the ALS atlas for the definition of pAF (which, in this PR, is currently being skipped). In this case, a particular part of the AAL atlas is extracted to serve as a waypoint ROI - this is already part of the templates.

examples/plot_baby_afq.py
grotheer: Just so I understand what is going on: Which registration algorithm are we using? Are we applying a brain mask in the step above or are we just checking that it is not a probability mask?

bloomdt-uw: `reg_algo=None` by default, which should use Symmetric Diffeomorphic Registration (SyN) Algorithm for registration.

In this step we are configuring pyAFQ and nothing is actually run until the `myafq.export_all()` command. According to the docs the `brain_mask` is applied before registration to a template.

AFQ/api.py
```diff
@@ -945,6 +1070,9 @@ def _get_best_scalar(self):
         return self.scalars[0]
 
     def get_reg_template(self):
+        if isinstance(self.reg_template, nib.Nifti1Image):
```
arokem, line 1073: @36000 : what do you think? Does this change make sense? Or should this go somewhere else?

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,92 @@
+"""
+==============================
+Pediatric Profiles:
+==============================
+
+The following is an example of tractometry for pediatric bundles.
+
+TODO cite @mareike
```
arokem, line 8: https://www.biorxiv.org/content/10.1101/2021.03.29.437583v1

docs/source/conf.py
```diff
@@ -209,6 +209,7 @@
      'examples_dirs': '../../examples',
      # path where to save gallery generated examples
      'gallery_dirs': 'auto_examples',
+     'ignore_pattern': 'plot_baby_afq.py',
```
arokem, line 212: I wonder why this is not in the `exclude_patterns` on line 82 of this file (and what the difference is between these), but maybe that's not a specific issue with this PR

examples/plot_baby_afq.py
```diff
@@ -0,0 +1,92 @@
...
+import AFQ.data as afd
+
+from AFQ.definitions.mask import RoiMask, MaskFile
+
+##########################################################################
+# Get example data:
+# -----------------
+# **Diffusion dataset**
+#
+# .. note::
```
arokem, line 29: We don't actually use any dataset here, so I would say that this is all somewhat theoretical, and link to the repo that contains the code for the dHCP data analysis, to say that's what we actually do.

arokem (commented):
I think that this is really close to being ready to merge! Just a few small comments here.

docs/source/conf.py
bloomdt-uw: I don't recall exactly why I made this change, it was part of the original commit, so it could be OBE

arokem:
I think this is good to go! @36000 : would you mind taking a quick look to confirm?

AFQ/api.py
arokem: @36000 : do you have any thoughts here? I would like to merge this PR, but need some input from you for this particular line.

arokem:
OK - I believe that the change alluded to [here](https://github.com/yeatmanlab/pyAFQ/pull/524#discussion_r688062629) is not harmful, so I am going to go ahead and merge this, so that we can avoid conflicts down the line.

PR #526, opened 2020-10-08
WIP: Render brainsprite htmls for streamline density (and other?) maps

arokem:
Very preliminary WIP for now, just trying to work this out. 

For now, the example assumes that plot_callosal_profile was run before this is run.


arokem:
An example: https://gist.githack.com/arokem/ef0a0a590be65545f0af21161b975f9b/raw/888c550bee59892a4240437252fb70aba6c2a7fb/density_map.html

PR #537, opened 2020-10-09
[WIP] [temp fix] ignore bundles for plot_tract_profiles when anat position unknown

36000:

Issue #494, opened 2020-10-03
Add Callosum ROIs support


bloomdt-uw:
pyAFQ already uses the midsagittal corpus callosum from the [waypoint ROIs from the human corpus callosum templates](https://figshare.com/articles/Templates_for_Automated_Fiber_Quantification_of_corpus_callosum_from_Diffusion_MRI_data/3381523). Should utilize the entire dataset and add support for using both waypoint and endpoint ROIs for the callosal bundles.

bloomdt-uw:
Originally from @bloomdt-uw [observations](https://github.com/yeatmanlab/pyAFQ/pull/476/files#r494807371) during #443

PR #538, opened 2020-10-09
Enh Add Callosum ROIs support

bloomdt-uw:
#494

- [x] Depends on merge with ~~#537~~ #539 

- [x] Remove `DEFAULT_BUNDLES`

- [ ] Add aal atlas segmentation endpoints for callosum ROIs

- Review `MNI_AAL.txt`. Plot resulting callosal streamlines with aal endpoints

- [ ] Explore additional data sets

- [Stanford Data Repo](https://stacks.stanford.edu/file/druid:rt034xr8593/)
(See https://github.com/nrdg/cloudknot/blob/master/examples/02_process_mri_data.ipynb)

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -100,7 +100,20 @@ def make_bundle_dict(bundle_names=BUNDLES,
         If set, templates will be resampled to the affine and shape of this
         image.
     """
+    logger = logging.getLogger('AFQ.api')
     if seg_algo == "afq":
+        if "FP" in bundle_names and "Occipital" in bundle_names:
+            logger.warning((
+                f"FP and Occipital bundles are co-located, and AFQ"
+                f" assigns each streamline to only one bundle."
+                f" Only Occipital will be used."))
+            bundle_names.remove("FP")
+        if "FA" in bundle_names and "Orbital" in bundle_names:
+            logger.warning((
+                f"FA and Orbital bundles are co-located, and AFQ"
+                f" assigns each streamline to only one bundle."
+                f" Only Orbital will be used."))
+            bundle_names.remove("FA")
         templates = afd.read_templates(resample_to=resample_to)
         callosal_templates = afd.read_callosum_templates(
             resample_to=resample_to)
@@ -161,7 +174,6 @@ def make_bundle_dict(bundle_names=BUNDLES,
                             'cross_midline': False,
                             'uid': uid}
                     else:
-                        logger = logging.getLogger('AFQ.api')
                         logger.warning(f"{name} is not in AFQ templates")
 
                     uid += 1
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -114,6 +114,12 @@ def make_bundle_dict(bundle_names=BUNDLES,
                 f" assigns each streamline to only one bundle."
                 f" Only Orbital will be used."))
             bundle_names.remove("FA")
+        if "FA" in bundle_names and "AntFrontal" in bundle_names:
+            logger.warning((
+                f"FA and AntFrontal bundles are co-located, and AFQ"
+                f" assigns each streamline to only one bundle."
+                f" Only AntFrontal will be used."))
+            bundle_names.remove("FA")
         templates = afd.read_templates(resample_to=resample_to)
         callosal_templates = afd.read_callosum_templates(
             resample_to=resample_to)
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -150,16 +150,21 @@ def make_bundle_dict(bundle_names=BUNDLES,
                     uid += 1
             else:
                 for hemi in ['_R', '_L']:
-                    afq_bundles[name + hemi] = {
-                        'ROIs': [templates[name + '_roi1' + hemi],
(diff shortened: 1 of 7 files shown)


AFQ/api.py
```diff
@@ -48,6 +48,12 @@ def do_preprocessing():
 BUNDLES = ["ATR", "CGC", "CST", "IFO", "ILF", "SLF", "ARC", "UNC",
            "FA", "FP"]
 
+CALLOSUM_BUNDLES = ["AntFrontal", "Motor", "Occipital", "Orbital",
+                    "PostParietal", "SupFrontal", "SupParietal",
+                    "Temporal"]
+
+DEFAULT_BUNDLES = BUNDLES + CALLOSUM_BUNDLES
```
arokem, line 55: This is a bit tricky. IIUC, this segmentation is not perfectly compatible with the default segmentation. 

That's because the CC bundles in the "standard"/default segmentation are partially overlapping with some of these custom CC bundles. So, if they are entered into segmentation together, they will "compete" for streamlines. 

So, I think that we'd want users to choose whether they want to segment CC *or* they want the standard segmentation.

bloomdt-uw: 👍 Figured that might be the case.

36000:
@bloomdt-uw #539 should fix the anatomical plot position problems by saving the plots individually.

AFQ/data.py
```diff
@@ -2006,7 +2006,15 @@ def bundles_to_aal(bundles, atlas=None):
         "UNC_L": [['leftanttemporal'], ['leftuncinatefront']],
         "UNC_R": [['rightanttemporal'], ['rightuncinatefront']],
         "ARC_L": [['leftfrontal'], ['leftarctemp']],
-        "ARC_R": [['rightfrontal'], ['rightarctemp']]}
+        "ARC_R": [['rightfrontal'], ['rightarctemp']],
+        "AntFrontal": [None, None],
```
arokem, line 2010: @jyeatman : what are the correct endpoints for each one of these segments of the CC?

36000:
I added issue #614 to remind us to figure out where the callosum endpoints are, but I don't think that is high priority.

36000:
@arokem this is ready for review.

@bloomdt-uw for your checklist, I moved the endpoint atlas to an issue, so we can merge this now and find out what the correct endpoints are later. I am not sure what the last part of your checklist is referring to, however this may merit its own issue.

I got rid of the test and instead have a callosum example. I also modified some of the examples to use less streamlines in tractography to save space, which I think was causing problems. I also modified the 'organize stanford data' function to optionally delete any afq derivatives folder in the stanfords bids directory. This is because the examples use the same BIDS directory and were using each others derivatives.

bloomdt-uw:
The task to explore other datasets was meant to be a quality control check

arokem:
Looks great. My main comment here is that we probably want to prevent users from asking for these new callosal bundles and also for FP/FA in the same AFQ object. The reason is that these are co-located and our current code makes sure that each streamline is classified into only one bundle. This would mean that (for example) the occipital callosum and FP would be competing for the same streamlines, which doesn't make sense.

36000:
@arokem I integrated your advice. This is what the example produces now.
image

AFQ/api.py
```diff
@@ -94,7 +100,20 @@ def make_bundle_dict(bundle_names=BUNDLES, seg_algo="afq", resample_to=False):
...
+                f"FP and Occipital bundles are co-located, and AFQ"
+                f" assigns each streamline to only one bundle."
+                f" Only Occipital will be used."))
+            bundle_names.remove("FP")
+        if "FA" in bundle_names and "Orbital" in bundle_names:
+            logger.warning((
+                f"FA and Orbital bundles are co-located, and AFQ"
+                f" assigns each streamline to only one bundle."
+                f" Only Orbital will be used."))
+            bundle_names.remove("FA")
```
arokem, line 116: I think we also need:

```suggestion
            bundle_names.remove("FA")
            if "FA" in bundle_names and "AntFrontal" in bundle_names:
            logger.warning((
                f"FA and AntFrontal bundles are co-located, and AFQ"
                f" assigns each streamline to only one bundle."
                f" Only AntFrontal will be used."))
            bundle_names.remove("FA")
            
```

arokem (commented):
Looks great! One more small suggestion to follow up on the recent change.

36000:
I added #616  for exploring different datasets.

examples/plot_afq_callosal.py
```diff
@@ -0,0 +1,60 @@
+"""
+==========================
+Calossal bundles using AFQ API
```
arokem, line 3:
```suggestion
Callosal bundles using AFQ API
```

examples/plot_afq_callosal.py
```diff
@@ -0,0 +1,60 @@
...
+==========================
+An example using the AFQ API to find calossal bundles using the templates from:
```
arokem, line 5:
```suggestion
An example using the AFQ API to find callosal bundles using the templates from:
```

PR #539, opened 2020-10-12
Save non anat positioned bundles in their own plots

36000:
- [x] Test this on reco80
- [x] make it work with api
- [x] Test on api

- [x] supporting plots blank

- [x] optionally only plot above some threshold


36000:
@arokem this is ready for review / merge. it is also important in #538 .

arokem:
Looks good!

PR #543, opened 2020-10-13
Adding dummy end points for custom bundles

### LINKED ISSUE #498: If define custom ROI and do not set `filter_by_endpoints=False` then segmentation crashes
If define custom ROI and do not set `filter_by_endpoints=False` then segmentation crashes. Could fail gracefully or ideally should warn that segmentation can't find and proceed with processing without endpoints (suggestion from @jyeatman).

_Originally posted by @bloomdt-uw in https://github.com/yeatmanlab/pyAFQ/pull/476#r494810189_

bloomdt-uw:
Fixes #498


arokem:
This is great. I just resolved a conflict here, but will merge as soon as the CI comes back green.

PR #545, opened 2020-10-14
Read and resample ROI data.

arokem:
This is a utility function for handling user-provided ROIs that are registered to MNI space, but are stored with an affine that differs from MNI.

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -151,7 +151,7 @@ def read_callosum_templates(resample_to=False):
     return template_dict
 
 
-def read_roi(roi, resample_to=None):
+def read_resample_roi(roi, resample_to=None):
     """
     Reads an roi from file-name/img and resamples it to conform with
     another file-name/img.
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -151,7 +151,7 @@ def read_callosum_templates(resample_to=False):
     return template_dict
 
 
-def read_resample_roi(roi, resample_to=None):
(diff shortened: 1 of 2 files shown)


arokem:
TODO: 

- [ ] Test

jyeatman:
1) I would suggest having the default be 'MNI' rather than NONE. That way it is clear reading the first line that the data is resampled to the MNI template. NONE should probably do nothing

2) Should there be a check to see if the affine matches the MNI affine? If I understand correctly, this is only necessary if they have different affines

jyeatman:
I would suggest renaming to read_template_roi since (I think) this would just be intended for template rois not for native space rois
Or maybe read_roi_resample

36000:
> 1. I would suggest having the default be 'MNI' rather than NONE. That way it is clear reading the first line that the data is resampled to the MNI template. NONE should probably do nothing
> 2. Should there be a check to see if the affine matches the MNI affine? If I understand correctly, this is only necessary if they have different affines

To address #1, I think it is common in python to have default arguments be `None`, which are then replaced by the actual default. See: https://unhashable.com/always-use-none-for-default-args-in-python/ . It's also something we commonly do in pyAFQ already. The docstrings explain what None does. However, if its important to give the user the option to not resample, we could make it so passing `False` means no resampling.

36000:
My vote is for read_resample_roi , but read_roi is also good

36000:
LGTM

Issue #550, opened 2020-10-15
Running with `force_recompute` causes tracking and segmentation to run multiple times


bloomdt-uw:
Likely due to lazy loading mechanism in `API` and interplay with `force_recompute`.

@arokem is suggesting removing this feature, and instead requiring users to delete appropriate derivatives files to replicate force pyAFQ to recompute derivatives.

PR #552, opened 2020-10-15
Eliminate `force_recompute` option.

arokem:
This is not a good way to do this. In particular, it triggers
multiple rounds of recomputing, for what should have been
obvious reasons. For now, if a user wants to force recomputing
the best approach would be to delete files.


arokem:
Addresses #550.

arokem:
@36000 : do you have any thoughts about this?

36000:
LGTM. merging

PR #555, opened 2020-10-17
Adds prealign stage to the examples.

### LINKED ISSUE #541: Add prealign to examples
This will make the examples more correct, in case some users are relying on the examples as a basis to start their own pipelines, and we want to make sure that it's closer to the correct way of doing things.

arokem:
Closes #541


36000:
LGTM. merging

PR #566, opened 2020-10-21
[FIX] Put tg in rasmm first for SLR registration

36000:
I think that SLR registration in pyAFQ with waypoint ROI segmentation has always had this problem. It works significantly better with this fixed.

Note: SLR with recobundles is handled elsewhere and did not have this bug, only SLR with waypoint ROI approach has this bug.


36000:

PR #533, opened 2020-10-09
[ENH] Add cloudknot example

36000:
- [x] simple cloudknot example
- [x] complicate cloudknot example compare tractography on hcp example
- [x] use ignore_pattern to ignore this example

Closes #440

PR #567, opened 2020-10-21
[ENH] added reco80 example

36000:
- [x] don't run this example in the docs test suite (see: #533 )

### DIFF
--- a/examples/plot_afq_reco80.py
+++ b/examples/plot_afq_reco80.py
@@ -1,0 +1,178 @@
+"""
+==========================
+RecoBundles80 using AFQ API
+==========================
+
+An example using the AFQ API to run recobundles with the 80 bundle atlas.
+
+
+"""
+import os.path as op
+
+import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


36000:

arokem:
I think this looks fine. Does this take very long to run? Is that why we don't want to run it in the documentation build? I worry about having too many examples that are not exercised by the CI, because they might go out of date, without us noticing. One additional comment is that we can shorten this example a little bit: maybe only run the bundle visualization, and skip the tract profiles and FA visualization. Those are shown in other examples already.

36000:
@arokem I made those changes. I think it is ready to merge now.
image

arokem:
@36000 : I've shortened and simplified the text a bit, focusing on the main differences with other examples (see https://github.com/yeatmanlab/pyAFQ/pull/567/commits/74f3f004daf9f328b2b1e4feced9f3eed388f78c). If that looks OK to you, I think this is ready to merge.

PR #568, opened 2020-10-21
[ENH] add weighted option for dice

36000:
Adds the option to calculate a wDSC (as in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6339903/). Defaults to True.

### DIFF
--- a/AFQ/utils/volume.py
+++ b/AFQ/utils/volume.py
@@ -103,11 +103,22 @@ def dice_coeff(arr1, arr2, weighted=True):
         converted to an ndarray.
     weighted : bool, optional
         Whether or not to weight the DICE coefficient as in [Cousineau2017]_.
+        The weighted Dice coefficient is calculated by adding the sum of all
+        values in arr1 where arr2 is nonzero to the sum of all values in arr2
+        where arr1 is nonzero, then dividing that by the sum of all values in
+        arr1 and arr2.
         Default: True
 
     Returns
     -------
     The dice similarity between the images.
+
+    Notes
+    -----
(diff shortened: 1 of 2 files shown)


AFQ/utils/volume.py
```diff
@@ -101,6 +101,9 @@ def dice_coeff(arr1, arr2):
     arr2 : Nifti1Image, str, ndarray
         The other ndarray to compare. Can be a path or image, which will be
         converted to an ndarray.
+    weighted : bool, optional
+        Whether or not to weight the DICE coefficient as in [Cousineau2017]_.
```
arokem, line 105: We'll need to add the full reference in a "Notes" section in this docstring.

AFQ/utils/volume.py
```diff
@@ -116,7 +119,16 @@ def dice_coeff(arr1, arr2):
...
-    return 1 - dice(
-        arr1.flatten().astype(bool),
-        arr2.flatten().astype(bool))
+    arr1 = arr1.flatten()
+    arr2 = arr2.flatten()
+
+    if weighted:
+        return (
+            np.sum(arr1 * arr2.astype(bool))
+            + np.sum(arr2 * arr1.astype(bool)))\
```
arokem, line 128: I don't think you need this slash because the statement is in parentheses

```suggestion
            + np.sum(arr2 * arr1.astype(bool)))
```

arokem (commented):
Good addition. Maybe we can add an explanation in the docstring about what "weighted" means here.

AFQ/utils/volume.py
36000: The backslash here is necessary because those two statements are in different parentheses

PR #573, opened 2020-10-26
Added an ROI pre segmentation as an option for recobundles

jyeatman:
@arokem @36000 Here is my proposal for how to add an option for an ROI pre-segmentation. If you think this will work, then I will draw all the ROIs and make the bundle_dict for all the recobundles fiber groups

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -862,7 +862,6 @@ class Segmentation:
                 # Now rb should be initialized based on the fiber group coming
                 # out of the roi segmentation
                 self.move_streamlines(indiv_tg, self.reg_algo)
-
                 rb = RecoBundles(
                     self.moved_sl,
                     verbose=False,
@@ -890,6 +889,7 @@ class Segmentation:
                     op.join(self.save_intermediates,
                             f"{bundle}_model.trk"),
                     bbox_valid_check=False)
+
             # Either whole brain tracgtogram or roi presegmented fiber group
             # goes to rb.recognize
             _, rec_labels = rb.recognize(model_bundle=model_sl,
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -871,7 +871,6 @@ class Segmentation:
                 # Now rb should be initialized based on the fiber group coming
                 # out of the roi segmentation
                 self.move_streamlines(indiv_tg, self.reg_algo)
-
                 rb = RecoBundles(
                     self.moved_sl,
                     verbose=False,
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -855,9 +855,9 @@ class Segmentation:
             if self.presegment_bundle_dict is not None:
                 if "return_idx" in self.presegment_kawrgs\
                         and self.presegment_kawrgs["return_idx"]:
-                    indiv_tg = roiseg_fg[bundle]['sl'].streamlines
+                    indiv_tg = roiseg_fg[bundle]['sl']
                 else:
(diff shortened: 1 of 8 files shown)


36000:
@arokem does this look good to you?

arokem:
Yes. I think so. It's a bit of a mind-bender that one of the object's methods includes the object itself, but there's nothing wrong with that :-) Have you had a chance to try running this?

36000:
I am running it now, and had to make a change to the api, so the user can specify which bundle dict to use for the pre segmentation.

36000:
I just ran this on HCP data and got no streamlines back ... will require some debugging!

36000:
World's first ever bundle segmented with both waypoint ROIs and recobundles:
https://gist.githack.com/36000/83b9ace9574cb4614966d35535fa8908/raw/615501de6dce434c31489546bd92fca801dd6ece/sub-1103_ses-pre_acq-b1500_aligned_trilin_dwi_space-RASMM_model-DTI_desc-det-Reco_viz.html

36000:

AFQ/segmentation.py
```diff
@@ -699,7 +717,7 @@ def segment_afq(self, tg=None):
                 # We need to check this again:
                 if len(new_select_sl) == 0:
                     self.logger.info("After filtering "
-                                     f"{len(select_sl)} streamlines")
+                                     f"{len(new_select_sl)} streamlines")
```
arokem, line 720: Good catch!

AFQ/segmentation.py
```diff
@@ -762,20 +779,20 @@ def move_streamlines(self, tg=None, reg_algo='slr'):
             self.logger.info("Registering tractogram with SLR")
             atlas = self.bundle_dict['whole_brain']
             self.moved_sl, _, _, _ = whole_brain_slr(
-                atlas, self.tg.streamlines, x0='affine', verbose=False,
+                atlas, tg.streamlines, x0='affine', verbose=False,
```
arokem, line 782: Maybe we should change the signature of this function to the following?

```
def move_streamlines(self, tg, reg_algo='slr')
```
It looks like you made it so that the `tg` argument is never going to be optional.

AFQ/segmentation.py
```diff
@@ -805,17 +822,90 @@ def segment_reco(self, tg=None):
...
 
         self.logger.info("Assigning Streamlines to Bundles")
         for bundle in bundle_list:
+            self.logger.info(f"Finding streamlines for {bundle}")
             model_sl = self.bundle_dict[bundle]['sl']
+
+            # If doing a presegmentation based on ROIs then initialize rb after
+            # Filtering the whole brain tractogram to pass through ROIs
+            if self.presegment_bundle_dict is not None:
+                afq_bundle_name = afd.BUNDLE_RECO_2_AFQ.get(bundle, bundle)
```
arokem, line 857: IIUC, this makes this rather dependent on the current relationships between waypoint and recobundle bundle names, right? What would happen if we had a `bundle_dict` with waypoint ROIs with names of RB bundles not currently defined as waypoint ROIs?

arokem (commented):
Overall looks great! We should do some more experiments with this. 

@jyeatman : any ideas on automatic generation of waypoint ROIs from the RB atlas bundles?

AFQ/segmentation.py
36000: The second argument in afd.BUNDLE_RECO_2_AFQ.get(bundle, **bundle**) means that if bundle is not in BUNDLE_RECO_2_AFQ, then **bundle** is returned. So this would still work!

PR #577, opened 2020-10-27
DOC: Example that explores BIDS and includes tractography from another pipeline.

arokem:
@jyeatman : this is the example that shows how to use tractography outputs from another pipeline

### DIFF
--- a/examples/plot_other_tracking.py
+++ b/examples/plot_other_tracking.py
@@ -101,7 +101,7 @@ afd.to_bids_description(
 bundle_info = ["SLF", "ARC", "CST", "FP"]
 
 ##########################################################################
-# Now, we can run AFQ, pointing to the derivatives of the
+# Now, we can define our AFQ object, pointing to the derivatives of the
 # "my_tractography" pipeline as inputs. This is done by setting the
 # `custom_tractography_bids_filters` key-word argument. We pass the
 # `bundle_info` defined above. We also point to the preprocessed
@@ -125,5 +125,10 @@ my_afq = api.AFQ(
         "scope": "my_tractography"
     },
     segmentation_params={'nb_streamlines': 10000})
+
+##########################################################################
+# Finally, to run the segmentation and extract tract profiles, we call
+# The `export_all` method. This creates all of the derivative outputs of
+# AFQ within the 'stanford_hardi/derivatives/afq' folder.
 
 my_afq.export_all()
--- a/examples/plot_other_tracking.py
+++ b/examples/plot_other_tracking.py
@@ -1,0 +1,108 @@
+"""
+=============================================
+Segmenting tractography from another pipeline
+=============================================
+
+The AFQ API provides facilities to segment tractography results obtained using
+other software. For example, we often use
+`qsiprep <https://qsiprep.readthedocs.io/en/latest/>`_ to preprocess our data
+and reconstruct tractographies with software such as
+`MRTRIX <https://www.mrtrix.org/>`_. Here, we will demonstrate how to use
+these reconstructions in the pyAFQ segmentation and tractometry pipeline.
+
+"""
+import os
+import os.path as op
+
+import matplotlib.pyplot as plt
+import nibabel as nib
+import plotly
+
+from AFQ import api
+import AFQ.data as afd
+
+##########################################################################
+# Example data
+# ---------------------
+# The example data we will use here is generated from the Stanford HARDI
+# dataset (https://purl.stanford.edu/ng782rw8378). The calls below organize the # preprocessed data according fetches
+# the results of tractography with this dataset and organizes it within
+# the `~/AFQ_data` folder.
+
+afd.organize_stanford_data()
+afd.fetch_stanford_hardi_tractography()
+
+##########################################################################
+# Reorganize data
+# ---------------------
+# We organize the data so that it conforms with the BIDS standard for
+# derivatives:
+
+bids_path = op.join(op.expanduser('~'), 'AFQ_data', 'stanford_hardi')
+tractography_path = op.join(bids_path, 'derivatives', 'my_tractography')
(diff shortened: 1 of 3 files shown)


arokem:
I think there's something wonky with this trk file. I get: 

```
In [5]: img = nib.load("/Users/arokem/AFQ_data/stanford_hardi/derivatives/vistas
   ...: oft/sub-01/ses-01/dwi/sub-01_ses-01_dwi.nii.gz")

In [6]: load_tractogram("/Users/arokem/AFQ_data/stanford_hardi/derivatives/my_tr
   ...: actography/sub-01/ses-01/dwi/sub-01_ses-01-dwi_tractography.trk", img)
WARNING:root:Dimensions must be above 0.
WARNING:root:Dimensions must be above 0.
WARNING:root:Dimensions must be above 0.
ERROR:root:Affine not equal
ERROR:root:Dimensions not equal
ERROR:root:Voxel_size not equal
ERROR:root:Trk file header does not match the provided reference.
Out[6]: False
```

examples/plot_other_tracking.py
```diff
@@ -0,0 +1,108 @@
...
+# |                     └── sub-01_ses-01_dwi.nii.gz
+
+
+##########################################################################
+# Now, we can run AFQ, pointing to the derivatives of the
+# "my_tractography" pipeline as inputs:
+
+bundle_names = ["SLF", "ARC", "CST", "FP"]
+
+
```
jyeatman, line 98: add reference to the budle_info options and how they can be set for different purposes

examples/plot_other_tracking.py
```diff
@@ -0,0 +1,108 @@
...
+# |                     ├── sub-01_ses-01_dwi.bvecs
+# |                     └── sub-01_ses-01_dwi.nii.gz
+
+
+##########################################################################
+# Now, we can run AFQ, pointing to the derivatives of the
+# "my_tractography" pipeline as inputs:
+
+bundle_names = ["SLF", "ARC", "CST", "FP"]
+
```
jyeatman, line 97: Add reference to different dmriprep options and how they should be set for different preprocessing pipelines. How would this be set for qsi-prep recon? Can we give a specific example for that since it is the most common

examples/plot_other_tracking.py
```diff
@@ -0,0 +1,108 @@
...
+
+my_afq = api.AFQ(
+    bids_path,
+    dmriprep='vistasoft',
+    bundle_info=bundle_names,
+    custom_tractography_bids_filters={
+        "suffix": "tractography",
+        "scope": "my_tractography"
```
jyeatman, line 105: Can you add an example of how you would set the nb_streamlines to subsample the 10M from qsi-prep

arokem:
Thanks for the comments. I think they're addressed in the last couple of commits, but let me know if you think I should be more specific about anything.

arokem:
On a bit of further thought, I think that this example could be expanded a bit to explain how/why we use BIDS more generally. Hopefully, I'll find some time to work on that tomorrow. And also to fix the failing test.

examples/plot_other_tracking.py
```diff
@@ -0,0 +1,129 @@
...
+# |     │               └── sub-01_ses-01-dwi_tractography.trk
+# |     └── vistasoft
+# |         ├── dataset_description.json
+# |         └── sub-01
+# |             └── ses-01
+# |                 └── dwi
+# |                     ├── sub-01_ses-01_dwi.bvals
+# |                     ├── sub-01_ses-01_dwi.bvecs
+# |                     └── sub-01_ses-01_dwi.nii.gz
+
```
bloomdt-uw, line 90: May want to identify any files that are under bids that cannot be located, do not conform to standards, and do not follow convention. If pybids doesn't support this we may want to open an issue.

In lieu can:
- run `bids-validator`
- run `bids_layout.get()` and compare to an `os.walk`
- run `get_entities()`

examples/plot_other_tracking.py
bloomdt-uw: We may want to ensure that the tck file is also valid -- perhaps run mrtrix `tckinfo`

examples/plot_other_tracking.py
bloomdt-uw: Motivation: Users may encounter a `MemoryError` depending on available system memory. Recommendation is to either specify `nb_streamlines`  or adjust system memory. This process may require some iteration to find suitable configuration. We recommend running with minimum of 32Gb? and maximum streamline count of 1M

examples/plot_other_tracking.py
```diff
@@ -0,0 +1,129 @@
...
+# a bit, we also sub-sample the provided tractography. This is
+# done by defining the segmentation_params dictionary input.
+# To sub-sample to 10,000 streamlines, we define
+# `'nb_streamlines' = 10000`.
+
+my_afq = api.AFQ(
+    bids_path,
+    dmriprep='vistasoft',
+    bundle_info=bundle_info,
+    custom_tractography_bids_filters={
```
bloomdt-uw, line 123: May want to include primer on how to configure/diagnose bids_filters:

```
bids_layout = BIDSLayout(BIDS_DIR, derivatives=True)
# find all files
bids_layout.files
# get a specific file; probably want to specify fill path
bids_layout.get()
# find out what entities are available for bids filters for file
bids_layout.get().get_entities()
```

Some issues have been `dataset_description.json`, file naming convention doesn't uniquely identify file with bids filters, ...

arokem:
Reading this and the comments again, I think that maybe this is enough for now? I think that a more detailed tutorial on BIDS/pyBIDS is out of scope for us, and users can get more information from the pybids documentation. So, I think this is ready for another review and potentially even merge.

36000:
LGTM!

PR #580, opened 2020-10-28
[ENH] Robust tensor fitting

36000:


AFQ/models/dti.py
```diff
@@ -6,19 +6,86 @@
...
+            gtab,
+            fit_method="RT",
+            sigma=sigma)
     return dtimodel.fit(data, mask=mask)
 
 
 def fit_dti(data_files, bval_files, bvec_files, mask=None,
-            out_dir=None, file_prefix=None, b0_threshold=50):
+            out_dir=None, file_prefix=None, b0_threshold=50,
+            rtf=False):
```
arokem, line 88: Looks like this key-word argument is not being used

arokem: Also, can we use something more explicit? For example `restore=False` instead of `rtf`?

arokem (commented):
Overall looks good. Have you had a chance to run this with some prek data?

AFQ/models/dti.py
36000: To include restore in this function, we need to write another function that allows us to load multiple bval and bvec files then merge them together, I think. Which I have not done yet - the AFQ object does not use this function, but I think the examples do.

36000:
I ran this on 8 subjects and the results are inconclusive. I am now running it on all subjects (prek doesn't take as long as HCP) to see if we see the same problems.

AFQ/models/dti.py
arokem: I think `ut.prepare_data` does what you want? It's the first thing this function calls. 

In general, as you noted, this function is not directly called by the API. 

When we began the project, @jyeatman and folks in his team wanted to be able to fit DIPY models from the command line, which is why we still have a bunch of CLIs that do that. I wonder whether we should remove functionality directly related just to that. I believe that DIPY now has CLIs that do all these things. @jyeatman : WDYT?

36000:
@arokem for now maybe we can ignore fit_dti. Other than that, this should be ready to merge. I just rebased it on master.

arokem:
Looks good!

PR #587, opened 2020-11-03
FIX: Traverse BIDS hierarchy to find masks, bvals, and bvecs

richford:
This PR traverses the BIDS hierarchy to find masks, bval files, and bvec files instead of assuming that they will be found in the same session. It is part bugfix, part enhancement because it also fixes the underscore error in `PFTMask.get_mask()`.

Traversing the hierarchy should allow it to interact with common output patterns from nipreps.

### DIFF
--- a/AFQ/tractography.py
+++ b/AFQ/tractography.py
@@ -1,3 +1,4 @@
+from collections.abc import Iterable
 import numpy as np
 import nibabel as nib
 import dipy.reconst.shm as shm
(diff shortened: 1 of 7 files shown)


richford:
@arokem and @36000, assuming the documentation builds pass, I think this is ready for review. In the latest commit, I changed all internal `.bvals` and `.bvecs` extensions to `.bval` and `.bvec`. As far as I know, the plural versions are not BIDS-compliant and they were making pybids mad.

AFQ/tractography.py
```diff
@@ -166,11 +167,11 @@ def track(params_file, directions="det", max_angle=30., sphere=None,
                 "You are using PFT tracking, but did not provide a string ",
                 "'stop_threshold' input. ",
                 "Possible inputs are: 'CMC' or 'ACT'")
-        if not isinstance(stop_mask, tuple):
+        if not (isinstance(stop_mask, Iterable) and len(stop_mask) == 3):
```
36000, line 170: this is perfect!

36000:
I changed some of the nighlty smoke tests to run tractography when testing tractography related things. If this runs into emmory errors I can reduce number of seeds, but that should be in a different PR. These new tests would have been able to trigger your error, I think.

AFQ/tractography.py
richford: Thanks!

Issue #368, opened 2020-08-17
Add more scalars


arokem:
DKI MK

Also AD / RD (see also #327), others? 

Might be interesting to explore whether a model can automatically tell us what all the interesting scalars are. Or do some other kind of auto-generation of the valid scalars dict in the API.

36000:
Conversation moved to #505

PR #588, opened 2020-11-04
Removes several unused CLIs.

### LINKED ISSUE #586: Deprecate CLIs
For historical reasons, we have several CLIs for fitting and predicting different models. I believe that we are not really using that and there are alternatives in DIPY, so I think that we should deprecate these. WDYT @jyeatman? Is anyone in your group using these CLIs?

arokem:
Resolves #586.


arokem:
For the time being, I am leaving the models module untouched, so we can continue to build on it. For example to address #368

36000:
@arokem I removed tests for the CLIs, and now it passes. I think this is ready to merge?

arokem:
I think you're right!

Issue #590, opened 2020-11-06
When we compute APM, we need to reuse CSD coefficients


arokem:
Right now, when we compute APM, we run a new fit of CSD just for that ([here](https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/models/csd.py#L166)). We should reuse previously fit CSD, if that's available. We should be able to ingest and convert SH coefficients from other pipelines (e.g., qsirecon) to compute APM.

PR #591, opened 2020-11-07
Reuse the CSD sh coefficients if you already have them.

arokem:
Or create them for the first time to be reused later on.

This addresses some of #590. Still need to figure out how to convert SH coefficients from mrtrix into our APM.


arokem:
Any thoughts about this @36000 ? This does not yet address the issue of bringing in sh coefficients from other software pipelines (e.g., qsirecon/mrtrix), but would save one round of CSD fitting that we are currently doing from happening. I think that in most cases, the order would be that APM would be computed first (for registration) and then these coefficients would get reused for tracking later on.

36000:
You can now remove our fit_anisotropic_power_map function, then LGTM

arokem:
OK - this is now rebased! @36000 : when you get a chance, could you please review/merge?

PR #596, opened 2020-11-12
[ENH] Avoid hard crash when bundle not in AAL atlas

36000:
If the user adds a new bundle and runs pyAFQ naively without changing the AAL atlas, pyAFQ will currently hard crash. I think the better solution is to just not endpoint filter that bundle. 

Custom endpoint filtering is being handled already in #589


36000:
@bloomdt-uw I just checked, and on master this is necessary. Maybe you fixed this elsewhere?

bloomdt-uw:
Believe I was thinking of #543

36000:
You are right. I will merge that and close this.

PR #603, opened 2020-11-18
RF: Get the information about number of streamlines from the header.

### LINKED ISSUE #602: pyAFQ hangs on this line with large segmentation files
https://github.com/yeatmanlab/pyAFQ/blob/87a85f71affb242c10cec73e4042ddaab17c1164/AFQ/api.py#L1397

arokem:
Potentially closes #602.


36000:
This is good, but I think the problem comes from loading the file at all.

36000:
Deprecated by #605

PR #610, opened 2020-12-03
BF: use get instead of get_nearest

36000:
get_nearest cannot find the customb0 file in the test that uses it, so I switched to using get instead. 
This fixes this test: https://github.com/yeatmanlab/pyAFQ/blob/71096f443fa76e98c3ecb24f6def125db67eedf6/AFQ/tests/test_api.py#L446
Which should fix one of the nightly builds.


36000:
@arokem this is ready for review / merge.

arokem:
LGTM. Thanks!

PR #613, opened 2020-12-03
[WIP] [ENH] allow user to specify callosal bundles in list of bundle names for bundle_info, add example

36000:
This allows the user to also give callosal names when using bundle_info as a list. It also adds an example where we use the api to run pyAFQ with both sets of bundles. 

- [ ] find source of calossal bundles


36000:
This is a duplicate of #538 , I have moved this example to that PR.

PR #615, opened 2020-12-07
BF: Need to read these parameters from file, before getting the data.

arokem:


arokem:
While I am here, I also added a calculation of mean kurtosis, another kurtosis-derived scalar.

AFQ/api.py
```diff
@@ -907,12 +907,25 @@ def _dki_awf(self, row, sphere='repulsion100', gtol=1e-2):
             afd.write_json(meta_fname, meta)
         return dki_awf_file
 
+    def _dki_mk(self, row):
+        dki_mk_file = self._get_fname(row, '_model-DKI_MK.nii.gz')
+        if not op.exists(dki_mk_file):
+            tf = self._dki_fit(row)
+            mk = tf.mk()
```
arokem, line 914: Note that `mk` is indeed a standard method of the object, and not an attribute (in contrast to `fa` and `md`)

36000:
LGTM

PR #618, opened 2020-12-14
Dilate the ROIs.

arokem:
This avoids strange edge cases, where the one voxel thick ROIs get squished by the SyN mapping and nothing remains of them, raising an error from `patch_up_roi`.

### DIFF
--- a/examples/plot_tract_profile.py
+++ b/examples/plot_tract_profile.py
@@ -26,7 +26,7 @@ import AFQ.tractography as aft
 import AFQ.registration as reg
 import AFQ.models.dti as dti
 import AFQ.segmentation as seg
-from AFQ.utils.volume import patch_up_roi
+from AFQ.utils.volume import transform_inverse_roi
 
 import logging
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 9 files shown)


examples/plot_tract_profile.py
```diff
@@ -124,10 +124,9 @@
     for bundle in bundles:
         for idx, roi in enumerate(bundles[bundle]['ROIs']):
             if bundles[bundle]['rules'][idx]:
-                warped_roi = patch_up_roi(
-                    mapping.transform_inverse(
-                        roi.get_fdata().astype(np.float32),
-                        interpolation='linear'),
+                warped_roi = auv.transform_inverse_roi(
```
arokem, line 127: You currently have: 

```
from AFQ.utils.volume import transform_inverse_roi

...

warped_roi = auv.transform_inverse_roi(...)

```

I think you want either: 


```
import AFQ.utils.volume as auv
...

warped_roi = auv.transform_inverse_roi(...)

```

or: 

```
from AFQ.utils.volume import transform_inverse_roi

...

warped_roi = transform_inverse_roi(...)

```

36000: Good catch! The convention in the examples seems to be to do the latter, while the former is used elsewhere in the code.

PR #622, opened 2020-12-28
Add more timing information

36000:
Specifically, pre-registration takes ~5x longer than mapping on pre-K data, so it needs to be in the timing file.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -520,7 +520,7 @@ class AFQ(object):
         # Initialize dict to store relevant timing information
         timing_dict = {
             "Tractography": 0,
-            "Registration, pre-align": 0,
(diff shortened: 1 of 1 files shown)


36000:
Change commas to underscores

AFQ/api.py
```diff
@@ -1042,6 +1050,8 @@ def _reg_prealign(self, row):
                 row, '_prealign_from-DWI_to-MNI_xfm.json')
             meta = dict(type="rigid")
             afd.write_json(meta_fname, meta)
+            row['timing']['Registration, pre-align'] =\
```
arokem, line 1053: This comma will wreak havoc as a field in a csv file 😄

Issue #624, opened 2021-01-15
bundle_dict should only be made when needed


36000:
This would avoid repeatedly downloading templates when using pyAFQ in docker.

36000:
closed by #625

PR #625, opened 2021-01-19
Major changes for processing optic radiations with CLI

36000:
- [x] export seed and stop masks for tractography
- [x] bundle dict not calculated until needed
- [x] prealign parameters can be changed from CLI
- [x] allow ROIs to have individual tolerances
- [x] minor bug fixes
- [x] parallelize segmentation
- [x] new mapping API to make custom maps easier
- [x] viz endpoint ROIs as well
- [x] new mapping API docs
- [x] new scalar API for custom scalars, docs


36000:
Closes #624

36000:

AFQ/api.py
```diff
@@ -268,6 +269,10 @@ def __init__(self,
         b0_threshold : int, optional
             [REGISTRATION] The value of b under which
             it is considered to be b0. Default: 50.
+        patch2self : bool, optional
+            [REGISTRATION] Whether to use patch2self when
```
arokem, line 273:
```suggestion
            [REGISTRATION] Whether to use patch2self
```

AFQ/data.py
```diff
@@ -2228,7 +2223,7 @@ def write_json(fname, data):
     None
     """
     with open(fname, 'w') as ff:
-        json.dump(data, ff)
+        json.dump(data, ff, default=lambda obj: "Not Serializable")
```
arokem, line 2226: What's the purpose of this? Wouldn't this cause some silent failures where no data would get written, but we'd just move on assuming it was written?

AFQ/definitions/scalar.py
```diff
@@ -0,0 +1,107 @@
...
+    suffix : str
+        suffix to pass to bids_layout.get() to identify the file.
+    filters : str
+        Additional filters to pass to bids_layout.get() to identify
+        the file.
+        Default: {}
+
+    Examples
+    --------
+    seed_mask = MaskFile(
```
arokem, line 51: This example. doesn't look right. Probably here because of a copy-paste?

AFQ/definitions/utils.py
```diff
@@ -0,0 +1,109 @@
...
+    Definitions have an init function which the users uses to specify
+    how they want the definition to behave.
+    The find_path and get_for_row functions are called by the AFQ API.
+    The api calls find_path to let the definition find relevant files
+    for the given subject and session. The api calls get_for_row to get the
+    mask, map, etc.
+    '''
+
+    def __init__(self):
+        raise NotImplementedError("Use a definition that inherits this class")
```
arokem, line 18: I think this error would be confusing, because it's raised by the inheriting class, not by this class.

AFQ/definitions/utils.py
```diff
@@ -0,0 +1,109 @@
+__all__ = ["Definition", "find_file"]
+
+
+class Definition():
```
arokem, line 4:
```suggestion
class Definition(object):
```

Does this even matter?

arokem (commented):
Just a few small things

AFQ/definitions/utils.py
```diff
@@ -0,0 +1,120 @@
+__all__ = ["check_definition_methods", "StrInstantiatesMixin", "find_file"]
+
+
+def check_definition_methods(definition, definition_name=False):
```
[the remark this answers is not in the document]
36000: Done

AFQ/data.py
36000: `If specified, default should be a function that gets called for objects that can’t otherwise be serialized. It should return a JSON encodable version of the object or raise a TypeError. If not specified, TypeError is raised.`

When we write our metadata, sometimes what we try to write, like the seed_mask, are objects that are not always serializable. So the file still gets written, its just that non-serializable objects will be printed as "Not Serializable" instead of raising a TypeError.

AFQ/definitions/utils.py
36000: Not in python 3, but I will do it anyways

AFQ/data.py
36000: Actually, seed_mask was not the problem. I think that passing a RandomState() object to Segmentation is a better example. We will try to serialize that object when we write the metadata. Previously, it would crash. Now, it simply prints something like this: "rng: Not Serializable" in the long list of other parameters.

PR #627, opened 2021-02-01
More plotting tweaks, gridspec compat

36000:
Some of the tweaks to the plotting code to make it output SVGs and be compatible with gridspec (accept ax, fig as arguments).


36000:

PR #628, opened 2021-02-03
export endpoint ROI when saving intermediates

36000:
- [x] Test this on PREK data


36000:

PR #632, opened 2021-02-15
Relax versions to avoid cvxpy/numpy incompatibility

gkiar:
Some tools have issues with various version of numpy (e.g. [cvxpy](https://github.com/cvxgrp/cvxpy/issues/1229) versions before `1.1.10` were only compiled on pypi with numpy `1.20`), which can in turn cause a bunch of install issues.

This PR proposes that we relax the numpy requirement here, using a version that was fixed for pyafq 0.6 as the lower bound, and requiring the version stay below the latest minor release, 1.20, for the time being. We also update the cvxpy requirement to get past the fixed numpy issue from their side.

I've tested this in a config locally with numpy `1.19.1` and things work as expected.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
     nibabel==3.0.2
     dipy>=1.2.0
     scipy>=1.2.0
-    numpy==1.20.0
+    numpy>=1.18.5,<1.20
     pandas==1.0.3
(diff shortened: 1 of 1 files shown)


arokem:
LGTM! @36000 : does this look OK to you? If so, feel free to merge when you get a chance.

PR #634, opened 2021-02-16
Allow more flexibility in dask version

arokem:
This is a problem in the context of our hub configuration, where we need a match between the versions of `dask` and `dask-distributed`.


36000:
I added an upper bound and rebased. If this passes, I think we can merge.

PR #641, opened 2021-03-12
[FIX] Plotly Bundle Visualization fixes

36000:


bloomdt-uw:
@36000 wasn't sure if this `flip_axial` should also apply to `fury_backend` since the RAS/RAS+ should be same regardless of visualization. also doesn't look like this is available to change from downstream apis like `viz_bundles` or `export_all`

PR #643, opened 2021-03-15
Dipy should be at least 1.4.0

36000:


arokem:
Have you had a chance to experiment with patch2self? Is it providing clear benefit in any of our objective metrics (e.g., TRR)?

36000:
Not yet. I will do that on the pre-K data.

richford:
Very minor comment here. I believe the title of this PR should reference DIPY 1.4.0, rather than 1.14.0. Not a big deal at all but might help anyone in the future who is reviewing the change logs.

PR #646, opened 2021-03-16
WIP: Add OR fetcher and example

arokem:
- [ ]  Update example to use the API with OR-oriented config.

### DIFF
--- a/AFQ/data.py
+++ b/AFQ/data.py
@@ -480,12 +480,12 @@ or_md5_hashes = [
 ]
 
 fetch_or_templates = _make_fetcher("fetch_or_templates",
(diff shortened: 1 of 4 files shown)


36000:
Do you think this is caused by running out of RAM? We might need to break the docs up into separate github actions.

arokem:
I think that the failure might be related to https://github.com/dipy/dipy/pull/2210. PFT just occasionally segfaults without any obvious reason. I think that I will switch over to another tractography method here.

36000:
I just checked code, LGTM

arokem:
Can confirm that this works locally on my laptop as well

PR #649, opened 2021-03-17
Median tract profile

36000:
gives user option to calculate the median tract profile, as well as provide their own weighting scheme.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -417,10 +417,10 @@ class AFQ(object):
             raise TypeError(
                 "mapping must be a mapping defined"
                 + " in `AFQ.definitions.mapping`")
-        if not (profile_weights is None or
-                isinstance(profile_weights, str) or
-                callable(profile_weights) or
-                hasattr(profile_weights, "__len__")):
+        if not (profile_weights is None
+                or isinstance(profile_weights, str)
+                or callable(profile_weights)
+                or hasattr(profile_weights, "__len__")):
             raise TypeError(
                 "profile_weights must be string, None, callable, or"
                 + "a 1D or 2D array")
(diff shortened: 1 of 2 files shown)


AFQ/api.py
```diff
@@ -407,6 +417,18 @@ def __init__(self,
             raise TypeError(
                 "mapping must be a mapping defined"
                 + " in `AFQ.definitions.mapping`")
+        if not (profile_weights is None
+                or isinstance(profile_weights, str)
+                or callable(profile_weights)
+                or hasattr(profile_weights, "__len__")):
+            raise TypeError(
+                "profile_weights must be string, None, callable, or"
+                + "a 1D or 2D array")
```
arokem, line 426: That's very restrictive! 😄

AFQ/api.py
```diff
@@ -1353,11 +1378,32 @@ def _tract_profiles(self, row):
...
+                                values = np.array(
+                                    values_from_volume(
+                                        scalar_data,
+                                        fgarray,
+                                        row["dwi_affine"]))
+                                weights = np.zeros(values.shape)
+                                for ii, jj in enumerate(
+                                    np.argsort(values, axis=0)[
+                                        len(values) // 2, :]):
+                                    weights[jj, ii] = 1
```
arokem, line 1397: This also seems like a potentially useful way to find this: https://stackoverflow.com/a/64015677/3532933

arokem (commented):
Thanks for the quick implementation! Does it work as expected? I realize that this is a bit hard to test. Maybe worth pushing this upstream into dipy's `afq_profile`?

36000:
It works as expected. I think in DIPY I will wait for next week to do the PR, and there I will just calculate the median directly.

PR #650, opened 2021-03-18
Allow ItkMap in pyAFQ

36000:
Allows user to describe Itk map to use instead of our generated mappings. I tested its compatibility with a map from QSIprep. The example in the docstring should be helpful.

This touches a lot of files because I fixed a few dosctrings in other definitions.

### DIFF
--- a/AFQ/definitions/mapping.py
+++ b/AFQ/definitions/mapping.py
@@ -121,7 +121,7 @@ class ConformedFnirtMapping():
 
 class ItkMap(Definition):
     """
-    Use an existing Itk map. Expects the warp file
+    Use an existing Itk map (e.g., from ANTS). Expects the warp file
     from MNI to T1.
 
     Parameters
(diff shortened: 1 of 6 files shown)


arokem:
Just making sure: this addresses https://github.com/PennBBL/qsiprep/issues/237 ?

AFQ/definitions/mapping.py
```diff
@@ -113,6 +119,76 @@ def transform(self, data, **kwargs):
             + " only transform from template to subject space")
 
 
+class ItkMap(Definition):
+    """
+    Use an existing Itk map. Expects the warp file
```
arokem, line 124:
```suggestion
    Use an existing Itk map (e.g., from ANTS). Expects the warp file
```

arokem (commented):
Looks great! One small documentation comment, but otherwise merge whenever you think this is ready. This will save so much time...

PR #651, opened 2021-03-19
[FIX] Propogate flip axial to export_all

36000:
allow user to flip axial in plotly visualization from API

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1593,11 +1593,13 @@ class AFQ(object):
             xform_color_by_volume=xform_color_by_volume
         )
 
-        flip_axial = (row['dwi_affine'][0, 0] < 0)
+        flip_axes = [False, False, False]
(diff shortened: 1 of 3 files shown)


arokem:
What does this fix? Mis-labeling of R/L bundles?

36000:
When visualizing with plotly, sometimes the streamlines show up right/left flipped. I believe it is because sometimes indices count up from right to left and other times left to right.

36000:
In the plotly backend, you can now tell it to flip any of the three axes for each of the functions (bundles, ROIs, and volumes). In the AFQ object, it will use the affine to automatically detect which axis to flip. this has been tested with RAS and LAS but should work with any coordinate system.

arokem:
Overall, looks good. I just had that one comment about the documentation, to make sure that we're explicit that RAS is where we're going.

PR #652, opened 2021-03-23
Parallelize tracking

arokem:
Along the way, this also fixes a couple of bugs in mask handling in csd/dti, and allows passing an already-initialized direction getter object into the tracking, in case that's a thing you want to do (I use it to inject a bootstrap direction getter into the tracking, without giving it the raw data).


arokem:
One challenge with the parallelization done this way is that I can't seem to be able to use other backends than "threading" because the direction getter and tracker objects that get sent into the parallelized function cannot be pickled by python. I still need to do some more experiments with this to see whether this even provides any performance gains.

arokem:
I should try out [pathos](https://pypi.org/project/pathos/), which uses dill to serialize Python objects. Might get past the issues I am seeing here now.

arokem:
I am trying to make the Cython objects in DIPY pickle-able, based on https://ask.sagemath.org/question/8376/pickling-extension-classes/, but still doesn't work yet.

PR #653, opened 2021-03-23
[ENH] make cvxpy optional

36000:


arokem:
LGTM! Let's wait for the CI to be done, in case something unexpected comes up, but then merge.

AFQ/_fixes.py
```diff
@@ -228,6 +227,8 @@ def solve_qp(P, Q, G, H):
     x : array
         Optimal solution to the QP problem.
     """
+    import cvxpy as cvx
```
richford, line 230: Do you want to catch import errors here to provide a more helpful error message as you did for the different visualization backends? e.g.
```
try:
   import cvxpy as cvx
except ImportError:
   raise ImportError("Using MSMT in pyAFQ requires the the cvxpy package. You can install it using `pip install cvxpy` or `pip install pyAFQ[msmt]`.")
```
or some such.

PR #654, opened 2021-03-24
[FIX] remove invalid sls from tractography, which could be custom

36000:
I will test this on the optic bids data


36000:
ORs still look great after doing this. Cannot post the image because its UK biobank data. Would be good to know why the gpu tractography is creating these negative vertex indices, though.

PR #655, opened 2021-03-24
[WIP] docker workflow for pyafq and afqsi

36000:
Should automate the creation of a pyAFQ python version 0.8 image and pyafq/qsi python version 0.7 image


36000:
Moved to here so that it can see this repo's secrets: https://github.com/yeatmanlab/pyAFQ/pull/659

PR #656, opened 2021-03-26
BF: _gen_sl_counts function was failing with more than one subject

arokem:
Also, it was generating sl_counts of 0 in my experiments with it.

I think that the failure with more than one subject arises because
when one subject has that added to their row, it gets added to the columns
for the entire dataframe, but no value gets added in other rows. So,
when pandas gets to the next row it freaks out.

I suspect this is also the reason that the nightly3 tests are currently failing (?).

At any rate, I think this fix is more consistent with how we are doing
things with other files/quantities here. The only challenge is that it requires one
more read of the entire segmented tractogram from file, so that can add some time.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1505,10 +1505,10 @@ class AFQ(object):
             bundles = list(self.bundle_dict.keys())
             if "whole_brain" not in bundles:
                 bundles.append("whole_brain")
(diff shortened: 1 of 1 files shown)


36000:
Yes, this is way better. The reason I did not initially do this is because it adds another tractography load. However the current implementation has some problems that I have noticed as well, so I think this is the smart way to do it.

PR #657, opened 2021-03-26
[FIX] More lenient reco defaults

36000:
These are more lenient parameters suggested by Chandio in a recent meeting. They have yet to be tested on our HCP pipeline.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -52,10 +52,10 @@ class Segmentation:
                  progressive=True,
                  greater_than=50,
                  rm_small_clusters=50,
-                 model_clust_thr=5,
-                 reduction_thr=20,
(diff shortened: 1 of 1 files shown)


36000:
slr=False and refine=True may be warranted depending on results of HCP pipeline tests

arokem:
We should match as closely as possible the parameters here: https://github.com/dipy/dipy/blob/master/dipy/workflows/segment.py#L94-L109. I am not sure we're exactly matching now.

36000:
Right now we are matching the lenient parameters during the meeting. Those parameters are a little more strict. They may still also work. For example, its possible our problem was with model_clust_thr (before pr: 5; in this pr: 1.25; in that link: 2.5)

Issue #62, opened 2018-03-26
Dockerize pyAFQ


richford:
As suggested by @akeshavan in the AFQ-Browser meeting on 26 Mar 2018.

arokem:
See https://github.com/scitran-apps/afq for a dockerized version of the Matlab implementation.

PR #659, opened 2021-03-26
[ENH] docker workflow for pyafq and afqsi

36000:
Should automate the creation of a pyAFQ python version 0.8 image and pyafq/qsi python version 0.7 image


36000:
This closes #62 .

.github/workflows/docker_afqsi.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Checkout repo
+      uses: actions/checkout@v2
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Login to DockerHub
```
arokem, line 32:
```suggestion
    - name: Login to GHCR
```

.github/workflows/docker_pyafq.yml
```diff
@@ -0,0 +1,28 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Login to DockerHub
```
arokem, line 15:
```suggestion
    - name: Login to GHCR
```

afqsi_docker/docker-build.sh
```diff
@@ -0,0 +1,9 @@
+COMMIT=${1}
+COMMIT="$(echo "${COMMIT}" | tr -d '[:space:]')"
+export COMMIT
+TAG="ghcr.io/${2}/afqsi:${COMMIT}"
+TAG="$(echo "${TAG}" | tr -d '[:space:]')"
+echo $TAG
+docker build --no-cache -t $TAG --build-arg COMMIT ./afqsi_docker
+docker run --rm -i --entrypoint /usr/local/miniconda/bin/pytest $TAG /usr/local/miniconda/lib/python3.7/site-packages/AFQ/tests/test_api.py::test_AFQ_data_waypoint
```
arokem, line 8: Does this mean that the test is run every time a user calls this script to build the docker?

arokem: Or do we never use these scripts to build? The builds are only ever done on GitHub Actions?

arokem (commented):
Would be great to add some documentation about this, but we can also do that on a separate PR.

afqsi_docker/docker-build.sh
36000: The latter. The user is not meant to use this script. We use this script to build. I want to test that pyAFQ installed correctly in the qsiprep environment everytime we make a new image. I could probably have found a way to write all of this inside the workflow file, but I am more familiar with bash syntax, so I wrote these two little shell scripts instead.

PR #660, opened 2021-03-26
[WIP] [FIX] try to make pyafq play nice with pybids 0.9.3

36000:
this is the version qsiprep uses

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,7 +46,7 @@ install_requires =
     popylar==0.2.*
     imageio>=2.8.0,<3.0.0
     templateflow==0.6.2
-    pybids>=0.11.0,<0.13.0
+    pybids==0.9.3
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -46,7 +46,7 @@ install_requires =
     popylar==0.2.*
     imageio>=2.8.0,<3.0.0
     templateflow==0.6.2
-    pybids>=0.11.0,<0.13.0
+    pybids==0.9.3
```
arokem, line 49: Can we do this?
```suggestion
    pybids>=0.9.3, <0.13.0
```
to be a little bit more flexible and avoid potential dependency conflicts?

36000: Yes I am just forcing this version to run the CLI tests on it for now (: . I will switch it to that once they pass

Issue #408, opened 2020-09-07
Syn Registration degrees of freedom


36000:
It should be possible for the user to specify more parameters in syn registration.

Issue #558, opened 2020-10-19
Use functions from dipy.align public API, instead of AFQ.registration


arokem:
In https://github.com/dipy/dipy/pull/2025, I pushed much of the functionality in AFQ.registration upstream into DIPY. This PR was now merged, and will be part of the upcoming 1.3 release. Once that is out, we should refactor here to remove AFQ.registration (or as much of it as possible) in favor of direct calls to DIPY.

PR #661, opened 2021-03-29
[ENH] A variety of registration improvements, primarily for babyAFQ

36000:
This PR:
1. use dipy functions instead of some of the functions we have in afq.registration . We moved afq.registration functions to dipy but have not started using the dipy functions yet. This PR gets the low hanging fruit. 
2. Give user the option to pass kwargs to slr and syn registration
3. Fixes AffMap to actually apply the correct affine
4. removes convex hull adjustment for transformed ROIs. This was dilating them.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1703,7 +1703,7 @@ class AFQ(object):
                             pp.get_fdata(),
                             self.reg_template_img,
                             pp.affine,
-                            self.reg_template_img.affine)
+                            self.reg_template_img.affine).get_fdata()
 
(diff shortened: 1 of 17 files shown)


36000:
partially addresses #558

36000:
closes #408

36000:
Should we make affine the default registration technique? What parameters can we tweak to constrain SyN registration?

36000:
image
From left to right: ROI in template, ROI transformed to subject without convex hull, ROI transformed to subject with convex hull.

PR #664, opened 2021-04-01
[Doc] try new docs layout

36000:
Trying out different documentation layouts.

Is there an easy way to generate the docs without running the examples?


arokem:
I think that commenting out [this line](https://github.com/yeatmanlab/pyAFQ/blob/master/docs/source/conf.py#L51) will disable sphinx gallery

36000:
Wow! I really like this new layout.

arokem:
Yeah. It's very nice. 

We might have to redo this diagram on the front page:

PR #665, opened 2021-04-06
[FIX] Minor BF for nightly tests

36000:
One is a copy/paste typo in SLR map.

The other is that when sl counts was removed from the AFQ object data_frame, the corresponding test had to be updated.

Also a minor change in the order of export all, so that the profiles are combined right after they are generated.


arokem:
Looks good! I will wait for the CI to complete and then merge.

36000:
@arokem should we cut a release once this is merged?

arokem:
Yes. I'll go ahead and do that now!

PR #669, opened 2021-04-08
[WIP] [FIX] Try modin for parallelization

36000:
This may be an easy way of fixing the parallelization code: https://github.com/modin-project/modin


arokem:
I'm thinking of giving [pathos](https://pypi.org/project/pathos/) a try over on https://github.com/yeatmanlab/pyAFQ/pull/652. Might help here as well? Maybe it's time to refactor our dataframe-driven dataflow? Try out pydra? It would be a big lift, but it might be worth doing at some point (?)

Issue #666, opened 2021-04-07
Build a docker image with a tag that matches GitHub tag


arokem:
I think that the docker image is currently built every time we push to master and then gets tagged with the sha of that commit. But it would be nice if we also had docker tags corresponding to our github tags, so we could do something like: `docker pull ghcr.io/nrdg/pyafq:0.7` to get the 0.7 release.

PR #677, opened 2021-04-14
Builds a tagged image when a tag is pushed.

arokem:
Addresses #666 

Based on https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions


36000:
LGTM

PR #689, opened 2021-05-20
Allow for other derivatives folders when downloading and combining

36000:
- [x] test

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -2231,7 +2231,8 @@ class AFQ(object):
 
 
 def download_and_combine_afq_profiles(bucket,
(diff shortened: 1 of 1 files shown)


arokem:
I think this makes sense. Is the use-case you are looking at right now downloading other tractography pipelines? Or downloading preprocessed data?

36000:
This is for downloading tract profiles that might not be in derivatives/afq . For example, if you run pyafq multiple times and put them in different folders, like derivatives/afq_csd . This was prompted by a problem Sendy had, which we just tested and this fixes, so I think it is good to merge now.

Issue #674, opened 2021-04-12
Joblib artifacts remain after segmentation


ValHayot:
After executing the segmentation pipeline, I noticed that Joblib Loky processes continue to run.

`ec2-user    5410  5.6  0.2 1150616 182724 pts/0  S    00:06   0:33 /home/ec2-user/venv/bin/python -m joblib.externals.loky.backend.popen_loky_posix --process-name LokyProcess-24 --pipe 45`

This appears to possibly be normal behavior ( [Joblib issue#945](https://github.com/joblib/joblib/issues/945)) except there wasn't a timeout in my runs and memory usage appeared to explode when running segmentation in a loop.

PR #691, opened 2021-05-28
[ENH] Provide more informative errors for incorrect BIDS structure, generate bundle dict lazily

36000:
This provides informative errors if no files are found inside the entire BIDS layout or in the specific derivatives folder, which would be a result of incorrect BIDS structure.

The added test adds ~5 minutes to the test because initializing api.AFQ objects takes ~1.5 minutes each.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -122,7 +122,8 @@ class BundleDict(MutableMapping):
                     expanded_bundle_names.append(bundle_name + "_R")
                     expanded_bundle_names.append(bundle_name + "_L")
             elif self.seg_algo.startswith("reco"):
-                if bundle_name in RECO_UNIQUE:
+                if bundle_name in RECO_UNIQUE\
+                        or bundle_name == "whole_brain":
                     expanded_bundle_names.append(bundle_name)
                 else:
                     expanded_bundle_names.append(bundle_name + "_R")
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -270,6 +270,9 @@ class BundleDict(MutableMapping):
     def __iter__(self):
         return iter(self._dict)
 
+    def copy(self):
+        return self._dict.copy()
+
 
 # this is parallelized below
 def _getter_helper(wf_dict, attr_name):
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -108,38 +108,38 @@ class BundleDict(MutableMapping):
             self.bundle_names = list(bundle_info.keys())
             self._dict = bundle_info.copy()
         else:
-            self.bundle_names = bundle_info.copy()
+            expanded_bundle_names = []
+            for bundle_name in bundle_info:
+                if self.seg_algo == "afq":
+                    if bundle_name in ["FA", "FP"]\
+                            or bundle_name in CALLOSUM_BUNDLES:
+                        expanded_bundle_names.append(bundle_name)
+                    else:
+                        expanded_bundle_names.append(bundle_name + "_R")
+                        expanded_bundle_names.append(bundle_name + "_L")
+                elif self.seg_algo.startswith("reco"):
+                    if bundle_name in RECO_UNIQUE\
+                            or bundle_name == "whole_brain":
+                        expanded_bundle_names.append(bundle_name)
+                    else:
+                        expanded_bundle_names.append(bundle_name + "_R")
+                        expanded_bundle_names.append(bundle_name + "_L")
+                else:
+                    raise ValueError(
+                        "Input: %s is not a valid input`seg_algo`"
+                        % self.seg_algo)
+            self.bundle_names = expanded_bundle_names
             self._dict = {}
         self.seg_algo = seg_algo.lower()
 
-        expanded_bundle_names = []
-        for bundle_name in self.bundle_names:
-            if self.seg_algo == "afq":
-                if bundle_name in ["FA", "FP"]\
-                        or bundle_name in CALLOSUM_BUNDLES:
-                    expanded_bundle_names.append(bundle_name)
-                else:
-                    expanded_bundle_names.append(bundle_name + "_R")
-                    expanded_bundle_names.append(bundle_name + "_L")
-            elif self.seg_algo.startswith("reco"):
-                if bundle_name in RECO_UNIQUE\
-                        or bundle_name == "whole_brain":
-                    expanded_bundle_names.append(bundle_name)
-                else:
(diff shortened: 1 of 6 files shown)


arokem:
Ouch. This is because of the bundle dict, isn't it? Is there any way we can defer reading in that data?

36000:
Maybe I could make bundle dict into a class, then move all of the bundle dict code into that class, and it would only do the actual calculations when it has to.

36000:
This commit brings it down to ~15 seconds by rewriting BundleDict as a class that inherits dict, but generates the bundle dict lazily.

AFQ/api.py
```diff
@@ -71,135 +72,206 @@ def do_preprocessing():
...
+                for ii, roi in enumerate(rois):
+                    self.bundle_dict_resampled[bundle]['ROIs'][ii] =\
+                        afd.read_resample_roi(
+                            roi, resample_to=self.resample_to)
+        else:
+            self.bundle_dict_resampled = self.bundle_dict_unresampled
+
+        self._dict = self.bundle_dict_resampled
+
+    def __getattribute__(self, attr):
```
arokem, line 252: I bet this could be made to be even lazier to set only the specific bundle each time, by passing the key into `_gen_resampled_dict`. But maybe that's not crucial 😄

36000:
I am having trouble pinning down this test suite error, as I cannot see what is causing it on the suite and I do not run into it locally.

AFQ/api.py
36000: I like this idea!

36000:
So the only remaining problem here is that its causing OOM in doc build. This is odd because I am not sure how this would increase memory usage by much at all?

AFQ/api.py
```diff
@@ -71,135 +72,223 @@ def do_preprocessing():
...
+        """
+        if not (isinstance(bundle_info, dict)
+                or isinstance(bundle_info, list)):
+            raise TypeError((
+                f"bundle_info must be a dict or a list,"
+                f" currently a {type(bundle_info)}"))
+
+        if isinstance(bundle_info, dict):
+            self.bundle_names = list(bundle_info.keys())
+            self._dict = bundle_info.copy()
```
arokem, line 109: Maybe this is causing the OOM? It looks like you are making a copy of the whole thing?

36000: No, here we are copying the user input (bundle_info) if it is a dict. It was the same in the old implementation. That way pyAFQ does not modify the user's passed in dictionary.

arokem: Oh yeah - I see. Hmmm.

arokem: Curiouser and curiouser: I thought maybe the BundleDict has a larger memory footprint than the dict implementation, but it seems to have a _smaller_ memory footprint. Baffling.

arokem: Maybe the OOM is related to this issue somehow? https://github.com/yeatmanlab/pyAFQ/issues/674

I see some loky messages in the documentation build logs that are throwing the OOM. I wonder if MutableMapping is somehow more resistant to garbage collection than a vanilla dict.

arokem:
FWIW, it doesn't always fail in the same place during the doc build. I reran that pipeline and now it failed at another spot, so we're not looking at some obvious huge memory bottleneck, but something a bit more intermittent (and hence, probably harder to pin down...).

36000:
When we generated individual bundles on the fly, we had to keep the afq and colossal templates around with the bundle dict object. This might have been causing the OOM errors? I tried deleting elements from the templates as we moved them into the bundle dict, but this did not solve the problem.

Instead the problem is solved by generating all of the bundles at once, which I think is fine since we seldom need only one bundle at a time. In this case we do not have to keep the afq/callosal templates attached to the bundle dict object.

So, I think this is ready for review/merge @arokem

AFQ/api.py
```diff
@@ -71,135 +72,216 @@ def do_preprocessing():
...
+                    "Input: %s is not a valid input`seg_algo`"
+                    % self.seg_algo)
+        self.bundle_names = expanded_bundle_names
+
+        # Each bundles gets a digit identifier
+        # (to be stored in the tractogram)
+        self.uid_dict = {}
+        for ii, b_name in enumerate(self.bundle_names):
+            self.uid_dict[b_name] = ii + 1
+            self.c_uid = ii + 2
```
arokem, line 142: Just making sure that I understand: this is to that when we call `__setitem__` you have an integer ready to give it? Maybe make this a private attribute and add a comment on this line?

AFQ/api.py
```diff
@@ -71,135 +72,216 @@ def do_preprocessing():
...
+        return self._dict[key]
 
-    return afq_bundles
+    def __len__(self):
+        return len(self.bundle_names)
+
+    def __delitem__(self, key):
+        if key in self._dict:
+            del self._dict[key]
+        if key in self.bundle_names:
```
arokem, line 275: Do you need to raise if it's not in either/both?

AFQ/api.py
```diff
@@ -71,135 +72,216 @@ def do_preprocessing():
...
+        else:
+            raise ValueError(
+                "Input: %s is not a valid input`seg_algo`" % self.seg_algo)
+        self.all_gen = True
+
+    def __setitem__(self, key, item):
+        self._dict[key] = item
+        self.uid_dict[key] = self.c_uid
+        self.c_uid += 1
+        self.bundle_names.append(key)
```
arokem, line 263: Do you need to reset `all_gen` to `False` here, if it's currently set to `True`?

arokem (commented):
OK - a couple more thoughts/questions here. 

Also, a thought about naming: instead of `BundleDict` maybe something like `BundleInfo` or `BundleMapping`, given it's not really a sub-class of a dict?

36000:
I intend `BundleDict`  to be used as a dict, I use `MutableMapping` because you are supposed to inherit that instead of dict (see here for explanation: http://www.kr41.net/2016/03-23-dont_inherit_python_builtin_dict_type.html) (I have read the same thing on other blogs as well)

AFQ/api.py
36000: Yes, I agree

AFQ/api.py
36000: I clarified the use of __setitem__

arokem:
Great! Just verifying: this is done, right? Feel free to merge if it is.

PR #693, opened 2021-05-29
[WIP] Allow use of Fnirt mapping with recobundles

36000:
This is important for UKbio CNII recognition

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -24,6 +24,7 @@ import AFQ.utils.models as ut
 import AFQ.utils.volume as auv
 import AFQ.data as afd
 from AFQ.utils.parallel import parfor
+from AFQ.definitions.mapping import ConformedFnirtMapping
 
 __all__ = ["Segmentation", "clean_bundles", "clean_by_endpoints"]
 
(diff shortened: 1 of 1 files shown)


arokem:
LGTM! Feel free to merge when you think this is ready.

36000:
Split this PR into original part and this new fix

36000:
I don't think there is an easy way to do this, because recobundles requires a transformation from template to subject space, however either I don't think the UK biobank provides that transformation, or, I am misunderstanding the fslpy interface. I am tempted to drop this as I am not sure if the CNII will actually end up being useful, reliability-wise.

arokem:
Isn't that the inverse of the transform in the opposite direction? That is, take each vector and rotate it 180 degrees. Apologies if this something that you've already tried. Looks like you've tried a lot of different things...

36000:
Would that work? Wouldn't you also have to shift those vectors by themselves, i.e., if a vector points from voxel a to voxel b, you don't just flip the vector, you also translate late it so it points from voxel b to voxel a?

36000:
the reason there are so many commits is because I am using this branch on UKbio server to test it, I will clean up the commit history before this is ready to merge.

36000:
@arokem is there an easy existing method for converting "absolute coordinates in the source space" to "relative displacement from the reference image space to the source"

arokem:
I'm not sure... it might be hiding somewhere in the DIPY source code for the non-linear registration.

arokem:
But I think you're right about this: https://github.com/yeatmanlab/pyAFQ/pull/693#issuecomment-853843084

36000:
I figured out all the nonlinear stuff I think, now I have been messing around with the affines. Can't quite seemed to get it right. There are a lot of affines running around: their template space, our template space, the subject space, as well as FSL coordinates vs voxel coordinates. Right now the affines are ever so slightly off.

arokem:
Ever so slightly, like an off-by-one? That sometimes happens if Matlab is in the mix.

36000:
More like 10% too big and 10% shifted in a direction, which is a lot closer than it starts out as! I have been trying out different affine orderings and trying to understand the different spaces everything can be in, as well as trying to replicate what the fslpy people are doing in their code, which we know works for one direction.

Issue #702, opened 2021-06-04
AFQ browser integration


36000:
Proposed changes to pyAFQ:

- [x] Make sure that pyAFQ's output tract profiles do not have the index column that pandas attaches by default
- [x] Update pyAFQ to have a function to generate an afq browser instance
- [x] generate streamlines.json from just first subject (How to handle no streamlines found?)

Proposed changes to the AFQ-Browser assemble API:

- [x] source can be existing nodes.csv file (this makes AFQ-Browser more easily compatible with other pipelines) (document this as pyAFQ first)
- [x] Add another streamlines.json path argument (note to the user that needs to be in MNI)
- [x] Extra: additional input that would be a dictionary, whose keys are subject/session and values are paths to plotly visualizations, these links show up somewhere in AFQ-Browser

36000:
Make sure that pyAFQ's output tract profiles do not have the index column that pandas attaches by default

bloomdt-uw:
There is an existing Helper function to create a minimal metadata file:

https://github.com/bloomdt-uw/AFQ-Browser/blob/fcb5222417c020afebaebab4307424b22f154c1e/afqbrowser/browser.py#L75-L79

36000:
Useful AFQ-browser links:
https://yeatmanlab.github.io/AFQ-Browser/dataformat.html
https://github.com/yeatmanlab/AFQ-Browser/blob/master/bin/afqbrowser-assemble

Example working pyAFQ-browser (took assembled instance, replaced nodes.csv and subjects.csv):
https://github.com/yeatmanlab/UW_PREK_pyAFQ_pre_browser

36000:
Proposed changes to the AFQ-Browser assemble API:

- [ ] source can be existing nodes.csv file (this makes AFQ-Browser more easily compatible with other pipelines)
- [ ] streamlines.json file should be an optional path just like metadata
- [ ] streamlines.json input could also be a dictionary, whose keys are subject/session and values are paths to plotly visualizations
- [ ] User tells the program if they are using matlab vs tracula vs nodes.csv (at the very least, throw an error earlier if input does not match one of these)

bloomdt-uw:
https://github.com/yeatmanlab/AFQ-Browser/pull/287

Adds support for `nodes.csv`, optional metadata, optional streamline file, and validates inputs.

To use can either call the `afqb.assemble` API or use `bin/afqbrowser-assemble` CLI

[...]

PR #703, opened 2021-06-04
[ENH] AFQ-Browser Integration

36000:
See #702 for gameplan.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1083,7 +1083,7 @@ class AFQ(object):
             "subjectID" column to work.
             Default: None
         page_title : str
-            Page title.
+            Page title. If None, "AFQ Browser" is used.
             Default: None
         page_subtitle : str
             Page subtitle.
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -1182,16 +1182,15 @@ class AFQ(object):
         # generate streamlines.json file
         sls_json_fname = self.get_streamlines_json()
 
-        # afqb.assemble(
-        #     op.abspath(op.join(self.afq_path, "tract_profiles.csv")),
-        #     target=output_path,
-        #     metadata=metadata,
-        #     # sls_json=sls_json_fname,
-        #     # mode="custom",
-        #     title=page_title,
-        #     subtitle=page_subtitle,
-        #     link=page_title_link,
-        #     sublink=page_subtitle_link)
+        afqb.assemble(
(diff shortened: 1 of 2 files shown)


AFQ/api.py
```diff
@@ -924,9 +929,38 @@ def combine_profiles(self):
...
+                        sft.streamlines[idx],
+                        img,
+                        Space.VOX)
+                    # TODO: transform this_tg to MNI and rasmm
+                    # TODO: calculate core fiber
+                    # TODO: output json file
+
+        return sls_json_fname
+
     def export_all(self):
```
36000, line 964: this calls assemble afq browser, optionally turn off expensive things like plotly viz, afq browser, individual bundles, etc.

36000:
This is almost good to go. The streamlines.json generated looks good, but for some reason does not work in AFQ browser. I think we need a better understanding of the streamline keys that afq browser suspects. Attached is a streamlines.json file generated from the stanford hardi dataset.

[afqb_streamlines.txt](https://github.com/yeatmanlab/pyAFQ/files/6611635/afqb_streamlines.txt)

36000:
The problem is how to handle no streamlines found for a given bundle. AFQ-browser expects at least 4 distinct points per bundle. I do not know the best way to handle this.

36000:
image

@bloomdt-uw @arokem  mission accomplished.

Note that this example is from the test we run on every commit and does not have many streamlines (they are subsampled).
This autogenerates on export_all call in afqbrowser folder.

setup.cfg
```diff
@@ -57,6 +57,8 @@ install_requires =
     pip-conflict-checker==0.6.*
     ipython>=7.13.0,<=7.20.0
     pimms
+    AFQ-Browser @ git+https://github.com/yeatmanlab/AFQ-Browser@83cc7261198b1a26b6ba07c323f94a2418cb9c82
```
arokem, line 60: Before we merge this, I should address https://github.com/yeatmanlab/AFQ-Browser/issues/289, so that we can pin a version from pypi here.

arokem (commented):
Looks good! I think that ball is in my court to update things on AFQ-Browser and cut a new release of that, before we merge this.

setup.cfg
arokem: I think this can be updated to:

```suggestion
    AFQ-Browser>=0.3
```

PR #705, opened 2021-06-11
[FIX] Mask getters have to be aware of whether they are being called from data task

36000:
This should fix the nightly pft task


36000:
@arokem I just tested this locally, it fixes the relevant test.

PR #708, opened 2021-06-15
BF: Fix docker builds 

arokem:
Based on https://github.community/t/how-to-get-just-the-tag-name/16241/21


arokem:
I'll go ahead and merge this, because it's broken on master and the only way to know whether it works is to merge it into master and trigger the action. To know whether this really fully works, we'd also need to create a tag/release.

PR #710, opened 2021-06-15
Move AAL atlas to figshare

arokem:
When the UW library's website is down for maintenance (as it is right now), the AAL atlas, which is stored there, becomes inaccessible, and pyAFQ raises an error when it tries to find it (but only on new installations, for example in ephemeral cloud computing resources). I am hoping that figshare is less likely to go down for maintenance without providing an alternative.


36000:
LGTM

PR #712, opened 2021-06-21
BF: loop over valid_sub and valid_ses lists correctly

36000:
This fixes a bug introduce in the Pimms PR. The master code in api.AFQ.__getattribute__ has a bug, it iterates over every subject in valid_sub_list and then for each subject iterates over every session in valid_ses_list. The valid sub and ses lists are not constructed this way, however, because not every subject is in every session. The master code is redundant and may lead to subject/session combinations that are not valid.


36000:

PR #718, opened 2021-06-24
[FIX] fix custom bundldict bug and add test

36000:


arokem:
Yep - looks good! I'll wait for the CI to return green and then rebase my PR and add the test you suggested.

arokem:
OK - looks like the tests are passing. Merging!

PR #725, opened 2021-06-30
[FIX] explicitly require afqbrowser dependencies

### LINKED ISSUE #724: Require git-python
This is a dependency of AFQ-browser that we can explicitly check for

36000:
Closes #724

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,8 @@ install_requires =
     ipython>=7.13.0,<=7.20.0
     pimms
     AFQ-Browser>=0.3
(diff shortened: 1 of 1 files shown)


36000:

arokem:
I am a bit confused: doesn't requiring that version of AFQ-Browser automatically pull in all of its requirements? Would we need to keep updating these requirements in two different places, as changes occur in AFQ-Browser?

richford:
I think this might be a non-issue. @36000, unless you've seen this error elsewhere, we should chalk this up to my defective conda environment.

PR #729, opened 2021-07-22
[DOC] Generate simple docs for export function outputs

36000:
This adds some basic documentation for the myafq.export_ functions (like myafq.export_b0()). This documentation can be seen by doing help(myafq). I will post a picture shortly of what these changes look like.

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -23,14 +23,36 @@ from AFQ.models.dti import _fit as dti_fit_model
 DIPY_GH = "https://github.com/dipy/dipy/blob/master/dipy/"
 
 
-outputs = [
-    "data", "gtab", "img", "b0_file", "masked_b0_file", "dti_tf",
-    "dti_params_file", "dki_tf", "dki_params_file", "csd_params_file",
-    "pmap_file", "dti_fa_file", "dti_cfa_file", "dti_pdd_file",
-    "dti_md_file", "dki_fa_file", "dki_md_file", "dki_awf_file",
-    "dti_ga_file", "dti_rd_file", "dti_ad_file",
-    "dki_ga_file", "dki_rd_file", "dki_ad_file",
-    "dki_rk_file", "dki_ak_file", "dki_mk_file", "brain_mask_file"]
+outputs = {
+    "data": """DWI data as an ndarray for selected b values.
+    If patch2self is used, it is already applied to this data""",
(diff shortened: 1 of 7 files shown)


36000:
image

36000:
@arokem what do you think of this system for documenting the export_ functions? Developers will have to add a description of the function output(s) to the outputs variable in the task file they are updating, so it makes it convenient for developers. Is there anything else we should add to the docstrings for the export_ functions?

AFQ/tasks/data.py
```diff
@@ -23,12 +23,28 @@
...
+    "pmap_file": """anisotropic power map file""",
+    "dti_fa_file": """DTI fractional anisotropy file""",
+    "dti_cfa_file": """DTI color fractional anisotropy file""",
+    "dti_pdd_file": """DTI PDD file""",
+    "dti_md_file": """DTI mean diffusivity file""",
+    "dki_fa_file": """DKI fractional anisotropy file""",
+    "dki_md_file": """DKI mean diffusivity file""",
+    "dki_awf_file": """DKI axonal water fraction file""",
+    "dki_mk_file": """DKI mean kurtosis file""",
+    "brain_mask_file": """brain mask file"""}
```
arokem, line 47: I think we could put a little bit more information here. 

For example: `"""full path to a nifti file containing a brain mask"""`

Or: `"""full path to a nifti file containing DKI mean diffusivity"""`

arokem (commented):
I think this is good. I would suggest adding just a tiny bit more information about each output. I.e., what the type is of this output (e.g., "full path to a nifti file"), see examples inline:

arokem:
Looks like the documentation build is running out of memory in GitHub Actions?

36000:
I am not sure why those memory errors were happening. What do PDD and GA stand for in the context of DTI?

36000:
This is ready for another review/merge

arokem:
PDD is the principal diffusion direction of the diffusion tensor. It's the eigenvector associated with the first eigenvalue, so requires no additional calculation above and beyond what FA and MD and others already require. 

GA is geodesic anisotropy. See code [here](https://github.com/dipy/dipy/blob/master/dipy/reconst/dti.py#L92-L170). Also doesn't look like it's particularly computationally demanding, except that there is an allocation of three empty volumes the XYZ size of the data on L157 - L159.

PR #731, opened 2021-07-31
DOC: Insert the current version into the documentation.

arokem:


arokem:
I was also trying to get the logo to appear in the top left corner of the front page (it's currently missing), but maybe we can take that on a separate PR. Things got a bit complicated after I upgraded sphinx on my laptop 😦

PR #732, opened 2021-08-04
[DOC] Adds an example to the custom scalar docs

36000:
@arokem thoughts?


arokem:
I think that this will be more realistic if we use an example of a diffusion analysis pipeline instead of freesurfer. For example, we can use "noddi" as the pipeline and show how to add "ICVF" and "ODI" maps to the tractometry. Does that make sense? I think that's a fairly realistic use-case.

PR #733, opened 2021-08-04
[ENH] add more scalars, add one to the test

36000:
These scalars include scalars that mAFQ already calculates.


arokem:
This is great. I wish that we had some way to introspect DIPY 'Fit' objects to find all of these metrics. It's a bit tricky because not all attributes of a 'Fit' object are metrics, and not all of the metrics behave the same way. For example, they're not all decorated with the `auto_attr` decorator, though many are (for example MK isn't). In general, it would be neat if all of this boiler plate was automatically generated somehow. But probably a project for another day 😄

36000:
I agree, it would be nice if we could use introspection to generate these.

PR #734, opened 2021-08-05
Adding a citation file.

arokem:
This is a new format that GitHub supports: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files. I did a bit of a hack on it here, and I wonder if it will work this way. We do want people to cite our paper moving on, not the repo...

### DIFF
--- a/.github/CITATION.cff
+++ b/.github/CITATION.cff
@@ -1,0 +1,2 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite our paper: https://www.biorxiv.org/content/10.1101/2021.02.24.432740v2"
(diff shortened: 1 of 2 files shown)


arokem:
While I am at it, I added some more citation information to the front page of the online docs.

36000:
Should we also add this to the readme.md to draw attention to ti?

arokem:
I would love to be able to avoid duplicating the contents of these. Maybe something like this: https://stackoverflow.com/questions/46278683/include-my-markdown-readme-into-sphinx

36000:
Following that link, we would move stuff into our readme.md then link the the readme from the webpage. Currently our readme links to the webpage, which is simpler. After thinking about this more I think our current system is good, this LGTM

arokem:
OK. I'll wait for the CI to render this and then go through and see that that there are no issues with links/references and if it's all 👍, I will merge.

PR #736, opened 2021-08-12
[ENH] Replicating mAFQ visualizations using our rendering framework

36000:
A user found these plots to be particularly helpful:
image

So I am trying to replicate these plots within our existing visualization framework. 
I am breaking this down into steps:

- [x] Show tract profiles with shading that is the same as the corresponding plotly visualization
- [x] Merge these into one picture
- [x] Add the 0, 50, and 100 marks to the plotly visualizations
- [x] Add the core streamline 3d visualization and not-core profiles
- [x] Color by streamline

### DIFF
--- a/AFQ/tasks/viz.py
+++ b/AFQ/tasks/viz.py
@@ -71,7 +71,7 @@ def viz_bundles(subses_dict,
 
     figure = make_subplots(
         rows=1, cols=2,
-        specs=[[{"type": "scene"}, {"type": "xy"}]])
+        specs=[[{"type": "scene"}, {"type": "scene"}]])
 
     figure = viz_backend.visualize_volume(
         volume,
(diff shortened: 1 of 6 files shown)


36000:
Here is an example of what this looks like with the first two bullet points done:

https://gistcdn.githack.com/36000/8c859e43095247aab646887d63fe88aa/raw/b5d42637805087f6fd5b5b554009d8fa3d36aab3/viz.html

36000:
Have an API which takes 100 numbers (a profile) and a trk file, and then returns the matlab-like visualization with core fiber and colormap.

36000:
Be able to turn on and off the volume slices

36000:
Here is an example output from a function that takes 100 numbers (a profile) and a trk file:

image

jyeatman:
Can you add interpolation so that the colors change continuously?
Also, for the core fiber tube, can you render it as a 3d shaded surface?

36000:
It is a 3d surface, I will play around with plotly's lighting settings so that it looks shaded. 
For the profile, I will try to make it an interpolated line plot.

36000:
I am also getting rid of the axes on the core bundle plot. If the user wants to use volume slices, they can add those using the visualize_volume function.

36000:
image
I tried to do some shading without switching to rendering as. a 3d surface. I can try to switch to defining a 3d mesh which would be a tube-shaped surface around the core, but I wanted to see if this looks good enough first.

36000:
@arokem if this passes, this should be ready to merge. I consider this a good first draft at these plots, and I can come back and further refine them if people find them useful.

arokem:
Yeah - looks great! The documentation examples are really neat. This comes pretty close to the AFQ-Browser functionality! 🚀

PR #739, opened 2021-09-15
[BF] ITK and FNIRT mappings had typo reversing reg_subject and reg_template

36000:
This breaks these mappings, and was introduced when introducing PIMMS. It would be nice to have smoke tests for these mappings, however then we would need to have fnirt/itk mappings generated for the tests.

### DIFF
--- a/AFQ/definitions/mapping.py
+++ b/AFQ/definitions/mapping.py
@@ -89,7 +89,7 @@ class FnirtMap(Definition):
 
         self.fnames[session][subject] = (nearest_warp, nearest_space)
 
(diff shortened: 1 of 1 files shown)


36000:

arokem:
Just checking: how do you know that this is correct? This is based on some experiments that you did? Maybe you can just describe them here, so that we can refer back to this PR if issues arise around this at some later point?

36000:
To check this is current, I ran pyAFQ using this version on UKBB subjects who have the FNIRT mapping

PR #745, opened 2021-10-01
Allow other extensions than nii.gz to be given by the user for optional input files

36000:


arokem:
What is the use-case for this?

36000:
Iliana & Sendy are working on a dataset where there are image files in "mgz" format. They want to do something like:
```
brain_mask = LabelledMaskFile(
    "seg",
    {"scope": "freesurfer", "extension": "mgz"},
    exclusive_labels=[0])
```

PR #755, opened 2021-10-14
[ENH] Change default BundleDict behavior to resample

36000:
Users who specify a custom bundledict usually still want to resample, and might miss the `resample_to` parameter.

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -92,7 +92,7 @@ class BundleDict(MutableMapping):
     def __init__(self,
                  bundle_info=BUNDLES,
                  seg_algo="afq",
-                 resample_to=afd.read_mni_template()):
+                 resample_to=None):
         """
         Create a bundle dictionary, needed for the segmentation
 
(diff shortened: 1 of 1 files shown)


AFQ/api.py
```diff
@@ -92,7 +92,7 @@ class BundleDict(MutableMapping):
     def __init__(self,
                  bundle_info=BUNDLES,
                  seg_algo="afq",
-                 resample_to=False):
+                 resample_to=afd.read_mni_template()):
```
arokem, line 95: This worries me a little bit. Having a function call in the signature of another function seems a bit risky. Doesn't this trigger the mni download on import? How about setting this to `None` and then have the default behavior be as follows? 

```
if resample_to is None: 
    resample_to = afd.read_mni_template()
```

arokem: Or is that too confusing? Because `None` sounds like you are not resampling?

arokem:
LGTM! Merging.

PR #756, opened 2021-10-18
Enh: rename this variable (thoughts?)

36000:


arokem:
Yeah - agreed - this is an improvement.

PR #764, opened 2021-11-02
Group AFQ / Participant AFQ

36000:
ending the god object by re-organizing variables into task modules and splitting api.AFQ into group and participant

- [x] move find path calls into when they are needed
- [x] Update documentation and add documentation to task modules, which is automatically scanned and attached to API modules
- [x] Integrate the individual kwargs into config parser
- [x] Fix bugs, allowing things to actually run
- [x] Change definitions to allow paths to be given
- [x] integrate config.json changes (means allowing choice of participant or group for cl)
- [x] add auto-generated doc pages from help(GroupAFQ) and/or help(ParticipantAFQ)
- [x] Consider reg_template_spec, bundle_dict_spec, other specs which may move code out of get_plans?

Issue #763, opened 2021-11-02
Trouble running the CLI


arokem:
Here is what the file-system in the input folder looks like: 

```
.
├── dataset_description.json
└── derivatives
    └── qsiprep
        ├── dataset_description.json
        └── dwi
            ├── sub-DWITest01_ses-DWI_dir-PA_space-T1w_desc-brain_mask.nii.gz
            ├── sub-DWITest01_ses-DWI_dir-PA_space-T1w_desc-preproc_dwi.bval
            ├── sub-DWITest01_ses-DWI_dir-PA_space-T1w_desc-preproc_dwi.bvec
            └── sub-DWITest01_ses-DWI_dir-PA_space-T1w_desc-preproc_dwi.nii.gz
```

And here's what I have in the config file: 

```


# Initialize an AFQ object.
# Some special notes on parameters:
# In tracking_params, parameters with the suffix mask which are also
# a mask from AFQ.definitions.mask will be handled automatically by the
# api.

# Use '' to indicate None
# Wrap dictionaries in quotes
# Wrap definition object instantiations in quotes

[BIDS]

# The path to preprocessed diffusion data organized in a BIDS
# dataset. This should contain a BIDS derivative dataset with
# preprocessed dwi/bvals/bvecs.
bids_path = "/path/to/bids/folder/"

# Filter to pass to bids_layout.get when finding DWI files.
# Default: {"suffix": "dwi"}
bids_filters = "{'suffix': 'dwi'}"

# The name of the pipeline used to preprocess the DWI data.
# Default: "all".
preproc_pipeline = "qsiprep"

# list of participant labels (subject IDs) to perform
# processing on. If None, all subjects are used.
# Default: None
participant_labels = ""

# path to output directory. If None, outputs are put
# in a AFQ pipeline folder in the derivatives folder of
# the BIDS directory. pyAFQ will use existing derivatives
# from the output directory if they exist, instead of recalculating
# them (this means you need to clear the output folder if you want
# to recalculate a derivative).
# Default: None
output_dir = ""

# BIDS filters for inputing a user made tractography file.
# If None, tractography will be performed automatically.
# Default: None
custom_tractography_bids_filters = ""

[DATA]

# The value of b under which
# it is considered to be b0. Default: 50.
b0_threshold = 50

# Whether to use robust_tensor_fitting when
# doing dti. Only applies to dti.
# Default: False
robust_tensor_fitting = false

# Minimum b value you want to use
[...]

PR #768, opened 2021-11-10
[FIX] this is a quick fix for the problem with using the default config file

### LINKED ISSUE #767: Trouble with pyAFQ CLI
This seems related to issue 763.

I have tried running pyAFQ with two different config.toml files (config_emu-AFQ.toml, config_emu-UNC.toml), where I was previously successful with config_emu-AFQ.toml. Installed (version 0.9) via pip in a new conda environment (python=3.8). I have the same error in Mac OSx, Ubuntu 20.04, and RHEL 7. Running via `$ pyAFQ config.toml --notrack |& tee log`.

I will attach the file tree, config files (txt versions for github), and captured stdout/err (log). Additionally, this is resolved via fix.py (txt here), so I'm pretty confident it is not my config.toml file, but the way an (empty) dictionary is passed between modules?

Finally, I would love an entrypoint or some guidance on the docker image, and to be able to convert the image to a singularity for CLI usage.

Big fan of your work!

Config files:
[config_emu-AFQ.txt](https://github.com/yeatmanlab/pyAFQ/files/7499476/config_emu-AFQ.txt)
[config_emu-UNC.txt](https://github.com/yeatmanlab/pyAFQ/files/7499484/config_emu-UNC.txt)

Log files:
[log_emu-AFQ.txt](https://github.com/yeatmanlab/pyAFQ/files/7499498/log_emu-AFQ.txt)
[log_emu-UNC.txt](https://github.com/yeatmanlab/pyAFQ/files/7499499/log_emu-UNC.txt)

Fix files:
[fix.txt](https://github.com/yeatmanlab/pyAFQ/files/7499503/fix.txt)
[log_fix.txt](https://github.com/yeatmanlab/pyAFQ/files/7499504/log_fix.txt)

Tree:
Screen Shot 2021-11-08 at 2 03 48 PM

36000:
Fixes #767 and #763 . I think this whole system will be redesigned in #764 anyways. The problem here is that the default config file was being populated with the default params from each function's documentation, but in in the seed mask case, that default is None. We want by default to replace the `None` arguments with proper seed mask definitions in this case. If the user doesn't want to mask, they should use FullMask

### DIFF
--- a/AFQ/api.py
+++ b/AFQ/api.py
@@ -804,12 +804,6 @@ class AFQ(object):
 
         best_scalar = self._get_best_scalar()
         default_tracking_params = get_default_args(aft.track)
-        default_tracking_params["seed_mask"] = ScalarMask(
-            best_scalar)
-        default_tracking_params["stop_mask"] = ScalarMask(
-            best_scalar)
-        default_tracking_params["seed_threshold"] = 0.2
-        default_tracking_params["stop_threshold"] = 0.2
         # Replace the defaults only for kwargs for which a non-default value
         # was given:
         if tracking_params is not None:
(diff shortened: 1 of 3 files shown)


arokem:
Any chance we could put in place a test (nightly, possibly) that fails without this fix? Or does #764 obviate all this? Either way, we should probably put out a micro-release with this fix, even before #764 is ready to merge.

36000:
Yes, I can put it in an existing nightly test. Do you understand this current iteration error?

arokem:
Not sure, but there's a new release out for pybids (released yesterday), so
that might be at the root of this.

Are you seeing this locally?

36000:
It pass locally, I updated pyBIDS, then it did not pass

arokem:
OK. If this passes, we should merge this PR, make a new 0.9.1 (maybe after also merging #769?), and then proceed to debug this (possibly raising an issue in pybids, if it's on their end). I don't think that we'd like to pin pybids in the long run.

PR #769, opened 2021-11-10
[ENH] Add entrypoint for pyAFQ docker image

36000:
This makes the image more useful, for example in working with singularity.
Example usage:
```
docker run -v bids_dir:/bids_dir:rw ghcr.io/nrdg/pyafq bids_dir/config.toml
```

### DIFF
--- a/docs/source/usage/docker.rst
+++ b/docs/source/usage/docker.rst
@@ -1,0 +1,11 @@
+The pyAFQ docker image
+~~~~~~~~~~~~~~~~~~~~~~
+
+Everytime a new commit is made to master the
+`pyAFQ github <https://github.com/yeatmanlab/pyAFQ>`,
+a new image is pushed to the
+`NRDG github <https://github.com/orgs/nrdg/packages/container/package/pyafq>`.
+This image contains an installation latest version of the latest version of
+pyAFQ with fslpy and h5py. This image also contains an entrypoint, and can be
+run with::
+    docker run -v bids_dir:/bids_dir:rw ghcr.io/nrdg/pyafq bids_dir/config.toml
--- a/docs/source/usage/docker.rst
+++ b/docs/source/usage/docker.rst
@@ -5,7 +5,7 @@ Everytime a new commit is made to master the
 `pyAFQ github <https://github.com/yeatmanlab/pyAFQ>`,
 a new image is pushed to the
 `NRDG github <https://github.com/orgs/nrdg/packages/container/package/pyafq>`.
-This image contains an installation latest version of the latest version of
+This image contains an installation of the latest version of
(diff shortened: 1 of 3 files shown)


arokem:
Simple! Could you please add some documentation about this functionality?

docs/source/usage/docker.rst
```diff
@@ -0,0 +1,11 @@
+The pyAFQ docker image
+~~~~~~~~~~~~~~~~~~~~~~
+
+Everytime a new commit is made to master the
+`pyAFQ github <https://github.com/yeatmanlab/pyAFQ>`,
+a new image is pushed to the
+`NRDG github <https://github.com/orgs/nrdg/packages/container/package/pyafq>`.
+This image contains an installation latest version of the latest version of
```
arokem, line 8:
```suggestion
This image contains an installation of the latest version of
```

arokem (commented):
Documentation looks good. Most users might want to use (and keep fixed) a particular tag of the image, so it might be worth showing how that's done.

Maybe unrelated to this PR, I noticed that we still don't have version tags [here](https://github.com/orgs/nrdg/packages/container/package/pyafq). That is, it would be great to be able to run: 

```
docker run -v bids_dir:/bids_dir:rw ghcr.io/nrdg/pyafq:0.9 bids_dir/config.toml
```

And know that you are running version 0.9 of the software. I thought that we had set up our docker builds to pull in tags, but that doesn't seem to be working.

36000:
I agree and I am not sure why the versions are not working.

arokem:
Looking back at the GH Actions run for when we pushed the 0.9 tag, it looks like only the documentation build action was executed, but not the docker builds. Maybe we need to change [this](https://github.com/yeatmanlab/pyAFQ/blob/master/.github/workflows/docker_pyafq.yml#L3-L9) to be more like [this](https://github.com/yeatmanlab/pyAFQ/blob/master/.github/workflows/docbuild.yml#L3)?

36000:
Maybe just: `on: [pull_request]`
We don't want it pushing an image for every commit to non-master branch, will `on: [push]` do that? I am not sure

arokem:
I wouldn't mind running `docker build` with every commit.

Maybe we should move the `docker push` into the workflow itself and only
have that happen when tags are pushed? Basically the same logic as in the
documentation build workflow.

36000:
the docker build would just be to test that it builds? Sounds good. I will make a PR

docs/source/usage/docker.rst
arokem: This is still a typo, I think!

arokem, line 8:
```suggestion
This image contains an installation of the latest version of
```

PR #770, opened 2021-11-10
[ENH] Update the docker worklow to hopefully push tags, as well as remove AFQsi docker

36000:


36000:
@arokem is this what you had in mind for a new yaml? I also deleted afqsi docker as I don't think it will be needed soon.

arokem:
Yes. I think so. Let's see how this works with 0.9.1 🤞

arokem:
And +1 to removing afqsi, which should soon be superseded by the qsiprep docker image.

arokem:
Whoops. I think that I merged this prematurely. Is it possible that you didn't add the 'docker-push.sh' script? No hurry, of course - I am only looking at this now because my kiddos just went to sleep.

PR #772, opened 2021-11-12
[FIX] Remember to add docker-push.sh

36000:


arokem:
:-)

arokem:
Looks like it's working as intended now. It pushed a new tagged docker image with the merge of this PR. Now we'll need to make a tag/release and see whether that also gets us a tagged docker image.

PR #775, opened 2021-11-15
[ENH] try to get the tag name even not on release

36000:


arokem:
Looking [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) would `GITHUB_REF_NAME` work? That is described as the "The branch or tag name that triggered the workflow run". Would it include uwanted cruft?

arokem:
Should we go ahead and merge this?

36000:
GITHUB_REF_NAME might work, but this is tested to work as well. I think we can merge this.

arokem:
Should I push a new tag to try it out?

arokem:
I will go ahead and try

arokem:
It works! https://github.com/orgs/nrdg/packages/container/pyafq/10528786?tag=0.9.2

PR #776, opened 2021-11-22
NF: Handle situations where CSD auto-response function cannot complete.

arokem:


36000:
Should we merge #774  and then rebase this on that, so that Sendy can test this?

AFQ/models/csd.py
```diff
@@ -19,6 +19,9 @@
 __all__ = ["fit_csd"]
 
 
+class CsdNanResponseError(Exception):
+
```
36000, line 23: You need a `    pass` here

arokem:
If #774 is ready to go, then yes. But it looks like that one is waiting on #764 to be merged?

arokem:
Looks like this is not fundamentally broken. @SendyCaffarra : do you want to pull this one down onto the cluster and run it? It should raise a (relatively) informative error for those cases where you've been getting CSD-related errors.

arokem:
Rebased on top of current master. Do we know if it does what it's supposed to? 

@SendyCaffarra : did you get these errors happening again? Were you able to figure out which subject they came from using this error?

SendyCaffarra:
Yes!
This time I got informative errors and I could detect the problematic subject.

arokem:
Great! @36000 : should we merge this and then cut a release early next week?

PR #777, opened 2021-11-25
BF: Pediatric bundle uid 

arokem:
The first commit fixes a big in the initialization of the pediatric bundle dict, whereby the UIDs assigned to these bundles start at the number at which the parent class stopped counting in its initializer. This resets this count (and is consistent with the data we've analyzed from dhcp, as I just verified).


36000:
I wonder if `self._c_uid = 1` would be necessary if you initialized the BundleDict with no bundles, then added them manually.

arokem:
This was made obsolete through #788

PR #785, opened 2021-12-10
[FIX] use plotly cmap instead of matplotlib cmap in plotly_backend

36000:


36000:
I think this new colorscale also looks a little better
image

arokem:
I think this kind of data is best visualized with a sequential (not
diverging) colormap. I think that "viridis" would work well here.

36000:
I was basing it off of the mAFQ implementation
image

arokem:
No one should ever use that colormap (jet) for anything: https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/

36000:
Here is viridis:
image
This plot from the "jet" rant in particular is very interesting:
image

36000:
newplot
Here are all of plotly's built in sequential colors, as well

arokem:
I think viridis looks great for this!

PR #794, opened 2022-02-09
[FIX] Some fixes for nightly tests

36000:
Fixes two error:
1. Simple typo within test, introduced by recent bundle dict PR
2. Add a new Error to the try/catch for cvx.Problem().solve(), because of recent cvxpy update


36000:
@arokem this should fix two of the nightly tests, and will be ready to merge after the docs finish

PR #801, opened 2022-02-18
[FIX] fix CLI, % sign in scraped docstring causes error in argparser for CLI

36000:
This should fix the CLI, however I wonder if this should be documented somewhere


36000:
Closes #797

arokem:
Good catch! Maybe this should be documented somewhere in here? https://github.com/yeatmanlab/pyAFQ/blob/master/docs/source/developing/tasks.rst

PR #803, opened 2022-02-24
Remove MSMT from pyAFQ

36000:
Also attempts to fix a ValueError thrown in the nightly tests here: https://github.com/yeatmanlab/pyAFQ/runs/5300191283?check_suite_focus=true


36000:
@arokem  this is ready to merge. The documentation runs seem to be taking longer and longer, even without major changes to code relevant to them. I wonder why?

arokem:
Thanks! Let's hope that the documentation slow-downs are just a temporary thing.

PR #804, opened 2022-02-25
[ENH] Allow segmentation tie breakers to be settled by distance from ROI

36000:
This may be useful for sub-bundles.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -813,8 +813,8 @@ class Segmentation:
             if self.clip_edges:
                 self.logger.info("Clipping Streamlines by ROI")
                 for idx in range(len(select_sl)):
-                    min0 = min_dist_coords_bundle[idx, bundle_idx, 0]
-                    min1 = min_dist_coords_bundle[idx, bundle_idx, 1]
+                    min0 = int(min_dist_coords_bundle[idx, bundle_idx, 0])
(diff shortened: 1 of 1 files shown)


AFQ/segmentation.py
```diff
@@ -563,10 +570,13 @@ def segment_afq(self, tg=None):
         fgarray = np.array(_resample_tg(tg, 100))
         n_streamlines = fgarray.shape[0]
 
-        streamlines_in_bundles = np.zeros(
+        streamlines_in_bundles = np.empty(
```
arokem, line 573: Should this be a boolean array set to all-False on initialization instead?

arokem (commented):
Looks good! I wonder whether we can get away with using a boolean instead of a float array for these intermediate variables? I might be missing something though.

AFQ/segmentation.py
36000: This name I think is confusing and maybe should be changed. It really stores the probability of being in that bundle or 0/np.nan if not in the bundle. So it can have multiple bundle probabilities per streamline. It is only collapsed down to a boolean array on line 700: `possible_fibers = np.nansum(streamlines_in_bundles, -1) > 0`

arokem:
Is this ready to be merged? Have you had a chance to try it out with a real use-case?

36000:
Yes I tried it out with the real use-case, its good to go.

PR #812, opened 2022-04-08
Reduce memory consumption of Reco80 test

36000:
This test passes locally but is causing memory errors in the nightly's


arokem:
This makes sense. Looks like this is not messing up the tests, so I am going to merge this and we can 🤞 that the nightly passes.

PR #813, opened 2022-04-10
Reduce number of streamlines in nightly reco80 test

36000:
Passes test locally.


arokem:
Looks like the nightly tests are now passing! And master seems to be passing as well. 

Should we cut a release? Maybe not quite 1.0 yet, but tag a 0.11?

PR #814, opened 2022-04-20
Update DIPY to 1.5

36000:


36000:
This is failing because the newest version of DIPY no longer has min_length for tractography. Why is this and should we also remove this feature?

arokem:
Aren't we overloading this in our implementation of the tracker here? https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/_fixes.py#L71

36000:
@arokem this is ready to merge

PR #816, opened 2022-04-20
[ENH] Autogenerate two separate QSIprep pipelines

36000:
This will be used to update QSIprep more easily in the future.


36000:
@arokem  this is ready to merge

PR #826, opened 2022-05-11
[ENH] Change pyAFQ to use python >=3.8

36000:


arokem:
Looks like tests are passing on both versions of Python, so I will go ahead and merge this.

Issue #824, opened 2022-05-11
AttributeError for custom scalars


jdias001:
I am currently working with dwi data that has been preprocessed in pydesigner and would like to use the scalars output by pydesigner in pyafq. However, when I format the ScalarFile objects for my scalars, giving them a name and the path to the file, I keep getting an AttributeError: 'ScalarFile' object has no attribute 'get_for_subses'. Is there a solution to this problem? My code and the error I receive are below.

    import os
    from glob import glob
    import shutil
    import re
    
    from AFQ.api.participant import ParticipantAFQ
    from AFQ.definitions.scalar import ScalarFile
    
    BIDSfolder = "C:\\Users\\james\\Documents\\DKI\\pyAFQ\\BIDS\\dmriprep\\"
    infolder = "C:\\Users\\james\\Documents\\DKI\\pyAFQ\\BIDS\\derivatives\\pydesigner\\"
    outfolder = "C:\\Users\\james\\Documents\\DKI\\pyAFQ\\BIDS\\derivatives\\afq\\"
    
    dwi_data_files = glob(infolder+"\\**\\ses-*\\*dwi*preprocessed.nii.gz")
    bval_files = glob(infolder+"\\**\\ses-*\\*preprocessed.bval")
    bvec_files = glob(infolder+"\\**\\ses-*\\*preprocessed.bvec")
    
    dki_ak_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dki_ak*.nii.gz")
    dki_kfa_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dki_kfa*.nii.gz")
    dki_mk_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dki_mk*.nii.gz")
    dki_rk_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dki_rk*.nii.gz")
    dti_ad_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dti_ad*.nii.gz")
    dti_fa_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dti_fa*.nii.gz")
    dti_md_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dti_md*.nii.gz")
    dti_rd_scalar_files = glob(infolder+"\\**\\ses-*\\metrics\\*dti_rd*.nii.gz")
    
    output_dirs = []
[...]

PR #827, opened 2022-05-11
[FIX] Replace split(.) with splitext()

36000:
Once this is merged, we should consider a minor release as this is needed to work with QSIprep

### DIFF
--- a/AFQ/utils/path.py
+++ b/AFQ/utils/path.py
@@ -1,4 +1,6 @@
+import os.path as op
+
+
 def drop_extension(path):
-    base_fname = path.split('/')[-1].split('.')[0]
-    path.split(base_fname)[0] + base_fname
-    return path
+    base_fname = op.basename(path).split('.')[0]
+    return path.split(base_fname)[0] + base_fname
--- a/AFQ/utils/path.py
+++ b/AFQ/utils/path.py
@@ -1,0 +1,4 @@
+def drop_extension(path):
+    base_fname = path.split('/')[-1].split('.')[0]
+    path.split(base_fname)[0] + base_fname
+    return path
--- a/AFQ/utils/tests/test_path.py
+++ b/AFQ/utils/tests/test_path.py
@@ -1,5 +1,6 @@
 from AFQ.utils.path import drop_extension
 
+
 def test_drop_extension():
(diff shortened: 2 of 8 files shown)


arokem:
Could you please rebase this on top of the now-merged #826?

AFQ/utils/path.py
```diff
@@ -0,0 +1,4 @@
+def drop_extension(path):
```
arokem, line 1: This will fail quite horribly on Windows machines (which our previous implementation probably also did...). Maybe use https://docs.python.org/3/library/pathlib.html?

arokem (commented):
We might want to take the opportunity to do something more OS-portable.

AFQ/utils/path.py
arokem: And maybe add some tests with Windows style paths? https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

AFQ/utils/tests/test_path.py
```diff
@@ -0,0 +1,9 @@
+from AFQ.utils.path import drop_extension
+
+
+def test_drop_extension():
+    assert "/my/.example/.path/to_file" == drop_extension(
+        "/my/.example/.path/to_file.nii.gz")
+    
+    assert "C:\Documents\Newsletters\Summer2018" == drop_extension(
```
arokem: I bet that the current test failures have to do with the fact that the webpage you copied this from was using some funky character set. Might work if you write it out by hand. Oh, Microsoft.

AFQ/utils/path.py
arokem: BTW: do you think this helps explain #824 ?

AFQ/utils/tests/test_path.py
36000: I think it was just that I needed to escape the backslash characters!

arokem: Oh - right. Good call.

36000:
@arokem  i think this is ready to merge, and we can rebase #825 on this

AFQ/utils/path.py
36000: I think that is different

PR #832, opened 2022-05-13
[ENH] Ignore pandas out of date warning / SFT reference warning

36000:


arokem:
Thanks for doing this. These warnings are filling my screen up! 😄 I think that we can merge this as soon as a test comes back green.

PR #839, opened 2022-05-25
[ENH] Return path to single viz file if its generated

36000:
useful for qsiprep

### DIFF
--- a/AFQ/tasks/viz.py
+++ b/AFQ/tasks/viz.py
@@ -66,6 +66,7 @@ def viz_bundles(base_fname,
 
     Returns
     -------
+    Tuple of Figure, String or just the Figure:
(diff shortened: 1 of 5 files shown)


AFQ/tasks/viz.py
```diff
@@ -45,7 +45,8 @@ def viz_bundles(base_fname,
                 n_points_bundles=40):
     """
     figure for the visualizaion of the recognized
-    bundles in the subject's brain
+    bundles in the subject's brain. Returns path
+    if file can be generated.
```
arokem, line 49: This should be in a "Returns" section:

```
Returns 
--------
String or None: if file can be generated, returns the full path to this file.

Issue #430, opened 2020-09-15
Add VOF segmentation


arokem:
There are several different approaches here: 

1. Use Freesurfer-based ROIs. This is the currently-implemented version in Matlab AFQ (@grotheer: where is this function?). 
2. Uses one waypoint ROI and a geometric constraint on the streamlines (have to be approximately horizontal). This is not currently implemented, but @grotheer says she can work on generating the waypoint ROI.
3. There is also an approach implemented here: https://github.com/yeatmanlab/AFQ/blob/master/functions/AFQ_Segment_VOF.m, that uses two waypoint ROIs. Is this approach good? This is a question for @jyeatman + @grotheer. This approach has two waypoint ROIs that both have a 'L' prefix and I don't see any R_VOF ROIs. Does that mean that they only segment the left hemisphere? Or are they somehow reflected into the right hemisphere. I don't see that in the code. Maybe they cover both hemispheres?

jyeatman:
Approach 3 should be deprecated. Here is the relevant matlab function https://github.com/yeatmanlab/AFQ/blob/master/vof/AFQ_FindVOF.m

arokem:
@jyeatman : thanks! IIUC, this is the one that implements approach 1?

jyeatman:
yes

PR #840, opened 2022-06-02
[WIP] [ENH] Add custom bundle recognizer

36000:
We can use this during a code review for adding the VOF.
This are the criterion for VOF:

1. Fibers have one endpoint on the ventral surface. This can be done by taking the regions of the atlas that are on the ventral surface
2. Fibers have a primary orientation in the Z axis. We could write a simple function to compute each fibers primary orientation
3. fibers are posterior to the arcuate fascicules

This are addressed: 1) using endpoint interface, 3) either using prob map or exclusion ROI based on fascicules prob map, and 2) using this PR

See #430

### DIFF
--- a/AFQ/api/custom_recognizer.py
+++ b/AFQ/api/custom_recognizer.py
@@ -1,0 +1,1 @@
+# put custom recognition functions here
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -810,6 +810,10 @@ class Segmentation:
                 self.logger.info("After filtering "
                                  f"{len(select_sl)} streamlines")
 
+            if "custom_recognizer" in self.bundle_dict[bundle]:
+                select_sl = self.bundle_dict[bundle]["custom_recognizer"](
+                    select_sl, self.img)
+
             if self.clip_edges:
                 self.logger.info("Clipping Streamlines by ROI")
                 for idx in range(len(select_sl)):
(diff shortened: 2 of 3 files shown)


jyeatman:
1) How does the order of operations work with the basic segmantation - Does this operate on all fibers or only on those that are unclassified? 
2) Follow up question: for the custom bundle recognizer, imagine a scenario where there are two potential bundles that a fiber might belong to. Is segmentation always done in sequence or is there a way to compute the probability for 2 separate bundles and assign to the one with the highest probability?

36000:
For (2) , tie breakers are done with probability maps which the user can input. Optionally distance to ROI can be used as the tie breaker instead. We can add more options if needed.
 For (1), first streamlines are selected using the ROIs and probability map threshold, then tie-breaked, then filtered by endpoints, then this new custom recognizer is called. 
 
 I think that for VOF, first we select using ROIs, then the the probability map can make sure that the fibers are in the right place relative to the arcuate. Then the custom recognizer is called and it will remove any streamlines whose primary orientation is not in the z axis. However if needed I could move the custom recognizer earlier in the process.
 
 BTW, I am open to changes for the name "custom recognizer". Here is the docstring:
 ```
 function which accepts two arguments: (1) array of
      streamlines of shape Kx3xN where K is the number of found streamlines, 
      and N is the number of points in each streamline (default is 100), and
      (2) a Nifti1Image for reference. Should return an array of streamlines
      of shape Lx3xN, where L is the number of accepted streamlines. This
      function is called near the end of bundle recognition, after cleaning
      by endpoints but before the optional ROI clipping.
 ```

Issue #843, opened 2022-06-07
Custom Scalars Error


cornelius-eichner:
Dear AFQ developers,

We are currently running pyAFQ (python 3.8.8) in developmental data and would like to map custom tissue properties (qT1) to the reconstructed fibers. We followed the description on this website: https://yeatmanlab.github.io/pyAFQ/usage/scalars.html

When we try to incorporate the qT1 in the analysis pipeline, we get an error (attached):

The data are organized in a BIDS structure and the standard AFQ pipeline for FA and MD runs smoothly (AFQ config attached). 

We tested and replicated this behavior in pyAFQ v0.9, v0.11, and v0.12. 

Is there something we can do to circumvent this error? Thank you in advance for your help! 

Best regards, Cornelius

AFQ Config:
~~~
	r1_scalar = ImageFile(
		suffix="R1_aligned",
		filters={"scope": "qmri"})

	# combine custom ROIs with default BundleDict ROIs
	#bundles = bundles + abd.BundleDict()

	tracking_params = dict(odf_model='CSD', directions='prob', n_seeds=3, step_size=0.5) 
	segmentation_params = dict(seg_algo='afq', prob_threshold=3)
	clean_params = dict(n_points=50, distance_threshold=3,min_sl=5)
	parallel_params = dict(n_jobs=8, engine='joblib', backend='loky')

	my_afq = GroupAFQ(
		bids_path,
		dmriprep='prepro',
		bundle_info = bundles,
		parallel_params = parallel_params,
		scalars=["dti_fa", "dti_md", "dti_rd", "dti_ad", r1_scalar],
		segmentation_params=segmentation_params,
		tracking_params=tracking_params,
		clean_params=clean_params)
	
	my_afq.export_all()

~~~

AFQ Error: 
~~~
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-663a074a9132> in <module>
----> 1 afq_config_synbrain_parallel_test_R1.afq_bids_run_all()

/data/pt_02568/software/anaconda3/lib/python3.8/site-packages/AFQ/afq_config_synbrain_parallel_test_R1.py in afq_bids_run_all()
     93         parallel_params = dict(n_jobs=8, engine='joblib', backend='loky')
     94 
---> 95 	my_afq = GroupAFQ(
     96                 bids_path,
     97                 dmriprep='prepro',

[...]

PR #844, opened 2022-06-07
[FIX] Error when file not found should be more informative

36000:


36000:
Should make issues like #843 easier to debug.

arokem:
The fix looks good to me! Do you want to add a test that this is the error that gets raised under circumstances similar to those observed in #843 ?

36000:
added test to nightly, passes locally

PR #851, opened 2022-07-05
[DOC] add fatal error when no bundles found at all

36000:


36000:
@arokem  ready for review / merge? This is useful because it would trigger in tests if there is a bug somewhere causing nothing to be found, when otherwise the tests would fail silently if nothing is found.

arokem:
Is there any way to test that this error gets raised for some obvious case?

36000:
yes, theres a few things I want to change in this PR now

PR #852, opened 2022-07-05
[ENH] add median bundle len function

36000:


36000:
@arokem  ready for review / merge

AFQ/tasks/segmentation.py
```diff
@@ -214,6 +214,46 @@ def export_sl_counts(data_imap,
...
+
+        for bundle in bundle_names:
+            if bundle == "whole_brain":
+                lens.append(np.median(
+                    seg_sft.sft._tractogram._streamlines._lengths))
+            else:
+                lens.append(np.median(
+                    seg_sft.get_bundle(
+                        bundle)._tractogram._streamlines._lengths))
+            if np.isnan(lens[-1]):
```
arokem, line 244: Where would a nan come from in this situation? Would it make sense to use a `nanmedian` on L238 and L241?

36000: No, the Nan only occurs on empty lengths list because of empty bundle. I should move this if statement further up to deal with this directly

AFQ/tasks/segmentation.py
```diff
@@ -214,6 +214,46 @@ def export_sl_counts(data_imap,
     return counts_df, dict(sources=bundles_files)
 
 
+@pimms.calc("sl_lengths")
+@as_file('_sl_lengths.csv', include_track=True, include_seg=True)
+def export_sl_lengths(data_imap,
```
arokem, line 219: Just one more nit here: do you want the name of the function to reflect the fact that these are *median* counts? Right now, it gives the impression that these are somehow "raw" counts.

Issue #819, opened 2022-04-27
Remove underscores in file names


36000:
With names like: `sub-100206_dwi_space-RASMM_model-DTI_desc-det-AFQ-clean_tractography.trk`
The suffix is only tractography. We should change this so there is no underscore or space or dash, so the suffix is more descriptive. Maybe we should use camelCase?

I would be interested in hearing your thoughts, @arokem ?

arokem:
I believe that we came up with this (addmitedly long) names based on our understanding of the draft of the BIDS derivatives spec for tractography results. That's where that `_tractography` suffix comes from. camelCase is definitely not the BIDS way 😄

Issue #799, opened 2022-02-16
Less space on disk


36000:
Something like myafq.cmd_streamlines("mv", "path") should move streamlines and everything that depends on them to path. 

Edit: maybe all commands should be like this, ie. export_profiles and myafq.profiles should be myafq.export("profiles")

36000:
This was done!

Issue #795, opened 2022-02-09
Pyafq trim method


36000:
There should be a pyAFQ trim method which deletes all but necessary derivatives (subjective, determined by arguments). This would be useful for running pyAFQ on large datasets.

arokem:
Slight quibble is that I would not call this "trim", because I start thinking about trimming the streamlines. Maybe "clobber", following the Cloudknot API's use of this word?

richford:
Or maybe "clean." But now we're really bike-shedding.

Also, can I request that this uses `fsspec` so that it works on local files as well as S3?

PR #853, opened 2022-07-11
[ENH] Apply arbitrary command to some/all pyAFQ outputs, more BIDSy names

36000:
This PR has two parts, both affect output file names.
1. Propose a framework for easily handling pyAFQ outputs according to what they dependent. In this case, we use a depends field we put in the json files. Also, you can exclude particular outputs if you want. This is, I think, a more elegant solution to #799 and #795
2. Make pyAFQ outputs follow a more BIDS-y format. Discussed previously in #819

### DIFF
--- a/AFQ/tasks/segmentation.py
+++ b/AFQ/tasks/segmentation.py
@@ -83,7 +83,7 @@ def segment(dwi, data_imap, mapping_imap,
 
 
 @pimms.calc("clean_bundles")
-@as_file('-cleanTractography.trk', include_track=True, include_seg=True)
+@as_file('_desc-clean_tractography.trk', include_track=True, include_seg=True)
 def clean_bundles(bundles, data_imap, clean_params=None):
     """
     full path to a trk file containting segmented
--- a/AFQ/tasks/segmentation.py
+++ b/AFQ/tasks/segmentation.py
@@ -9,7 +9,7 @@ import logging
 import pimms
 
 from AFQ.tasks.decorators import as_file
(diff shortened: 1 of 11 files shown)


36000:
@arokem  this is ready for review. I moved the dependent information to the json file.

AFQ/tasks/segmentation.py
```diff
@@ -83,7 +83,7 @@ def segment(dwi, data_imap, mapping_imap,
 
 
 @pimms.calc("clean_bundles")
-@as_file('-clean_tractography.trk', include_track=True, include_seg=True)
+@as_file('-cleanTractography.trk', include_track=True, include_seg=True)
```
arokem, line 86: `_tractography` will eventually be a recognized BIDS suffix, so I think this should be OK without the change. Same is true for `_xform`.

arokem:
I think that we should adhere to suffixes that are already or will eventually be part of the BIDS standard: `_mask`, `_xform`, `_tractography`. The problem is then what to do with `clean`. I think we can make that part of the `desc` portion of the filename, which can maybe be camel-cased. WDYT?

36000:
Putting in the desc makes sense to me!

36000:
Can now address #795  as well.

36000:
@arokem  I updated the top comment so this is ready to review / merge now.

arokem:
We should follow up at some point add some documentation explaining how to "trim" and rerun pyAFQ while changing parameters (I think that's probably a typical use-case).

PR #854, opened 2022-07-14
BF: Need to import ItkMap as part of __all__.

arokem:


arokem:
Duplicates some of #855

PR #855, opened 2022-07-14
[FIX] Some ITK map fixes for running with groupAFQ and in CLI

36000:


arokem:
Yep! I was just looking at the same LOC. Will merge and try this locally.

PR #858, opened 2022-07-18
FIX: Reorder endpoint

36000:
@arokem  this PR puts the endpoint segmentation before the tie breaking, and works. I also made some changes to the API of the clean endpoints function, and I label this as WIP because I have a few ideas about rewriting the segmentation module based on how we currently use it.

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -763,12 +763,20 @@ class Segmentation:
 
         # Eliminate any fibers not selected using the waypoint ROIs:
         possible_fibers = np.nansum(streamlines_in_bundles, -1) > 0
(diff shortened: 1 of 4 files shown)


36000:
@arokem  this works as intended, ready for review, and we can discuss tomorrow. I will restructure segmentation module in a different PR.

36000:
Push up endpoint ROIs to be first thing to do (should be commutative).

36000:
TODO: Push up the endpoint ROIs above include / exclude.

36000:
Such that :
prob map -> endpoint -> waypoint include / exclude -> tie break

put up warning for number of streamlines with more than one possible bundle ID, most likely chosen using tie breaking method.

PR #862, opened 2022-08-02
[FIX] only calc subject registration images when necessary

36000:
Fixes a few random things from what I have seen in NKI run. CSD APM calculated even though it is not used. b0_transform calculated even though our ITK map imported does not do subject->template transformation. Does not fix the ITK mapping problem though, as the template_xform also shows the mapping problem.


36000:
@arokem

PR #863, opened 2022-08-03
ENH: add lower triangular scalars from DTI (could be useful for ML)?

36000:
These can already be calculated from our derivatives, but now they are formally added to the task system. 
Also adds a change for tract profile calculation that should significantly speed it up.

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -280,6 +280,24 @@ def dti_fa(dti_tf):
     return dti_tf.fa
 
 
+@pimms.calc("dti_lt0", "dti_lt1", "dti_lt2", "dti_lt3", "dti_lt4", "dti_lt5")
+def dti_lt(dti_tf, dwi_affine):
+    """
+    Image of first element in the DTI tensor,
+    Image of second element in the DTI tensor,
+    Image of third element in the DTI tensor,
+    Image of fourth element in the DTI tensor,
+    Image of fifth element in the DTI tensor,
+    Image of sixth element in the DTI tensor
+    """
+    dti_lt_dict = {}
+    for ii in range(6):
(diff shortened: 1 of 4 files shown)


AFQ/tasks/segmentation.py
```diff
@@ -304,11 +304,16 @@ def tract_profiles(clean_bundles, data_imap,
         this_sl = seg_sft.get_bundle(bundle_name).streamlines
         if len(this_sl) == 0:
             continue
+        if profile_weights == "gauss":
+            # calculate only once per bundle
+            bundle_profile_weights = gaussian_weights(this_sl)
```
arokem, line 309: Good call!

AFQ/tasks/data.py
```diff
@@ -280,6 +280,24 @@ def dti_fa(dti_tf):
     return dti_tf.fa
 
 
+@pimms.calc("dti_lt0", "dti_lt1", "dti_lt2", "dti_lt3", "dti_lt4", "dti_lt5")
```
arokem, line 283: Do we want `dki_lt` elements as well? I believe there are 15 elements in DKI. 

And relatedly, we often calculate FA/MD from DKI, when that's doable (i.e., when there are multiple non-zero b-values). Do we want to be able to extract the six DTI elements from DKI?

36000:
@arokem this might be useful for the final NKI run.

arokem (commented):
Definitely an improvement! I have some questions about how we might incorporate DKI in here as well.

arokem:
> @arokem this might be useful for the final NKI run.

Agreed! And the DKI questions do not pertain to that, so we could merge this as-is for that, and then come back to deal with those questions later on.

36000:
@arokem  can we merge this and do DKI in a separate PR?

Issue #857, opened 2022-07-15
Logs suppressed


arokem:
In a run of pyAFQ that I did yesterday, I get a few warnings raised and lots of tqdm progress bars, but no logs otherwise: 

```
(afq) ➜  NKI pyAFQ ./afq_config.toml    
Loading AFQ libraries...
INFO:pyAFQ:Your use of pyAFQ is being recorded using Google Analytics. For more information, please refer to the pyAFQ documentation: https://yeatmanlab.github.io/pyAFQ/usage.html#usage-tracking-with-google-analytics. To turn this off, use the `--notrack` flag when using the pyAFQ CLI
100%|███████████████████████████████| 172104/172104.0 [00:48<00:00, 3532.95it/s]
100%|███████████████████████████████| 209744/209744.0 [01:01<00:00, 3393.21it/s]
100%|█████████████████████████████████| 164266/164266 [1:16:54<00:00, 35.60it/s]
WARNING:AFQ.api.seg:0 invalid streamlines removed
100%|███████████████████████████████████| 77070/77070 [00:13<00:00, 5528.98it/s]
100%|███████████████████████████████████| 75069/75069 [00:11<00:00, 6491.58it/s]
100%|███████████████████████████████████| 21284/21284 [00:06<00:00, 3065.88it/s]
100%|███████████████████████████████████| 19408/19408 [00:05<00:00, 3257.69it/s]
100%|██████████████████████████████████| 41860/41860 [00:04<00:00, 10412.55it/s]
100%|███████████████████████████████████| 34826/34826 [00:03<00:00, 9804.04it/s]
100%|███████████████████████████████████| 75223/75223 [00:31<00:00, 2380.06it/s]
100%|███████████████████████████████████| 77786/77786 [00:32<00:00, 2421.78it/s]
100%|███████████████████████████████████| 68736/68736 [00:59<00:00, 1159.66it/s]
100%|████████████████████████████████████| 73393/73393 [01:15<00:00, 978.39it/s]
100%|█████████████████████████████████| 100648/100648 [00:59<00:00, 1689.81it/s]
100%|███████████████████████████████████| 97667/97667 [00:51<00:00, 1889.75it/s]
100%|███████████████████████████████████| 80346/80346 [00:31<00:00, 2557.74it/s]
100%|███████████████████████████████████| 74406/74406 [00:29<00:00, 2550.86it/s]
[...]

PR #867, opened 2022-08-13
Set logger levels to INFO

arokem:
Particularly for saving of files, so that we get more information about what is being saved and when. Output is more verbose, but that should help with #857, where not a lot of information is provided except for the prog bars.

### DIFF
--- a/AFQ/api/bundle_dict.py
+++ b/AFQ/api/bundle_dict.py
@@ -168,7 +168,7 @@ class BundleDict(MutableMapping):
             for bundle_name in bundle_info:
                 self.bundle_names.append(bundle_name)
 
-        self.logger = logging.getLogger('AFQ.api')
+        self.logger = logging.getLogger('AFQ')
 
         if self.seg_algo == "afq":
             if "FP" in self.bundle_names\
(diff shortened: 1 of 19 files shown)


36000:
Should we do this in all files? Is there some way to set this universally or have the user do it?

36000:
@arokem  maybe we should replace things like
```
logger = logging.getLogger('AFQ.Segmentation')
logger.info()
````
with:
```
logging.info()
```
This is what dipy uses in most cases, however in some cases they do use getlogger (like stateful tractogram) . For the size of our package I think we may not need to have our own getLogger. What do you think?

arokem:
I don't have much experience with this. I think that using the `logging.info` function directly would be fine, so long as we can set the logging levels somewhere global. For example, in the package `__init__.py` file. And it should be possible to over-ride for users who want less verbose output everywhere.

36000:
Trying middle ground, setting everything to 'AFQ' logger. Does this work in the situations where we were having problems?

36000:
@arokem  if this also solves the problems you were seeing earlier, we can merge this

arokem:
OK. Let's merge this and adjust if we see this is too verbose or not verbose enough.

Issue #866, opened 2022-08-11
Not finding DWI when BIDS structure doesn't include sessions?


arokem:
I am seeing this warning:

https://github.com/yeatmanlab/pyAFQ/blob/8ac41183ebc0d684cb047b4f2280ae6bb73db36d/AFQ/api/group.py#L266

In the case of dataset that has just one session per subject, so the BIDS structure doesn't have explicit session designations. IIUC, this is the case in which `session` is `None`. I wonder if this throws off the bids filtering operation on https://github.com/yeatmanlab/pyAFQ/blob/8ac41183ebc0d684cb047b4f2280ae6bb73db36d/AFQ/api/group.py#L262

arokem:
It's actually stranger than that. Some of the subjects within the same dataset are found and others are not. Puzzling.

arokem:
A thing I learned about pybids today is that if you have a dataset that has a `derivatives` folder and in that derivatives folder some sub-folders have n subjects and some derivatives have m subjects, pybids can get confused and not find the data for subjects that are in one sub-folder but in the other sub-folders. So, one has to be careful with messy derivative folders and datasets that are analyzed as they are being collected. Either way, not a pyAFQ issue, so I am closing this.

PR #869, opened 2022-08-13
Upgrade pybids.

arokem:
I am opening this PR to explore what we need to change in order to be able to upgrade pybids to the more current version.

### DIFF
--- a/AFQ/data/fetch.py
+++ b/AFQ/data/fetch.py
@@ -20,7 +20,9 @@ import logging
 import time
 
 import bids.config as bids_config
+import warnings
 try:
+    warnings.simplefilter(action='ignore', category=FutureWarning)
     bids_config.set_option('extension_initial_dot', True)
 except ValueError:
     pass
@@ -28,7 +30,6 @@ import nibabel as nib
 import boto3
 
 # capture templateflow resource warning and log
-import warnings
 default_warning_format = warnings.formatwarning
 try:
     warnings.formatwarning = lambda msg, *args, **kwargs: f'{msg}'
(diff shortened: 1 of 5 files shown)


arokem:
I'm hoping this will help resolve the so far pretty hard-to-reproduce #866

arokem:
Looks like tests are passing with an upgrade to current pybids. However, it still doesn't get rid of the ever-present warning: 

```
WARNING:py.warnings:The ability to pass arguments to BIDSLayout that control indexing is likely to be removed in future; possibly as early as PyBIDS 0.14. This includes the `config_filename`, `ignore`, `force_index`, and `index_metadata` arguments. The recommended usage pattern is to initialize a new BIDSLayoutIndexer with these arguments, and pass it to the BIDSLayout via the `indexer` argument.
```

Which is pretty annoying, and printed many many times each run. I wonder what we'd need to change in our code in order to get rid of that one.

36000:
@arokem the logs look clean, this is ready for review / merge

AFQ/data/fetch.py
```diff
@@ -20,15 +20,16 @@
 import time
 
 import bids.config as bids_config
+import warnings
 try:
+    warnings.simplefilter(action='ignore', category=FutureWarning)
     bids_config.set_option('extension_initial_dot', True)
```
arokem, line 26: Remind me what `extension_initial_dot` does?

Do I understand correctly that this will stop working in pybids 0.16? Or is this config simply not needed anymore at that point?

36000:
@arokem you are correct about the extension_initial_dot, it was only necessary in much earlier versions, so I removed it

arokem:
Excellent! Off we go.

PR #870, opened 2022-08-16
Update to generate by for BIDS

36000:
this PR is held up until pybids fixes this: https://github.com/bids-standard/pybids/issues/864
Also do not merge this until it is rebased on top of #869

### DIFF
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -185,8 +185,9 @@ class GroupAFQ(object):
         pipeline_description = {
             "Name": bids_description["Name"],
             "BIDSVersion": bids_description["BIDSVersion"],
-            "PipelineDescription": {"Name": "pyAFQ",
-                                    "Version": pyafq_version}}
+            "GeneratedBy": [{
+                "Name": "pyAFQ",
+                "Version": pyafq_version}]}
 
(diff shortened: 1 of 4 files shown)


36000:
Not sure why but this still doesn't work....

36000:
@arokem could you create a minimal example to demonstrate whether this is a pybids problem?

arokem:
I created a gist that replicates the problem that we are seeing: https://gist.github.com/arokem/7505a75031924960fc9c18a88589900e

Along the way, I also found out that the datasets that we create in these tests do not pass BIDS validation, so I had to tweak the test data a little bit to add synthetic "raw" data (this is because you can't have a valid BIDS dataset that doesn't have subject folders at top level). 

Nevertheless, these datasets are now BIDS valid, but still when there is only a GeneratedBy field in the `dataset_description.json`, the call to `get` with the derivative scope returns an empty list, instead of the list populated with the file pointers to the files that live under that scope. I'll post an issue to the pybids GitHub as well.

arokem:
OK - I can confirm that this will be fixed upstream when https://github.com/bids-standard/pybids/issues/864 is merged. So I guess we keep this PR open until that appears in a pybids release version.

Issue #872, opened 2022-08-17
Error handling broken


arokem:
Looks like the recent merge (of #853, I assume) broke the error handling in this nightly test: https://github.com/yeatmanlab/pyAFQ/runs/7878789485?check_suite_focus=true#step:6:57

Issue #753, opened 2021-10-12
Broken nightly


arokem:
Looks like maybe #750 broke things? 

https://github.com/yeatmanlab/pyAFQ/runs/3867843788?check_suite_focus=true#step:6:153

36000:
May be fixed by #752

PR #873, opened 2022-08-17
Fix nightly tests

36000:
SLR test is also revamped to be much faster (a few minutes or less)


arokem:
Just making sure: this closes #753?

36000:
I think thats an old issue , this fixes #872 (and other broken nightly tests from the merge of #853 )

arokem:
Ha! Well, let's merge it and see what happens.

PR #874, opened 2022-08-18
FIX: nightly broken online, but passes locally

36000:
This is odd because this test passes locally but seems to fail on the nightly CircleCI. I assume that either the random seed is the issue or different versions, however our versions should be the same. So maybe adding `np.random.seed(42)` will work? Not sure.


36000:
It's also true that this feature is very much not used, I wonder if we should keep it in pyAFQ, especially considering we want to bring in ANTSpy?

arokem:
Addmitedly, this is not something we have been using a lot. But it is distinct from a potential switch to antspy, because this is about streamline registration, not image registration. I admit that it's not worth a lot of maintenance effort, but if we can make it pass and kick the decision can down the road, maybe that would be nice?

arokem:
Looks like this was superseded by #929

PR #880, opened 2022-08-24
Added streamline filtering by primary orientation

jyeatman:
Added a function to filter streamlines by their primary orientation. We now need to add that to the bundle dict as a criteria

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -28,6 +28,9 @@ from dipy.utils.parallel import paramap
 __all__ = ["Segmentation", "clean_bundle", "clean_by_endpoints"]
 
 
+logger = logging.getLogger('AFQ')
+
+
 def _resample_tg(tg, n_points):
     # reformat for dipy's set_number_of_points
     if isinstance(tg, np.ndarray):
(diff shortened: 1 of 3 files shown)


AFQ/segmentation.py
```diff
@@ -1183,6 +1183,23 @@ def _is_streamline_in_ROIs_parallel(indiv_args, tol, include_roi,
...
+    """
+    Compute the cardinal orientation of each streamline
+    Parmaeters
+    __________
+    streamlines : sequence of 3XN_i arrays
+    """
+
+    axis_diff = np.zeros((len(streamlines),3))
+    for ii, sl in enumerate(streamlines):
+        axis_diff[ii,:] = np.abs(sl[0,:] - sl[-1,:])
```
arokem, line 1197:
```suggestion
        axis_diff[ii, :] = np.abs(sl[0, :] - sl[-1, :])
```

arokem: PEP8

arokem (commented):
Very nice! Would be great to add a test.

arokem:
IIUC, this currently decides whether a streamline is vertical based on the start and end points. But you could imagine a streamline that curves around horizontally for most of its length and gets filtered here as vertical if its start and end points are slightly vertical.

36000:
Superseded by #898

PR #881, opened 2022-08-24
[FIX] cmd output only run on our files

36000:


arokem:
What happens otherwise?

36000:
can crash when it tries to , for example, find the json sidecar for ".DS_STORE" (or even just process that filename)

arokem:
Oh yeah. That's not good.

PR #898, opened 2022-08-26
Added streamline filtering by primary orientation; other bundle definition fixes

36000:


36000:
@arokem  ready for review

AFQ/api/bundle_dict.py
```diff
@@ -31,6 +31,8 @@ def append_l_r(bundle_list, no_lr_list):
            "FA", "FP"]
 BUNDLES = append_l_r(BUNDLES, ["FA", "FP"])
 
+READ_BUNDLES = append_l_r(["pARC", "VOF"], [])
```
arokem, line 34: Shouldn't we make this part of our standard set?

AFQ/api/group.py
```diff
@@ -511,6 +514,8 @@ def cmd_outputs(self, cmd="rm", dependent_on=None, exceptions=[]):
         for pAFQ in self.pAFQ_list:
             pAFQ.cmd_outputs(cmd, dependent_on, exceptions)
 
+    clobber = cmd_outputs
```
arokem, line 517: This is just an alias for the default?

arokem: If yes, a comment to that effect might be good to put here.

AFQ/api/participant.py
```diff
@@ -235,12 +237,22 @@ def cmd_outputs(self, cmd="rm", dependent_on=None, exceptions=[]):
...
             elif os.path.isdir(full_path):
                 # other than ROIs, folders are dependent on everything
                 if dependent_on is None or filename != "ROIs":
                     os.system(f"{cmd} -r {full_path}")
+
+        # do not assume previous calculations are still valid
+        # after file operations
+        self.make_workflow()
+
+    clobber = cmd_outputs
```
arokem, line 258: And same here.

AFQ/definitions/image.py
```diff
@@ -229,7 +229,12 @@ class RoiImage(ImageDefinition):
     api.GroupAFQ(tracking_params={"seed_image": seed_image})
     """
 
-    def __init__(self, use_presegment=False, use_endpoints=False):
+    def __init__(self,
+                 use_waypoints=True,
+                 use_presegment=False,
+                 use_endpoints=False):
+        # TODO: assert at least one is true
```
arokem: Still a TODO? To add in this PR? (I vote yes)

AFQ/tests/test_bundle_dict.py
```diff
@@ -52,13 +52,12 @@ def test_BundleDict():
...
-    # not included and does not exist in afq templates
+    # mispelled bundle that does not exist in afq templates
     with pytest.raises(
             ValueError,
-            match="VOF_L is not in AFQ templates"):
-        afq_bundles = abd.BundleDict(["VOF_L", "VOF_R"])
-        afq_bundles["VOF_R"]
+            match=" is not in AFQ templates"):
+        afq_bundles = abd.BundleDict(["VOQ_L", "VOQ_R"])
+        afq_bundles["VOQ_R"]
```
arokem, line 60: Nice

AFQ/tractography.py
```diff
@@ -21,7 +21,7 @@
 def track(params_file, directions="prob", max_angle=30., sphere=None,
           seed_mask=None, seed_threshold=0, n_seeds=1, random_seeds=False,
           rng_seed=None, stop_mask=None, stop_threshold=0, step_size=0.5,
-          min_length=10, max_length=1000, odf_model="CSD",
+          min_length=40, max_length=200, odf_model="CSD",
```
arokem, line 24: These are pretty dramatic changes! Can you say a bit more about this?

AFQ/segmentation.py
```diff
@@ -1184,6 +1243,43 @@ def _is_streamline_in_ROIs_parallel(indiv_args, tol, include_roi,
             streamlines_in_bundles)
 
 
+def clean_by_orientation(streamlines, primary_axis, tol=None):
```
arokem, line 1246: I think that it would be great to have a unit test for this function

AFQ/segmentation.py
```diff
@@ -1184,6 +1243,43 @@ def _is_streamline_in_ROIs_parallel(indiv_args, tol, include_roi,
...
+    """
+    Compute the cardinal orientation of each streamline
+    Parmaeters
+    __________
+    streamlines : sequence of 3XN_i arrays
```
arokem, line 1251: Need to finish this docstring

arokem: What is `N_i`?

arokem (commented):
Main comment is whether we should make these part of the default. Why not?

AFQ/tractography.py
36000: This should bring us back towards mAFQ defaults. I don't remember why we chose 10 and 1000 in the first place, but Jason Maya and I had a conversation about these and in most cases these lengths will serve better. One can adjust min length if they need to. The smaller streamlines were also making cleaning harder by changing the distribution.

arokem: That makes sense! Thanks for clarifying.

PR #900, opened 2022-08-26
[ENH] Density task

36000:
Here is a good place to have a discussion about what options should be available for the group density map.

### DIFF
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -808,7 +808,7 @@ class GroupAFQ(object):
 
         out_fname = op.abspath(op.join(
             self.afq_path,
-            f"desc-density_subjects-all_space-MNI_dwi.nii.gz"))
+            f"desc-density_subjects-all_space-MNI_dwi.npy"))
         np.save(out_fname, group_density)
         return out_fname
 
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -776,6 +776,31 @@ class GroupAFQ(object):
         if op.exists(self.afqb_path):
             s3fs.put(self.afqb_path, remote_path, recursive=True)
 
+    def export_group_density(self):
+        """
+        Generate a group density map by combining single subject density maps.
+        """
+        densities = self.export("density_maps", collapse=False)
+        ex_density_img = densities[
+            self.valid_sub_list[0]][
+                self.valid_ses_list[0]]  # for shape and affine
+
+        group_density = np.zeros_like(ex_density_img.shape)
+        self.logger.info("Generating Group Density...")
+        for ii in tqdm(range(len(self.valid_ses_list))):
+            this_sub = self.valid_sub_list[ii]
+            this_ses = self.valid_ses_list[ii]
+            this_density = nib.load(densities[this_sub][this_ses]).get_fdata()
+
+            group_density = group_density + this_density  # TODO: more options here
+
+        out_fname = op.abspath(op.join(
+            self.afq_path,
+            f"desc-density_subjects-all_space-MNI_dwi.nii.gz"))
+        nib.save(nib.Nifti1Image(
+            group_density, ex_density_img.affine), out_fname)
+        return out_fname
+
     def assemble_AFQ_browser(self, output_path=None, metadata=None,
                              page_title="AFQ Browser", page_subtitle="",
                              page_title_link="", page_subtitle_link=""):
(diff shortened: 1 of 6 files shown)


36000:
@jyeatman  thoughts?

36000:
@arokem this is finally ready for review / merge

AFQ/tasks/segmentation.py
```diff
@@ -260,6 +261,26 @@ def export_bundle_lengths(data_imap,
     return counts_df, dict(sources=bundles_files)
 
 
+@pimms.calc("density_maps")
+@as_file('_desc-density_dwi.npy', include_track=True, include_seg=True)
```
arokem, line 265: Why is this saved into an npy file? Rather than a nifti, I mean.

Also, do we need to save metadata about which volume encodes which bundle?

AFQ/api/group.py
```diff
@@ -763,6 +763,42 @@ def upload_to_s3(self, s3fs, remote_path):
...
+        group_density = np.zeros_like(ex_density_init)
+        self.logger.info("Generating Group Density...")
+        for ii in tqdm(range(len(self.valid_ses_list))):
+            this_sub = self.valid_sub_list[ii]
+            this_ses = self.valid_ses_list[ii]
+            this_density = np.load(densities[this_sub][this_ses])
+            if boolify:
+                this_density = this_density.astype(bool)
+
+            group_density = group_density + this_density
```
arokem, line 794: This sums the densities. Do we want to divide by number of subjects at some point to get the mean density?

arokem (commented):
A couple of questions/comments

AFQ/api/group.py
36000: Yes

AFQ/tasks/segmentation.py
36000: Yes I should save that to the metadata. I guess it should be .nii.gz

36000: So right now we are saving a list of the bundles in the metadata. It is in the same order as the volumes. I think this is good enough

36000: see: `source=clean_bundles, bundles=list(bundle_dict.keys()))`

arokem:
Looks like there's still some issue here where the density map is not being generated: https://github.com/yeatmanlab/pyAFQ/actions/runs/3446380558/jobs/5751208459#step:5:1766

36000:
Its likely because I left a .npy somewhere

examples/plot_afq_callosal.py
```diff
@@ -74,6 +76,22 @@
 myafq.export_all()
 
 
+##########################################################################
+# Create Group Density Maps:
+# -------------------------
+#
+# pyAFQ can make density maps of streamline counts per subject/session
+# by doing myafq.export("density_map") . When using GroupAFQ, you can also
+# combine these into one file by doing myafq.export_group_density() .
```
arokem, line 85:
```suggestion
# combine these into one file by calling `myafq.export_group_density()`.
```

examples/plot_afq_callosal.py
```diff
@@ -74,6 +76,22 @@
 myafq.export_all()
 
 
+##########################################################################
+# Create Group Density Maps:
+# -------------------------
+#
+# pyAFQ can make density maps of streamline counts per subject/session
+# by doing myafq.export("density_map") . When using GroupAFQ, you can also
```
arokem, line 84:
```suggestion
# by calling `myafq.export("density_map")`. When using `GroupAFQ`, you can also
```

AFQ/api/group.py
```diff
@@ -776,6 +776,48 @@ def upload_to_s3(self, s3fs, remote_path):
         if op.exists(self.afqb_path):
             s3fs.put(self.afqb_path, remote_path, recursive=True)
 
+    def export_group_density(self, boolify=True):
+        """
+        Generate a group density map by combining single subject density maps.
+
+        Parameters
+        ----------
+        boolify : bool
```
arokem, line 785: Sorry - one more thing: with `boolify` set to true, this will count the number of subjects who have at least one streamline in that location (in the mapped template space). Is that really the intended behavior? I am actually not sure what the "correct" behavior should be. That is, how we should treat voxels that have large discrepancies between subjects.

arokem (commented):
Sorry: one more small question for clarification. What is the intended behavior with many subjects? Right now, with `boolify=True` (the default) I think that it will report not the average density, but the proportion of subjects that have any number of streamlines at every location. Is that really what we want?

jyeatman:
For a group map that is common behavior: The individual maps get thresholded (eg at zero) and then the group map counts the number of subjects who have at least 1 streamline at a given vertex
________________________________

arokem:
Off we go then!

PR #907, opened 2022-09-27
Replace all instances of PipelineDescription by GeneratedBy

arokem:


arokem:
This is a duplicate of #870

PR #909, opened 2022-10-04
DOC: Document S3BIDS access examples.

arokem:
So at least I don't have to stumble around every time I try doing this.


36000:
LGTM

PR #910, opened 2022-10-07
DOC: Adds an example of visualizations using FURY.

arokem:
Following up on the conversation we had at our group meeting about visualizations, here is a documentation example that uses fury to visualize streamlines, core bundles with profiles (emulating Figure 1 [here](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0049790), but with a better colormap 😉), and surface renderings (waypoint ROIs in this case). I tried to write detailed explanations for every step here, but let me know if something is not clear. 

This example generates the following figures step-by-step: 

arc_cst1
arc_cst2
arc_cst3
arc_cst4

### DIFF
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -22,7 +22,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[dev,fury,afqbrowser,plot]
+        conda install -c conda-forge xvfbwrapper
     - name: Build docs
+      env:
+        XVFB: 1
       run: |
         cd docs
         make html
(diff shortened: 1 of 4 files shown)


36000:
LGTM

arokem:
I still need to sort out how to integrate the image pngs into the rendered webpage. I think we already sorted that out in other places, so should be doable, but have not gotten around to it yet.

arokem:
Looks like some custom code was needed to scrape the generated pngs. Didn't need that before because plotly have a custom scraper implemented and we used that. But this now works well, so I am going to go ahead and merge it.

PR #911, opened 2022-10-11
[FIX] Updating DIPY tracking params

36000:
I am still a little confused by what the parameters input to DIPY's tractography algorithm means in terms of units. For example, a max length of 250mm causes problems for many bundles in HBN, however that should be long enough for a typical brain? This was triggered by lowering of max length from 1000mm in recent PR.

### DIFF
--- a/AFQ/tractography.py
+++ b/AFQ/tractography.py
@@ -21,7 +21,7 @@ from AFQ._fixes import (VerboseLocalTracking, VerboseParticleFilteringTracking,
 def track(params_file, directions="prob", max_angle=30., sphere=None,
           seed_mask=None, seed_threshold=0, n_seeds=1, random_seeds=False,
           rng_seed=None, stop_mask=None, stop_threshold=0, step_size=0.5,
-          min_length=20, max_length=250, odf_model="CSD",
+          min_length=20, max_length=300, odf_model="CSD",
           tracker="local"):
     """
     Tractography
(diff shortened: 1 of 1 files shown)


arokem:
Since we are using DIPY's local tracking with `fixed_step` set to True (the default), we believe that the size of the steps is defined in mm (https://github.com/dipy/dipy/blob/a8432c26b42051abf0fc14695b0794e9099476f6/dipy/tracking/localtrack.pyx#L44). This means that max_length and min_length should be in units of numbers of steps. Meaning that with this PR in place, our users are providing these inputs in units of mm (which makes sense) and then we convert (via the code in this PR) to number of steps, which then gets passed to DIPY to do the right thing. Whew.

arokem:
One thought regarding this is that we could verify what we think is correct by running this version on some data and then looking at the distribution of streamline lengths _measured in mm_. We can get that by taking the point-by-point Euclidean distance between nodes in each streamline, after transferring the streamlines to the RASMM space, where things should be, well, in mm. Then, we can ask what is the maximal length of the streamlines in the whole-brain tractography and see where the distribution cuts off. A bit empirical, but might serve to validate our reasoning. WDYT?

36000:
And here is the distribution for those accepted into at least one bundle:
test

36000:
And here for another HBN subjects:
test

36000:
OK, I think when I was checking lengths earlier I was not using the fix. With the fix in this PR, the HBN work with the matlab default max length of 250 and min length of 50. Will push that change and let's see if the automatic tests agree.

36000:
I think part of the confusion here was I was looking at just a bad HBN subject. I think the current setup should work, though 🤞

36000:
@arokem this is ready for review / merge.

arokem:
I am not sure that I understand what the conclusion of the testing you did was. Does the distribution of streamline lengths curtail at 500 steps (250mm)?

36000:
Yes, it looks like there are very few streamlines longer than 500 steps / 250 mm

PR #912, opened 2022-10-12
[ENH] VOF endpoints don't include pAF endpoints

36000:


36000:
The removed rightparietal/leftparietal lines were because they are defined below as well, so those lines were not being used.

PR #913, opened 2022-10-18
[FIX] put afq_metadata.toml in the correct spot

36000:
If output_dir is set, afq_metadata.toml needs to be put there instead of derivatives/afq


arokem:
Does this need to rebased on top of #911 ?

PR #919, opened 2022-10-21
[FIX] don't resample subject-space ROIs unless user provides something

36000:
Currently I think subject-space ROIs are resampled to MNI, which is a bug. 
It used to say:
"
            If None, the template will be overriden when passed to
            an API class."
But I see no evidence that I actually implemented this.


arokem:
Do we need to add tests to keep this from breaking?

36000:
I think we should. do we have some subject-space ROIs for this subject we could use?

arokem:
There's a functionally-defined LV1 for the Stanford HARDI dataset here: https://searchworks.stanford.edu/view/ng782rw8378

This notebook showw how it could be downloaded: https://github.com/arokem/visual-white-matter/blob/master/download-data.ipynb

36000:
What does LV1 mean? lower primary visual cortex? Than should the other ROI between the LGN for the bundle?

arokem:
It's _left_ visual cortex. Maybe we can find another ROI in the aparc reduced file that is also in that same collection?

36000:
@arokem  I brought in your changes from the PR in my repository. This is now ready for review

AFQ/tests/test_api.py
```diff
@@ -735,10 +736,37 @@ def test_AFQ_data_waypoint():
...
+        bids_path,
+        "derivatives/vistasoft/sub-01/ses-01/dwi")
+    freesurfer_folder = op.join(
+        bids_path,
+        "derivatives/freesurfer/sub-01/ses-01/anat")
+    lv1_fname = op.join(
+        freesurfer_folder,
+        "sub-01_ses-01_desc-LV1_anat.nii.gz")
+    if not op.exists(lv1_fname):
+        r = get(op.join(
```
arokem, line 749: I think it would be cleaner to make a fetcher for this in the data module.

36000: done

36000:

arokem:
Looks good. At some point we should also make an example of the subject-space ROI functionality for endpoints, since this is a pretty cool use-case.

PR #922, opened 2022-10-29
[FIX] Some pediatric bundle dict fixes

36000:
In a later PR we can add an example for this, which we can use to automatically error check this.


arokem:
LGTM

PR #929, opened 2022-12-05
TST: Fix a random seed for this test.

arokem:
On my laptop, I get intermittent failures on this test, when the seed is not fixed.


arokem:
I'm going to merge this, because it does not have any effect outside of that one test, which is only exercised by the nightly CI.

PR #930, opened 2022-12-05
DOC: Fix the random seed for the OR example.

arokem:
This resolves the issue we are currently seeing on master, at least on my laptop.


arokem:
I believe this will fix the breakage in current master, so merging.

PR #931, opened 2022-12-06
Add fwDTI

arokem:


AFQ/tasks/data.py
```diff
@@ -151,14 +153,54 @@ def dti(brain_mask, data, gtab,
...
+    fwdti_params = nib.load(fwdti_params).get_fdata()
+    fwtm = dpy_fwdti.FreeWaterTensorModel(gtab)
+    return dpy_fwdti.FreeWaterTensorFit(fwtm, fwdti_params)
+
+
+@as_file(suffix='_model-FWDTI_desc-diffmodel_dwi.nii.gz')
+@as_img
+def fwdti(brain_mask, data, gtab):
+    """
+    full path to a nifti file containing parameters
```
arokem, line 176: This docstring needs to be edited.

arokem: We just need a description, not the parameter section for this function

arokem:
Remaining work: 

- [ ] Create an example that uses fwDTI and see that it works as expected.
- [ ] Make sure documentation is up to spec.

arokem:
I think this is mostly done. Any ideas on where to add this to existing testing? I am thinking of adding it to the list of already calculated scalars in `test_api.test_AFQ_data_waypoint`. It would make that test a little longer, but maybe it seems the most straightforward approach.

arokem:
New test errors might be related to the new release of numpy that happened yesterday.

arokem:
It looks like all that remains here is the OR example: https://github.com/yeatmanlab/pyAFQ/actions/runs/3741923319/jobs/6352177395#step:5:1817, which also seems broken on master. It's puzzling, because it works locally for me. We might be into similar issues to what we are seeing with the SLR tests, where we don't actually have control over the randomness generated by probabilistic tracking. I think I'm going to try deterministic tracking here and see whether that works.

arokem:
Looks like using the HBN data sorted out the OR example issues! I wonder if it would sort out our SLR woes as well... 

I still need to figure out how to embed the montage image in the example webpage. I might do that tomorrow.

36000:
When this is done, let me know so I can rebase #937 on top of this, and see if it works then. I think SLR works there now, I just need the OR example changes.

examples/plot_optic_radiations.py
```diff
@@ -55,39 +83,100 @@
...
+# the "L_OR" bundle. After verifying that streamlines were found, we ask for a
+# montage with an axial view and display that. After creating the montage, we
+# can also ask for an interactive browser-based view of the bundles to be
+# displayed.
+
+if len(aus.SegmentedSFT.fromfile(
+    my_afq.export("clean_bundles")["NDARAA948VFH"]).get_bundle(
+        "L_OR").streamlines) > 1:
     montages = my_afq.montage("L_OR", (1, 1), "Axial")
     my_afq.combine_bundle("L_OR")
```
arokem, line 150: @36000 :  why is `combine_bundle` called here? At this point, the image has presumably already been created.

36000: This is a smoke test for that function. It's the only place it is called. It could probably be moved into it's own, more comprehensive test using example streamlines but it would not be trivial, as it is attached and enmeshed with the GroupAFQ object.

arokem: I'd rather not include this test here. I think that the examples should primarily serve as guidance for users getting acquainted with the software. Admittedly, we need to do some work in some of the other examples, to move them in that direction.

arokem:
I think this is done now. @36000 : when you get a chance, could you please take a look? If it looks OK to you, go ahead and merge, so that you can rebase #937 on top of this.

PR #932, opened 2022-12-06
TST: Fixes another failing nightly test by setting the seed.

arokem:
https://github.com/yeatmanlab/pyAFQ/actions/runs/3629543211/jobs/6121870288#step:6:141


arokem:
Merging immediately, since this will only be tested as nightly.

PR #936, opened 2022-12-09
[FIX] Import tract docs

36000:
Two things here:
1. Update docs regarding name change to import_tract for importing custom tractographies
2. Only run additional bundle recognition like endpoint filtering and orientation filtering if streamlines are found in the first place

### DIFF
--- a/AFQ/segmentation.py
+++ b/AFQ/segmentation.py
@@ -696,93 +696,95 @@ class Segmentation:
 
             possible_fibers = streamlines_in_bundles[:, bundle_idx] > 0
             possible_fibers = np.where(possible_fibers)[0]
-            og_possible_fibers = possible_fibers.copy()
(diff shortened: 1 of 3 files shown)


36000:
@arokem this is ready for review / merge. All of the changes in AFQ/segmentation.py are because of indent after: `if len(possible_fibers) > 0:` (so not a lot has actually changed here)

36000:
There is already this line: https://github.com/yeatmanlab/pyAFQ/blob/e73ea77d2a2e2c3cbedd485ccc271572e0fabced/AFQ/segmentation.py#L693

arokem:
Oh! That works. I'll wait for the docbuild to finish and then merge this.

arokem:
OR still seems to be pretty finicky. But I think we should solve that elsewhere, so merging.

Issue #908, opened 2022-10-04
Should we make s3bids its own library? 


arokem:
Unless I am missing something, it seems to have only rather tenuous dependencies on the rest of pyAFQ, and I can imagine use-cases outside of pyAFQ, where the pyAFQ dependencies may be considered onerous.

arokem:
I forgot about https://github.com/nrdg/cloud_bids_layout, which should eventually evolve into that. 

For the time being, I wonder whether we should add some kind of warning to the s3bids module here, to communicate that this will eventually become its own freestanding thing. 

If we want to do that, might be good to do that for the 1.0 release, so that anyone using this knows that this will not be a long-standing feature they can depend on.

36000:
Closed by #1109

PR #943, opened 2023-01-02
WIP: s3bids deprecation

arokem:
See #908


36000:
This says WIP. What else is needed to put this over the finish line?

arokem:
If you think that's a good idea, I was thinking of raising a deprecation warning whenever s3bids is imported. 

Something like eca0074f

arokem:
I guess we can be more precise and set ourselves a kind of deadline, by saying something like: "will be removed no later than pyAFQ 2.0", or somesuch.

arokem:
BTW: if we are punting this as far as 2.0, I think that it's fine to go ahead with the release with s3bids in it, and then add this warning in the 1.0 cycle.

PR #947, opened 2023-01-14
MAINT: pep517 => build.

arokem:
Apparently "pep517" is being replace by "build".


arokem:
As this is just a maintenance change with no actual effect on the software itself, I will go ahead and merge without waiting for the CI (which is definitely the right thing to do on the eve of a long weekend).

PR #951, opened 2023-02-07
BF: Calculate mean b0 instead of mean not-b0

### LINKED ISSUE #950: [BUG] Should this have a tilde?
https://github.com/yeatmanlab/pyAFQ/blob/afcc50caa8d58da35570e0bca64091ffbf406035/AFQ/tasks/data.py#L87

I think this is actually calculating the mean when b!=0. The tilde should be removed. For example, in the current setup the ventricles are dark but if the tilde is removed, the ventricles are white, as they should be in a b0.

@arokem  does this make sense?

arokem:
Fixes #950

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -84,7 +84,7 @@ def b0(dwi, data, gtab, img):
     """
     full path to a nifti file containing the mean b0
     """
-    mean_b0 = np.mean(data[..., ~gtab.b0s_mask], -1)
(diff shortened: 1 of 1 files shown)


arokem:
We might need to rethink our median_otsu parameters for brain-masking after this fix is applied. The docbuild failures suggest to me that the mask has changed so much that parts of the frontal cortex are now not included.

36000:
Fixed by #955

PR #953, opened 2023-02-14
[ENH/FIX] Improvements for finding the Visual Pathway

36000:
This includes:
-> Refactoring of the segmentation.py file
-> Adds Generalized Q-Sampling (GQ) model from DIPY, as well as tissue properties from OPDT and CSA
-> Option to use a GQ thresholded map to make a brain mask
-> more flexible apply function in bundle dict
-> fixes bug in VOF example, clarifies that bundle dict entries are not mutable

One feature regression (I am open to adding this back in if we can come up with a reason for it):
-> ROIs should be paths to NIFTI files, NIFTI images, or dictionaries containing BIDS filters. Previously they could be 'ImageDefinition's, which are more flexible for defining combinations of different patterns. But I don't think they are very useful for ROI creation.


AFQ/segmentation.py
```diff
@@ -604,187 +562,234 @@ def segment_afq(self, tg=None):
...
+                    include_roi_tols = [tol**2] * len(bundle_def["include"])
+
+                include_rois = []
+                for include_roi in b_sls["include"]:
+                    include_rois.append(np.array(
+                        np.where(include_roi)).T)
+
+                # with parallel segmentation, the first for loop will
+                # only collect streamlines and does not need tqdm
+                if self.parallel_segmentation["engine"] != "serial":
```
36000, line 704: chunk by number of jobs (resources available if n_jobs=-1)

AFQ/segmentation.py
```diff
@@ -796,26 +801,14 @@ def segment_afq(self, tg=None):
...
+            np.save(bc_path, bundle_votes)
 
-        streamlines_in_bundles = streamlines_in_bundles[possible_fibers]
-        min_dist_coords = min_dist_coords[possible_fibers]
-        if self.roi_dist_tie_break:
-            bundle_choice = np.nanargmin(np.nanmin(min_dist_coords, -1), -1)
-        else:
-            bundle_choice = np.nanargmax(streamlines_in_bundles, -1)
+        bundle_choice = np.argmax(bundle_votes, -1)
+        bundle_choice[bundle_votes.max(-1) == -np.inf] = -1
```
36000, line 811: np.nan instead of -1 here

AFQ/segmentation.py
```diff
@@ -1292,29 +1243,26 @@ def clean_by_orientation(streamlines, primary_axis, tol=None):
...
-        The targets. Numerical values in the atlas array for targets for the
-        first and last node in each streamline respectively, or NX3 arrays with
-        each row containing the indices for these locations in the atlas.
-        If provided a None, this means no restriction on that end.
+    target: Nifti1Image
+        Nifti1Image containing a boolean representation of the ROI.
+    target_idx: int.
+        Index within each streamline to check if within the target region.
+        Typically 0 for startpoint ROIs or -1 for endpoint ROIs.
+        If using flip_sls, this becomes -()
```
36000, line 1262: fill this in

arokem (commented):
A few small things

AFQ/tractography.py
```diff
@@ -143,6 +144,10 @@ def track(params_file, directions="prob", max_angle=30., sphere=None,
         dg = DeterministicMaximumDirectionGetter
     elif directions == "prob":
         dg = ProbabilisticDirectionGetter
+    elif directions == "ptt":
+        dg = PTTDirectionGetter
```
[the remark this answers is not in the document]
36000: I can just remove this for now, it was not that many lines of code and I did not end up using it.

36000:

AFQ/api/bundle_dict.py
```diff
@@ -334,7 +336,7 @@ def cond_load(sl): return load_tractogram(
             if "resampled" not in self._dict[key] or not self._dict[
                     key]["resampled"]:
                 self._resample_roi(key)
-        _item = self._dict[key].copy()
+        _item = ReadOnlyDictProxy(self._dict[key].copy())
```
arokem, line 339: Can you sub-class this with a more informative error message on set attr?

AFQ/tasks/data.py
```diff
@@ -84,7 +89,7 @@ def b0(dwi, data, gtab, img):
     """
     full path to a nifti file containing the mean b0
     """
-    mean_b0 = np.mean(data[..., ~gtab.b0s_mask], -1)
+    mean_b0 = np.mean(data[..., gtab.b0s_mask], -1)
```
arokem, line 92: Isn't this already fixed in master?

AFQ/tasks/data.py
```diff
@@ -285,22 +288,131 @@ def csd(dwi, brain_mask, gtab, data,
...
+    """
+    gqmodel = GeneralizedQSamplingModel(
+        gtab,
+        sampling_length=gq_sampling_length)
+
+    gqi_vector = np.real(
+        squared_radial_component(np.dot(
+            gqmodel.b_vector, default_sphere.vertices.T)
+            * gqmodel.Lambda))
+    ODF = blas.dgemm(
```
36000, line 343: add a test here that it produces the same result as in dipy.

36000: Once you figure out why the dipy version takes too long, start a PR to change this in DIPY

36000:
@arokem  it is finally passing tests. I had reduced the number of seeds in the tutorial examples

arokem:
I see your recent commit seems to have resolved the memory issues on the CI! 🎉 

Just anecdotally: I have been running pyAFQ a lot in recent days, and it is clearly the pARC and VOF selection that causes my laptop to start stuttering, so I suspect that a good place to optimize would be in `clean_by_orientation`. This is also confounded by the fact that these are the last bundles to be identified, so it might just also be mounting memory pressure along the way.

arokem:
So WDYT: ready for a review and potentially merge?

36000:
Yes I think it is ready for review and potentially merge.
This one of the few cases where I really wish this were in C/C++, so that there is better control over the memory. I have a sneaking suspicion that there is some reference to the previous SlsbeingRecognized class in each loop, which would explain the memory build up over the course of the segmentation. However the memory profiler does not indicate that this PR uses much more memory than previously.

AFQ/_fixes.py
```diff
@@ -12,6 +11,23 @@
 
 import math
 
+from dipy.reconst.gqi import squared_radial_component
+from dipy.data import default_sphere
+from scipy.linalg import blas
+
+
+def GWI_ODF(gqmodel, data):
```
arokem, line 19:
```suggestion
def gwi_odf(gqmodel, data):
```

AFQ/_fixes.py
```diff
@@ -12,6 +11,23 @@
...
+    gqi_vector = np.real(
+        squared_radial_component(np.dot(
+            gqmodel.b_vector, default_sphere.vertices.T)
+            * gqmodel.Lambda))
+    ODF = blas.dgemm(
```
arokem, line 24:
```suggestion
    odf = blas.dgemm(
```

AFQ/_fixes.py
```diff
@@ -12,6 +11,23 @@
...
+        alpha=1.,
+        a=data.reshape(-1, gqi_vector.shape[0]),
+        b=gqi_vector
+    ).reshape((*data.shape[:-1], gqi_vector.shape[1]))
+    return ODF
```
arokem, line 29:
```suggestion
    return odf
```

AFQ/models/QBallTP.py
```diff
@@ -0,0 +1,38 @@
...
+        alpha=1.,
+        a=ODF.reshape(-1, invB.shape[0]), b=invB).reshape(
+        (*ODF.shape[:-1], invB.shape[1]))
+
+    return shm, ASO, ISO
+
+
+def anisotropic_index(shm):
+    """
+    Calculates anisotropic index based on spherical harmonics coefficients.
```
arokem, line 32: Where is this definition taken from? (Reference? Link to other code?)

36000:
I am going to compare the memory usage here to the result of #969

36000:
@arokem I think this is ready to merge. It still seems to use slightly more memory than master, but that can be fixed in future PRs.

arokem:
Congratulations on finishing this beast of a PR! 🎉

36000:
Thank you!

PR #955, opened 2023-02-17
[FIX] Add max length constraint, b0 fix

36000:
Finding the FP in this subject with this number of streamlines has always been touchy, so I am not too worried about that.

After these fixes pass the test, we can do a minor release.

### DIFF
--- a/AFQ/_fixes.py
+++ b/AFQ/_fixes.py
@@ -76,11 +76,12 @@ def _verbose_generate_tractogram(self):
             # move to the next streamline if only the seed position
             # and not return all
             len_sl = len(streamline)
-            if len_sl > self.min_length:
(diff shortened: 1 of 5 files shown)


AFQ/_fixes.py
```diff
@@ -76,11 +76,12 @@ def _verbose_generate_tractogram(self):
             # move to the next streamline if only the seed position
             # and not return all
             len_sl = len(streamline)
-            if len_sl > self.min_length:
-                if self.save_seeds:
-                    yield streamline, s
-                else:
-                    yield streamline
+            if len_sl >= self.min_length:
```
arokem, line 79: Is there any kind of test we can add here, to make sure this is doing the right thing now?

PR #961, opened 2023-04-20
[WIP, ENH, FIX] Re-organize the mapping code and remove ITK support

36000:
This PR features:

- Change to saving mappings as forward and backwards replacement fields. This is more general and easier (at least for me) to understand than saving as an image with forward and backward resampled to the same dimensions with affine in a separate numpy file.
- Fixes the "exceptions" argument for the clobber command
- Removes ITK class for now, as it only works in certain cases (not with current versions of qsiprep) (hopefully will be solved in future with nitransforms)


36000:
Now in #981

PR #962, opened 2023-04-24
[ENH] add gpu tracking to pyAFQ

36000:


36000:

AFQ/tractography/gputractography.py
```diff
@@ -0,0 +1,84 @@
...
+
+from nibabel.streamlines.array_sequence import concatenate
+
+from AFQ.tractography import get_percentile_threshold
+
+
+# Modified from https://github.com/dipy/GPUStreamlines/blob/master/run_dipy_gpu.py
+def gpu_track(data, gtab, seed_img, stop_img,
+              seed_threshold, stop_threshold, thresholds_as_percentages,
+              max_angle, step_size, sampling_density, ngpus):
```
arokem, line 21: Needs a docstring

arokem (commented):
This is very cool. I think that it would be nice to make the API of the new "gpu_track" function as similar as possible to the API of "track", so that they can more readily be interchanged (and so that they can share much of the boiler-plate code). So, for example, the seed mask is generated within "track", but it is assumed that it is passed into  "gpu_track" (and always as Nifti1Image!). The new "gpu_track" function needs data and gradient table, whereas "track" needs model params. WDYT?

arokem:
OK - that makes sense. Other than that, I think that this is ready to merge, so if you agree, please go ahead!

PR #981, opened 2023-08-08
[FIX] Remove itk

36000:
Cannot get this to work right now. Will ultimately be handled by nitransforms.

Also fixes a bug in the clobber command.
Also removes unused register_dwi function which is now in dipy as `register_dwi_series`


arokem (approved):
This is fine. Just to see that I understand: the removal of `register_dwi` is unrelated to the main purpose of this PR, right? But can be removed nevertheless, as it's now implemented in DIPY.

36000:
Yes

PR #982, opened 2023-08-10
[ENH] Much faster mahal cleaning

36000:
It my case, it reduced the clean time from around 15 minutes to around 4 seconds (225x speedup).
We do this by using numpy functions instead of for loops. In _fixes for now, but can be moved to dipy later.

### DIFF
--- a/AFQ/_fixes.py
+++ b/AFQ/_fixes.py
@@ -249,7 +249,16 @@ def fast_mahal(sls, stat):
         Mahalonobis distance between each streamline and
         the reference streamline.
     """
-    ref_sl = stat(sls, axis=0)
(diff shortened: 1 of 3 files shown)


arokem:
This is awesome! Probably also more memory efficient (?). Can you write a test that compares it against the slow implementation? We should definitely push this upstream to DIPY.

36000:
Good call. I was incorrectly oversimplifying the covariance calculation, but now I fixed that. It is now around 8 seconds on my test datasets, so still a significant improvement on the dipy code, and with the same result! It is probably more memory efficient, or at least as memory efificent.

36000:
TODO: rename and conform to dipy api and system

arokem (approved):
Looks great now!

PR #983, opened 2023-08-11
Use a simpler way of using tqdm on the tractography

36000:
Now we aren't maintaining our own tracking class. Also removes min length, but this can now be handled in the bundle definitions.

### DIFF
--- a/AFQ/_fixes.py
+++ b/AFQ/_fixes.py
@@ -3,11 +3,6 @@ import numpy as np
 from scipy.special import lpmv, gammaln
 
 from tqdm import tqdm
-from dipy.align import Bunch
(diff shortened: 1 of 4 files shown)


arokem:
What's the motivation to ditch `min_length`?

36000:
We now have this in the bundle recognition step

36000:
Do you think I should make a PR on dipy to add min_length to the tractography up there? Like do you think it is a feature that other DIPY users will find useful?

arokem:
I think so! It's nice to do this while tracks are being generated (i.e., by not `yield`ing them), so you don't have to store these very short streamlines in memory at any point.

PR #985, opened 2023-08-16
[FIX] fix a small bug in the curvature code

36000:


arokem:
Is it possible to test this code with a test that fails without this fix?

AFQ/tests/test_segmentation.py
```diff
@@ -60,6 +73,8 @@ def test_segment():
     npt.assert_equal(len(fiber_groups), 2)
     # Here's one of them:
     CST_R_sl = fiber_groups['CST_R']
+    import dipy.tracking.streamlinespeed as dps
```
arokem, line 76: What is this import doing here?

36000: I forgot to remove this!

36000:
I have no idea why this test is not working, as it seems to pass locally

36000:
it appears to be because I was using an older version of dipy

PR #986, opened 2023-08-21
[ENH] update dipy

36000:


arokem:
Looks like the documentation build ran out of memory on 3.8. I am inclined to merge this, because I doubt this is because of the newer DIPY version. WDYT?

PR #988, opened 2023-08-21
Installation with quotes

arokem:


36000:
what are all these other commits? Otherwise good to merge!

arokem:
I think those commits are switching between branches and then pulling from upstream. But not sure.

Issue #990, opened 2023-08-21
Update to most recent pybids for faster indexing


pierre-nedelec:
`pybids` 0.16.2 has improved indexing speed (https://github.com/bids-standard/pybids/pull/1013). Would be nice to update to this:
https://github.com/yeatmanlab/pyAFQ/blob/0c4a726134d727f908cf4e23d1021d6b22ad3e90/setup.cfg#L33

arokem:
Closed by #991

PR #991, opened 2023-08-21
Update setup.cfg with newer pybids

pierre-nedelec:
Fix issue #990


arokem:
Assuming it doesn't break current functionality, this would be great to upgrade.

arokem (approved):
LGTM!

PR #993, opened 2023-08-21
Update plot_viz.py

mayayab:
fix cst actor to use cst values instead of arc values


arokem:
Good catch! Thanks for the fix!

PR #994, opened 2023-08-21
[ENH] Bundle montage

36000:
add a bundle montage for each subject

bundle_montage

### DIFF
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -549,6 +549,23 @@ class GroupAFQ(object):
 
     clobber = cmd_outputs  # alias for default of cmd_outputs
 
+    def make_all_participant_montages(self, images_per_row=2):
+        """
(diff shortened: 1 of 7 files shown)


36000:
bundle_montage
I think this is a good first draft that I would like to merge, but we can revisit and refine this in the future.

36000:
I think we need to wait to merge this until it is rebased on #997 and the common code is extracted into functions as well.

arokem:
Could you please go ahead and rebase now that #997 is merged?

arokem:
The new images look great! Maybe apart from SLF_L, which fades into the background, because it's dark gray. I wonder whether there is some lighting effect that could mitigate that. Also, I'd try to make these tubes a little thinner or resample the streamlines with a lot more nodes (e.g., 1,000) so that the tubes don't look so "broken up".

36000:
bundle_montage

PR #999, opened 2023-08-22
[Doc] add an example of single subject rois; reduce example time

36000:
reduce example time by allowing them to share derivatives not dependent on tractography, such as mapping, csd, etc


36000:

examples/tutorial_examples/use_subject_space_rois_from_freesurfer.py
```diff
@@ -0,0 +1,135 @@
...
+# only within the endpoint ROIs and not throughout the white matter.
+# This is controlled by passing
+# `"seed_mask": RoiImage()` in the `tracking_params` dict.
+#
+# We only do this to make this example faster and consume less space.
+
+tracking_params = dict(n_seeds=10000,
+                       random_seeds=True,
+                       rng_seed=42,
+                       seed_mask=RoiImage(use_endpoints=True))
```
arokem, line 96: Do I understand correctly that tracking will only happen if it hasn't happened in other examples?

examples/tutorial_examples/use_subject_space_rois_from_freesurfer.py
```diff
@@ -0,0 +1,135 @@
...
+
+#############################################################################
+# Define custom `BundleDict` object
+# --------------------------------
+# In a typical `BundleDict` object, ROIs are passed as paths to Nifti files.
+# Here, we define ROIs as dictionary instead, containing BIDS filters.
```
arokem, line 102:
```suggestion
# Here, we define ROIs as dictionaries instead, containing BIDS filters.
```

arokem (changes requested):
This is mostly great. I think that the example text itself could be slightly improved to be a bit more explanatory. Also, I believe that this currently wouldn't automatically be run, because the file name doesn't start with "plot_". Was that your intention? Or should we run this and produce the outputs?

examples/tutorial_examples/use_subject_space_rois_from_freesurfer.py
```diff
@@ -0,0 +1,135 @@
...
+# data directory (by default in the users home directory) under:
+#
+#   ``AFQ_data/stanford_hardi/``
+#
+# This data represents the required preprocessed diffusion data necessary for
+# intializing the GroupAFQ object (which we will do next)
+#
+# The clear_previous_afq is used to remove any previous runs of the afq object
+# stored in the AFQ_data/stanford_hardi/ BIDS directory. Set it to false if
+# you want to use the results of previous runs.
```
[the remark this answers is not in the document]
36000: That's what recog will do. "track" with also clear the tractography. I will explain that in the example

examples/tutorial_examples/use_subject_space_rois_from_freesurfer.py
36000: Whats being kept is basically the mapping and already fit models.

36000:
I don't think this needs to be run. I am going to merge this for now

PR #1000, opened 2023-08-22
Adds a notebook with a rather complete baby_afq tutorial.

arokem:


arokem:
This notebook now works when I run it locally, but I still need to figure out how to make it run in the context of sphinx-gallery, so before we go ahead and merge this, I'd like to create a separate simpler PR to experiment with that. So, please don't merge for the time being.

PR #818, opened 2022-04-25
WIP: Incorporate use of the trx file format

arokem:

PR #1006, opened 2023-08-23
Incorporate TRX

arokem:
This supersedes #818, because the rebase became too complicated.

### DIFF
--- a/AFQ/tasks/decorators.py
+++ b/AFQ/tasks/decorators.py
@@ -135,13 +135,13 @@ def as_file(suffix, include_track=False, include_seg=False):
                       or op.exists(this_file + ".trk")
                       or op.exists(this_file + ".trx"))
             if not exists:
+                is_trx = False
                 gen, meta = func(*args[:og_arg_count], **kwargs)
 
                 logger.info(f"Saving {this_file}")
                 if isinstance(gen, nib.Nifti1Image):
                     nib.save(gen, this_file)
                 elif isinstance(gen, StatefulTractogram):
-                    is_trx = False
                     this_file = this_file + ".trk"
                     save_tractogram(
                         gen, this_file, bbox_valid_check=False)
--- a/AFQ/tasks/decorators.py
+++ b/AFQ/tasks/decorators.py
@@ -8,6 +8,9 @@ import nibabel as nib
 from dipy.io.streamline import save_tractogram
 from dipy.io.stateful_tractogram import StatefulTractogram
 
+from trx.trx_file_memmap import TrxFile
+from trx.io import save as save_trx
+
 import numpy as np
 
 from AFQ.tasks.utils import get_fname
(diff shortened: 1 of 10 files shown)


AFQ/tasks/segmentation.py
```diff
@@ -229,9 +244,12 @@ def tract_profiles(bundles,
...
-
+    reference = nib.load(scalar_dict[list(scalar_dict.keys())[0]])
+    print(type(reference))
     seg_sft = aus.SegmentedSFT.fromfile(
-        bundles)
+        bundles,
+        reference=reference)
+
     seg_sft.sft.to_rasmm()
     for bundle_name in seg_sft.bundle_names:
```
arokem, line 254: We should be able to use the fact that the trx has groups here in a more sophisticated manner.

AFQ/tasks/decorators.py
```diff
@@ -7,6 +7,10 @@
 import nibabel as nib
 from dipy.io.streamline import save_tractogram
 from dipy.io.stateful_tractogram import StatefulTractogram
+
+from trx.trx_file_memmap import TrxFile
```
arokem, line 11: Imports from trx need to be optional

arokem:
This is now ready for review. It is working on my laptop with the option of running it as usual with trk outputs, or, by adding `trx=True` to the tracking parameters, it can produce trx files. I am not sure that it is making full use of the advantages of trx as currently written, but it's a start.

36000:
What if we made this as optional for storing the results of bundle recognition, but always used for the results of the initial tractogaphy. This would save disk space and I think would have no downsides (we don't load the entire tractographies into MRtrix except on rare occasions, and we can just convert it then).

36000:
If these pass, this is good to merge

AFQ/tasks/tractography.py
```diff
@@ -123,14 +130,27 @@ def streamlines(data_imap, seed, stop,
...
 
-    # perform tractography
-    start_time = time()
-    sft = aft.track(params_file, **this_tracking_params)
-    sft.to_vox()
+    is_trx = this_tracking_params.get("trx", False)
+
+    if is_trx:
+        start_time = time()
+        dtype_dict = {'positions': np.float32, 'offsets': np.uint32}
```
arokem, line 137: I think that we need to change this into half-floats as well

AFQ/tasks/segmentation.py
```diff
@@ -47,9 +54,17 @@ def segment(dwi, data_imap, mapping_imap,
...
-        bbox_valid_check=False)
+    if streamlines.endswith(".trk"):
+        tg = load_tractogram(
+            streamlines, img, Space.VOX,
+            bbox_valid_check=False)
+        is_trx = False
+    elif streamlines.endswith(".trx"):
+        is_trx = True
+        trx = load_trx(streamlines, img)
+        trx.streamlines._data = trx.streamlines._data.astype(np.float32)
```
arokem, line 65: Is it necessary to cast to float32 here? If the trx is saved at float16 this would double memory usage of this without increasing precision.

36000: this was so it would play nice with certain dipy cipy functions which I guess assume float32. We will have to make these functions compatible with float16 in the long run, but I figured this was good for now.

arokem: Makes sense!

PR #1014, opened 2023-09-07
[ENH] Proposal for better bd management

36000:
Currently, the API for users to create a BundleDict from builtin bundles is with a list of names. Here, I propose we separate out the builtin bundles from the BundleDict class. Right now they are in the form functions, which seem to work well, but I am open to ideas here. One benefit of this system is that adding new built-in bundles (like the cerebellar bundles) should be easier.

Another important note also makes the bundle dictionary code easier to read and maintain in my opinion. Even if it is longer, its longer because its more explicit.

I have not updates the tests/examples yet because I would like feedback first.


AFQ/api/bundle_dict.py
```diff
@@ -450,14 +657,32 @@ def copy(self):
...
         """
-        self.gen_all()
         return self.__class__(
             self._dict.copy(),
             seg_algo=self.seg_algo,
             resample_to=self.resample_to,
             resample_subject_to=self.resample_subject_to,
             keep_in_memory=self.keep_in_memory)
 
+    def sub(self, bundle_names):
```
arokem, line 667: What do you think about overloading the `__getitem__`  method for this?

arokem (commented):
Overall, I think this is a good change - removing special cases from within the class logic. 

I have one conceptual suggestion, which is to use `__getitem__` to create sub-dicts (instead of the `sub` method). 

For now, I have tagged this for the pyAFQ 1.2 release, because I don't think that we need to complete this before releasing a 1.1

PR #1019, opened 2023-09-12
add "GeneratedBy"s in `dataset_description.json` for BIDS 1.4+ compliance

### LINKED ISSUE #1018: BIDS v1.4+ requires "GeneratedBy" instead of "PipelineDescription" in `dataset_description.json`
Similar to https://github.com/PennLINC/qsiprep/issues/630.

Current behavior generate a `UserWarning` when instantiating a `GroupAFQ` object:
```
...bids/layout/validation.py:122: UserWarning: The PipelineDescription field was superseded by GeneratedBy in BIDS 1.4.0. You can use ``pybids upgrade`` to update your derivative dataset.
  warnings.warn("The PipelineDescription field was superseded "
```

pierre-nedelec:
Fixes #1018

### DIFF
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -187,7 +187,9 @@ class GroupAFQ(object):
             "Name": bids_description["Name"],
             "BIDSVersion": bids_description["BIDSVersion"],
             "PipelineDescription": {"Name": "pyAFQ",
-                                    "Version": pyafq_version}}
+                                    "Version": pyafq_version},
+            "GeneratedBy": [{"Name": op.basename(self.afq_path),
+                             "Version": pyafq_version}]}
 
         pl_desc_file = op.join(self.afq_path, 'dataset_description.json')
 
--- a/AFQ/data/fetch.py
+++ b/AFQ/data/fetch.py
@@ -1020,7 +1020,8 @@ def organize_cfin_data(path=None):
     to_bids_description(
         dmriprep_folder,
         **{"Name": "CFIN",
-           "PipelineDescription": {"Name": "dipy"}})
+           "PipelineDescription": {"Name": "dipy"},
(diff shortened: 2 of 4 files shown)


AFQ/api/group.py
```diff
@@ -187,7 +187,9 @@ def __init__(self,
             "Name": bids_description["Name"],
             "BIDSVersion": bids_description["BIDSVersion"],
             "PipelineDescription": {"Name": "pyAFQ",
-                                    "Version": pyafq_version}}
+                                    "Version": pyafq_version},
+            "GeneratedBy": [{"Name": op.basename(self.afq_path),
+                             "Version": pyafq_version}]}
```
arokem, line 192: Can you really have both of these side-by-side?

arokem (commented):
We had this in a previous PR, that was eventually closed: https://github.com/yeatmanlab/pyAFQ/pull/870. I think that we were waiting for https://github.com/bids-standard/pybids/issues/864 to get merged in pybids first, but if this now works without that, I would be OK making this change now.

AFQ/api/group.py
arokem, line 192: Also, while we're here, could we add something like: 
```suggestion
                                    "Version": pyafq_version},
            "GeneratedBy": [{"Name": op.basename(self.afq_path),
                             "Version": pyafq_version}], 
            "HowToAcknowledge": "Please cite this paper: https://apertureneuro.org/article/77465-evaluating-the-reliability-of-human-brain-white-matter-tractometry"}
```

arokem (approved):
Looks like this is passing the CI tests, so I will go ahead and merge it.

AFQ/api/group.py
pierre-nedelec: I left the PipelineDescription for compatibility. I think it's just going to be extra and should be deleted in the near future.

PR #1030, opened 2023-09-18
[FIX] set default back to serial

36000:
It seems as though the different parallel engines occasionally cause crashes on different systems. These crashes are also hard to debug. Here I set the default segmentation to be serial, which is often what we end up doing anyways. Users can still set this to parallel if they need to speed up the segmentation step.


36000:
I updated the docs for this parameter

arokem (approved):
LGTM!

Issue #1013, opened 2023-09-01
Issue with tractography output files and BIDS format


pierre-nedelec:
Two output tractography files are created each time, their path differing only with the addition of `algo-AFQ_` for the one with classified bundles. This way of naming does not seem to be BIDS friendly. First `algo` is not a default recognized BIDS entity. This can be worked around for nor now [like this](https://github.com/bids-standard/pybids/issues/1020#issuecomment-1703233782). However a problem remains: if you try to access the file without the algo descriptor, you can't just using bids' `layout.get()` to access the one file that doesn't have "algo" setup.

```
from bids.layout import BIDSLayout, add_config_paths
from bids.exceptions import ConfigError
from pathlib import Path

root_dir = Path("to/your/BIDS/dataset")

try:
    add_config_paths(
        **{
            "bidsafq": "path/to/bidsafq.json"
        }
    )
except ConfigError as e:
    print(e)
except Exception as e:
    raise e

layout = BIDSLayout(root_dir)
layout.add_derivatives(root_dir / "derivatives", config=["bids", "bidsafq"])
params = {
    "subject": layout.get_subject()[0],
    "session": layout.get_sessions()[0],
    "extension": "trx",
    # "return_type": "filename",
    "suffix": "tractography",
    "algorithm": "AFQ",
}
layout.get(**params)
```
<details>
  <summary>Attaching here again the json config</summary>

```json
{
  "name": "bidsafq",
  "entities": [
    {
      "name": "algorithm",
      "pattern": "[_/\\\\]+algo-([a-zA-Z0-9]+)"
    }
  ]
}
```
</details>

With `"algorithm": "AFQ",` it returns one file. Without it, it returns both files. And I don't think there is way to access just the other one?

pierre-nedelec:
A follow up issue here: if running [pyAFQ](https://github.com/yeatmanlab/pyAFQ) after [QSIprep](https://github.com/PennLINC/qsiprep) (@mattcieslak), some entities are used twice, e.g. `desc`. Here's a path demonstrating this:
```
..._run-1_space-T1w_desc-preproc_dwi_space-RASMM_model-probCSD_algo-AFQ_desc-profiles_dwi.csv
```
(the suffix `_dwi_` is also repeated btw)

This makes the `layout.get(desc='profiles')` unusable, as it only parses the first occurence.

[...]

PR #1035, opened 2023-09-21
[ENH] more specific bids entities and fixes to be bids compatible

36000:
This is an attempt to address some of the issues raised in #1013 , and to make pyAFQ outputs more bids compatible

### DIFF
--- a/AFQ/afq_bids_entities.json
+++ b/AFQ/afq_bids_entities.json
@@ -10,12 +10,12 @@
         "pattern": "[_/\\\\]+coordsys-([a-zA-Z0-9]+)"
       },
       {
-        "name": "trkmodel",
-        "pattern": "[_/\\\\]+trkmodel-([a-zA-Z0-9]+)"
+        "name": "trkmethod",
+        "pattern": "[_/\\\\]+trkmethod-([a-zA-Z0-9]+)"
       },
       {
-        "name": "bundlerec",
-        "pattern": "[_/\\\\]+bundlerec-([a-zA-Z0-9]+)"
+        "name": "recogmethod",
+        "pattern": "[_/\\\\]+recogmethod-([a-zA-Z0-9]+)"
       }
     ]
   }
--- a/AFQ/afq_bids_entities.json
+++ b/AFQ/afq_bids_entities.json
@@ -1,0 +1,22 @@
+{
+    "name": "bidsafq",
+    "entities": [
+      {
+        "name": "odfmodel",
+        "pattern": "[_/\\\\]+odfmodel-([a-zA-Z0-9]+)"
+      },
+      {
+        "name": "coordsys",
+        "pattern": "[_/\\\\]+coordsys-([a-zA-Z0-9]+)"
+      },
+      {
+        "name": "trkmodel",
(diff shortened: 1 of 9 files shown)


36000:
produces names like this:
image

arokem:
I think that should keep "space" (rather than "coordsys") as that's already
an existing BIDS entity.

36000:
But space is a different concept. You could use RASMM in template space or in subject space. I think if we use too many existing BIDS entities, we will just create more conflicts in the file name. It is better to have entities that are specific, no?

36000:
drop coordsys, its implied

36000:
This is going to make it so pyAFQ doesn't recognize old datasets. how do we handle this? do we need a function that updates the names? This could get tricky to maintain, especially as the BIDS spec keeps getting updated

36000:
@arokem  the docs are failing with this error that I do not understand:
```
Handler <function run_autoapi at 0x7f394a2d2a60> for event 'builder-inited' threw an exception (exception: 'Module' object has no attribute 'doc')
```

arokem:
Yeah - we're seeing this across all our CI, so this is not specific to this
PR. I wonder if it's related to a recent (Sep 13th) release of Sphinx.
Would you mind pinning sphinx to the 7.2.5 and seeing whether we still get
this?

AFQ/afq_bids_entities.json
```diff
@@ -0,0 +1,22 @@
...
+        "name": "odfmodel",
+        "pattern": "[_/\\\\]+odfmodel-([a-zA-Z0-9]+)"
+      },
+      {
+        "name": "coordsys",
+        "pattern": "[_/\\\\]+coordsys-([a-zA-Z0-9]+)"
+      },
+      {
+        "name": "trkmodel",
+        "pattern": "[_/\\\\]+trkmodel-([a-zA-Z0-9]+)"
```
arokem, line 14: This entity takes the values `prob` or `det`? I am not sure that the word `model` is the most accurate for this. This was referred to as "method" in [this comment on bep16](https://github.com/bids-standard/bids-bep016/issues/43#issue-1069019873), so this would become:

```suggestion
        "pattern": "[_/\\\\]+trkmethod-([a-zA-Z0-9]+)"
```

WDYT?

AFQ/afq_bids_entities.json
```diff
@@ -0,0 +1,22 @@
...
+      },
+      {
+        "name": "bundlerec",
```
arokem, line 17: Along the same lines, maybe this should be:

```suggestion
        "name": "recogmethod",
```

That's so long, though.

arokem (commented):
OK - back to this one, now that we have the CI issue under control 🙌 

I think that the issue of support for old datasets is of real concern, but I am not sure how to handle this. I think that maintaining converters between datasets from different versions of pyAFQ would become a real hassle to keep track of, and I do think that we want to keep improving here. My inclination would be to address issues with old and less compliant naming schemes as users bring this up. I wonder whether we'd need to add a documentation page that keeps track of the naming schemes as they evolve, so that at least there is one place that we could refer to for guidance about this.

PR #1038, opened 2023-09-29
[FIX] pin astroid

36000:


36000:
@arokem it was astroid, not sphinx! Merging and rebasing other PRs.

arokem:
Good catch! How did you realize it was astroid? Has it been reported on the astroid bug tracker?

36000:
I looked more closely at the timeline and realized it was the only one that updated at the right time to cause this error (other than the AWS packages which are always updating)

arokem:
Looks like this will be fixed in an upcoming release of sphinx-autoapi: https://github.com/readthedocs/sphinx-autoapi/issues/407, so we should keep our eyes out for that. This fix should be fine for now, though.

PR #1048, opened 2023-10-27
[FIX] bf for roi clipping

36000:


arokem:
Thanks for the quick fix! This no longer raises the error. But it seems that it clips the streamlines at ROI0, but not at ROI1. 

Here two examples: 

Screenshot 2023-10-27 at 2 33 28 PM

Screenshot 2023-10-27 at 2 31 48 PM

36000:
this new commit should fix the error, but it seems like a few streamlines here and there don't get cut, so I need to figure that out

arokem:
Yeah. I wonder what's up with this one: 
Screenshot 2023-10-27 at 5 42 30 PM

arokem:
We (@teresamg and I) saw similar kinds of things when segmenting cerebellar peduncle bundles.

36000:
@arokem  @teresamg let me know if there are any more bugs in ROI clipping when using this PR. I think it should be fixed now

36000:
image

arokem:
Yes - I think this fixes the issues previously observed, so I can go ahead and merge as soon as the CI is done.

But, while sometimes it works just as I'd expect:

Screenshot 2023-10-30 at 12 14 20 PM

I still wonder why the streamlines exceed the extent of the ROIS in some other cases: 

Screenshot 2023-10-30 at 12 13 29 PM

36000:
I see the same thing for some bundles and I am not sure why yet

36000:
lets merge this for now then fix the other problems in a future PR

PR #1055, opened 2023-11-13
DOC: finish tractometry figure

36000:
This is what it looks like now:
image


36000:
I guess they are filling in where they can. Maybe I can add a green check mark there to push them down

36000:
image

arokem:
Nice hack! 😄

PR #1067, opened 2023-11-29
[FIX] More correct way of calculating mahal distances for cleaning

36000:
I have tried this on the stanford hardi dataset example, but it would be nice to try on a few other subjects before we merge.


36000:
TODO: add test that verifies this with some outside calculation

arokem (approved):

PR #1070, opened 2023-12-06
fixed bundle prob map order

zikaste:

PR #1071, opened 2023-12-07
[FIX] possible fix to pediatric bundle dict; some convenience updates in babyafq example

36000:


AFQ/api/bundle_dict.py
```diff
@@ -287,7 +287,7 @@ def baby_bd():
         "ATR_L", "ATR_R", "CST_L", "CST_R", "CGC_L", "CGC_R",
         "CCH_L", "CCH_R", "FP", "FA", "IFO_L", "IFO_R", "ILF_L",
         "ILF_R", "SLF_L", "SLF_R", "UNC_L", "UNC_R",
-        "ARC_L", "ARC_R", "MdLF_L", "MdLF_R"]
+        "ARC_L", "ARC_R"]
```
arokem, line 290: @zikaste and @grotheer: can you confirm that we're dropping MdLF?

AFQ/api/bundle_dict.py
```diff
@@ -287,7 +287,7 @@ def baby_bd():
         "ATR_L", "ATR_R", "CST_L", "CST_R", "CGC_L", "CGC_R",
         "CCH_L", "CCH_R", "FP", "FA", "IFO_L", "IFO_R", "ILF_L",
```
arokem, line 288: I believe we also need to eliminate "CCH" here, as per https://github.com/yeatmanlab/pyAFQ/pull/1070

arokem (commented):
Is this the only difference between the branch/fork that @zikaste has been using and current master?

AFQ/api/bundle_dict.py
arokem: I'm wrong, as @36000 just explained to me.

AFQ/api/bundle_dict.py
arokem: Seems very likely given code further down: https://github.com/yeatmanlab/pyAFQ/pull/1071/files#diff-96764dfa545e2e28e90a01bb917265f5860ce2d39904199f2b41ca0f98455a6aR303. 

I think this means that we're still going to have MdLF bundles, but that they reuse the probability maps for ILF, for the first pass through.

zikaste: We are not exactly dropping MdLF, we are just using it as a "placeholder" for CCH because in the next step the probmap of MdLF is overwritten as the probmap of ILF anyways.

arokem:
> Hey @36000 : when you get a chance, could you please apply the changes from #1072 into this PR? That way, we'll get the best of both.

I did this in [0a8bacf](https://github.com/yeatmanlab/pyAFQ/pull/1071/commits/0a8bacfcd4f468ba523e469a6f301a31983d8236)

PR #1074, opened 2023-12-11
[FIX] Add exclusion ROI to IFOF

36000:


36000:

arokem:
Oh - interesting - the spurious streamlines follow the arcuate trajectory? 

Have you had a chance to try this out with some data? Would be good to have a few examples where we know IFOF previously captured these spurious streamlines. One example would be `sub-NDARAA948VFH_ses-HBNsiteRU` from HBN (which we use in the visualization examples). We can also use tractoscope to find some other examples in HBN.

36000:
I have tried this on a few subjects and it does remove the spurious streamlines. Also, both the CGC and ARC worked but ARC seemed slightly better placed.

arokem:
OK - let me give this a try with some data as well, and then (barring any objections) we can merge it.

arokem (approved):
OK - I've tried this and looked closely at the ROIs, and I agree that this looks good. Merging!

PR #1087, opened 2024-01-09
[ENH] simplify verbose tracking

36000:


36000:
Supersedes #983

arokem (approved):
Yeah - I think that should work. To the extent that the implementation has diverged from the on in DIPY, we should conform to that one anyway.

PR #1089, opened 2024-01-16
Use only the lowest shell in calculation of the CSD response function.

arokem:
Based on a recent [discussion on the dipy repo](https://github.com/dipy/dipy/discussions/3037#discussioncomment-8148381)


36000:
Once #1091 is merged, rebase on that, and the problem in test_api.py should be fixed, so we can merge this. It's just that DIPY updated the tractography and now some of the smoke tests have to have parameters tweaked.

PR #1091, opened 2024-01-23
[FIX] Fix GPU Tractography

36000:
A lot of the changes here are due to a reorganization of how the DWI files are handled, so that everything goes through `get_data_gtab `

We also update the GPU tractography to add an option to use the recently implemented CSA model

We add a Singularity Def file to use in place of the Dockerfile if wanted

We fix the GPU tractography code by making sure everything is in contiguous, C-order before passing to the GPU code

Finally, we download all of the templates into the Docker/Apptainer images


36000:
This is ready to merge. The failed image tests will succeed on master, they should not be running on a branch.

arokem:
Oh wow! Did you crack this for QSIPrep'd datasets? What was the issue?

arokem:
Oh - good sleuthing! I guess this didn't come up in these other cases
because we generated masks in the code from other things, like Freesurfer
label files. At any rate, good thing you managed to figure this out. I'll
review shortly...

arokem:
I think the setting on this line: https://github.com/yeatmanlab/pyAFQ/blob/18687eed05e718f59114731aed7c6706f6290e65/.github/workflows/docker_pyafq_cuda12.yml#L3 is the reason that this is running on the PR. If not intended to run on the PR, I think that you'd want to change that line.

PR #1094, opened 2024-01-25
[ENH] Some segmentation speed-ups

36000:
- Endpoint ROIs filter by indexing, not cdist
- prob map and cross_streamlines calculated on the fly, in case of bundledict that does not require them


arokem:
How much speedup do we get with this?

36000:
It looks like its ~5x speedup on the the endpoint segmentation step. this can be particularly seen in the accoustic radiation example, because that bundle I defined using endpoints only. In that example, it goes from around 3.5s to 0.5s.

Issue #418, opened 2020-09-10
dmri_segmenter


arokem:
Try this out for skull-stripping in cases where no brain mask is available: https://github.com/captainnova/dmri_segmenter

PR #1096, opened 2024-01-27
[ENH] Seeding on the white matter-gray mattter interface

36000:
Here is a first attempt at seeding in the gray matter white matter interface using hidden Markov Random Field models (as implemented in DIPY) on slightly modified anisotropic power maps (no T1 needed). I also set this as the default seeding to see which tests it would throw

image
image
https://github.com/yeatmanlab/pyAFQ/assets/9562159/cb448ceb-641b-4ff1-b527-c3e9b74bee45

There are some other minor bug fixes in here:
    Fix IFOF color scheme 
    Throw warning and exit in `apply_cmd_to_afq_derivs` if you are trying to clobber() a directory that doesnt exist yet


36000:
Should consider trying this instead: https://github.com/yeatmanlab/pyAFQ/issues/418

36000:
Or use #1129

PR #1097, opened 2024-01-29
[ENH] add anisotropic kurtosis, make dki default?

36000:
This adds a new anisotropic kurtosis parameter. 

It also makes DKI tand fa/md/kfa/mk the default scalars for multishell data, and DTI FA/MD the default for single shell (our current default)

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -773,6 +773,17 @@ def dki_mk(dki_tf):
     return dki_tf.mk()
 
 
+@pimms.calc("dki_kfa")
+@as_file('_odfmodel-DKI_desc-KFA_dwi.nii.gz')
+@as_fit_deriv('DKI')
+def dki_kfa(dki_tf):
+    """
+    full path to a nifti file containing
+    the DKI mean kurtosis file
+    """
+    return dki_tf.kfa()
+
+
 @pimms.calc("dki_ga")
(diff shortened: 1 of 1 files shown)


36000:
image

DKI KFA (left) vs DKI FA (right) on HCP subejct

36000:
@arokem  this is ready for review / merge!

AFQ/tasks/data.py
```diff
@@ -773,6 +772,17 @@ def dki_mk(dki_tf):
     return dki_tf.mk()
 
 
+@pimms.calc("dki_kfa")
+@as_file('_odfmodel-DKI_desc-KFA_dwi.nii.gz')
+@as_fit_deriv('DKI')
+def dki_kfa(dki_tf):
+    """
+    full path to a nifti file containing
+    the DKI mean kurtosis file
```
arokem, line 781:
```suggestion
    the DKI kurtosis FA file
```

arokem (commented):
Thanks for adding KFA! We might want to add a link to this paper somewhere in the documentation: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7028548/, which will help users understand how this differs from the DTI FA. 

Regarding making DKI the default, I would be fine with that, but I'd suggest adding a check that the data can be used for DKI (i.e., that it has multiple non-zero b-values) before starting the run, and raising an informative error when that's not the case, telling users how they should change their call (i.e., where to change into "dti_fa" and "dti_md").

PR #1101, opened 2024-02-03
Make the altair code a little more flexible

36000:


examples/tutorial_examples/plot_001_afq_api.py
```diff
@@ -204,10 +204,6 @@
 for ind in bundle_counts.index:
     if ind == "Total Recognized":
         threshold = 1000
-    elif ind == "Forceps Major":
```
arokem, line 207: No longer an issue?

arokem (approved):

arokem:
Also I'd still love to have some examples of how to use the Altair visualizations

examples/tutorial_examples/plot_001_afq_api.py
36000: We use the other callosal bundles now!

arokem: Ah yes of course!

PR #1103, opened 2024-02-05
[ENH/FIX] Smarter file finding

36000:
This solves two problems:
    1. Fix files being assigned to the wrong session
    2. Allow for subjects to have missing scalars


arokem:
Looks like maybe this broke the nightly PFT test? 

https://github.com/yeatmanlab/pyAFQ/actions/runs/8036268264/job/21949771007

Strangely, I can't reproduce this locally, so it might be something subtle (and annoying).

Issue #1106, opened 2024-02-08
Revert #1086? 


arokem:
Some bleeding-edge users have reported that they dislike the changes introduced in #1086, that now use full tract names in the file-names of the results. One option would be to revert that PR and go back to cryptic acronyms, but provide a table in the documentation that lays out what these acronyms mean (and that we can copy to every paper that we write using the software). OTOH, arguably the new filenames are not that much longer than a BIDSish filename is anyway. For example: 

`sub-100206_coordsys-RASMM_trkmethod-probCSD_recogmethod-AFQ_desc-LeftCorticospinal_tractography.trk`

vs. 

`sub-100206_coordsys-RASMM_trkmethod-probCSD_recogmethod-AFQ_desc-LCST_tractography.trk`

So, I am not strongly pro or con here, and would love to hear users/developers chiming in.

36000:
For now there is a dictionary to convert between the two here: #1107

For me, I think we should keep the longer names, for these reasons:
    I think that the filenames are not that much longer
    there are diffusion MRI researchers who are familiar with major tracts but are not necessarily familiar with the AFQ acronyms, and we want people like that to be able to read our papers more easily
    the smaller acronyms are harder for new people to memorize than the longer names, which have some meaning in them (from what I have seen)
    it makes code and documentation more readable for new people
    
The counter arguments I have seen are:
    People who use AFQ are already familiar with these acronyms
    It's harder to compare results across versions, breaking consistency and requiring additional time/code
    They are shorter and easier to type, taking less space in code and filenames

36000:
One thing you could also do is have abbreviated tract names that are not just acronyms, like:
L InfFroOccipital
L SupLong
L PostArcuate
etc.

arokem:
That reads like a 🦓 mixed with a 🦒 to me! 😄

mayayab:
[...]

PR #1107, opened 2024-02-08
Add conversion dictionary for old acronyms

36000:


arokem:
I think this is already useful, but maybe we can address #1106 more thoroughly here? 

@36000 : do you have any strong opinions about that one?

PR #1111, opened 2024-02-25
[FIX] Set parc/vof cleaning params to default

36000:


arokem (approved):
Straightforward! 

@mayayab : are you able to include images of the characterization that you did that led us to conclude this change is appropriate?

arokem:
I'll merge this for now, and we can add a picture later.

mayayab:
Here are two subjects from our data. For each tract, the left column shows the more restrictive cleaning parameters. @jyeatman thinks the VOF is more accurate when it captures a 'curtain' rather than a thin isolated bundle, so it looks like doing 10 iterations of cleaning is removing too many streamlines. There could be more tweaking done to optimize but based on the few subjects that we have looked at, the defaults seem to be doing better than the restrictive criteria, and it would be more consistent to use the same cleaning as the rest of the bundles (and leave it for users to tweak)
image

PR #1113, opened 2024-03-05
[DOC] try to reproduce and fix nightly test

36000:


36000:
@arokem  this seems to fix the nightly, merging

Issue #1114, opened 2024-03-07
HTML tract colors got scrambled with renaming of tracts?


arokem:
In an older version of pyAFQ, this was what IFOF looked like: 

![old](https://github.com/yeatmanlab/pyAFQ/assets/118582/dea7335e-7736-41b9-999b-2934808df725)

In master, we now get this: 

![new](https://github.com/yeatmanlab/pyAFQ/assets/118582/4d3ada29-b572-43a2-b86e-208207f90141)

I think that we want the old colors. That is, for every tract, the left/right should be the same color but with different saturation.

arokem:
From the image I posted, it looks like this only affected some tracts: IFO, VOF, pARC.

PR #1116, opened 2024-03-13
[FIX] fix bundle colors

36000:
Fixes https://github.com/yeatmanlab/pyAFQ/issues/1114 by correcting a bug in the viz/utils inferior fronto-occipital name, and adding some colors to use for the pARC/VOF
image

### DIFF
--- a/AFQ/viz/utils.py
+++ b/AFQ/viz/utils.py
@@ -42,6 +42,11 @@ tableau_20 = [
     (0.8588235294117647, 0.8588235294117647, 0.5529411764705883),
     (0.09019607843137255, 0.7450980392156863, 0.8117647058823529),
     (0.6196078431372549, 0.8549019607843137, 0.8980392156862745)]
+tableau_extension = [
+    (0.95, 0.85, 0.25),  # Warm, Soft Yellow
+    (0.98, 0.92, 0.5),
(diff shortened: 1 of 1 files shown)


AFQ/viz/utils.py
```diff
@@ -68,6 +73,10 @@
     "Right Uncinate": tableau_20[17], "UF_R": tableau_20[17],
     "Left Arcuate": tableau_20[18], "AF_L": tableau_20[18],
     "Right Arcuate": tableau_20[19], "AF_R": tableau_20[19],
+    "Left Posterior Arcuate": tableau_20[6],
+    "Right Posterior Arcuate": tableau_20[7],
+    "Left Vertical Occipital": tableau_extension[0],
+    "Right Vertical Occipital": tableau_extension[1],
```
arokem, line 79: Looks like you're not using all the colors in the extension set. Is that intentional?

36000: Yes, I added another color to the extension set if we need it in the future, but for now we only have 22 standard bundles (18 original, without the HCC, plus 4 from pARC/VOF). I am making up the colors for the extension set, trying to make them different from the existing 20 colors, but not perfect (for example, here, the VOF color looks similar to the UNC color).

arokem: I think it's fine!

PR #1118, opened 2024-03-25
[ENH] make source field relative in json sidecars

### LINKED ISSUE #1098: Paths should be relative
the "source" field path in many of our json sidecars should be relative, not absolute, as is BIDS standard

36000:
Closes #1098 
Also adds some small changes:
 - make it more clear that you don't need to set the bids info when using participant afq
 - adds more info in profiles sidecar
 - makes co-located messages goto info, not warning, as they occur by default
 - add info on seed/stop mask


36000:
@arokem this is ready for review and merge

PR #1122, opened 2024-03-30
[WIP] GAM Tract Profile Tutorial

jyeatman:
Got this tutorial started


arokem:
I think maybe this fits better here? https://github.com/yeatmanlab/tractable/tree/main/vignettes

We should definitely also add a link to AFQ-Insight and tractable on the front page of the pyAFQ documentation, though.

PR #1124, opened 2024-04-10
Adds ParallelGroupAFQ

teresamg:
Creates a ParallelGroupAFQ class inheriting from GroupAFQ to allow for easier parallelization through pydra. Does not currently work; in progress.

### DIFF
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -40,6 +40,9 @@ try:
 except ImportError:
     using_afqb = False
 
+from dipy.utils.optpkg import optional_package
+pydra, has_pydra, _ = optional_package('pydra')
+
 
 __all__ = ["GroupAFQ", "get_afq_bids_entities_fname"]
 
@@ -954,8 +957,6 @@ class ParallelGroupAFQ():
             output.
             Default: True
         """
-        import pydra
-
         @pydra.mark.task
         def export_sub(pAFQ_kwargs, viz, xforms, indiv):
             pAFQ = ParticipantAFQ(**pAFQ_kwargs)
--- a/AFQ/api/group.py
+++ b/AFQ/api/group.py
@@ -40,8 +40,6 @@ try:
 except ImportError:
     using_afqb = False
 
-import shutil
-
 
 __all__ = ["GroupAFQ", "get_afq_bids_entities_fname"]
 
@@ -149,9 +147,7 @@ class GroupAFQ(object):
             raise TypeError("bids_layout_kwargs must be a dict")
 
         self.logger = logger
-        self.bids_path = bids_path
-        self.output_dir = output_dir
-        self.preproc_pipeline = preproc_pipeline
+
         self.parallel_params = parallel_params
         self.wf_dict = {}
 
@@ -328,6 +324,12 @@ class GroupAFQ(object):
             self.afq_path, "tract_profiles.csv"))
         os.makedirs(op.dirname(out_file), exist_ok=True)
         _df = clean_pandas_df(_df)
+
+        if "submitter_params" in self.parallel_params:
+            if op.exists(out_file):
+                df_prev = pd.read_csv(out_file)
+                _df = _df.append(df_prev)
+            
         _df.to_csv(out_file, index=False)
         return _df
 
@@ -921,6 +923,21 @@ class GroupAFQ(object):
 
 
 class ParallelGroupAFQ(GroupAFQ):
+    def __init__(self, *args, **kwargs):
+        orig = GroupAFQ(*args, **kwargs)
+
+        if "submitter_params" not in orig.parallel_params:
+            orig.parallel_params["submitter_params"] = {"plugin": "cf"}
+
+        if "cache_dir" not in orig.parallel_params:
+            orig.parallel_params["cache_dir"] = None
+
+        self.parallel_params = orig.parallel_params
+        self.subjects = orig.subjects
+
+        self.args = args
+        self.kwargs = kwargs
+    
     def export_all(self, viz=True, afqbrowser=True, xforms=True, indiv=True):
         """ Exports all the possible outputs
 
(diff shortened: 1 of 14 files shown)


AFQ/api/group.py
```diff
@@ -916,6 +916,89 @@ def assemble_AFQ_browser(self, output_path=None, metadata=None,
             sublink=page_subtitle_link)
 
 
+class ParallelGroupAFQ(GroupAFQ):
```
arokem, line 919: I don't think we're using inheritance from GroupAFQ:

```suggestion
class ParallelGroupAFQ():
```

AFQ/api/group.py
```diff
@@ -916,6 +916,89 @@ def assemble_AFQ_browser(self, output_path=None, metadata=None,
...
+            output the b0 in template space.
+            Default: True
+        indiv : bool
+            Whether to output individual bundles in their own files, in
+            addition to the one file containing all bundles. If using
+            the AFQ segmentation algorithm, individual ROIs are also
+            output.
+            Default: True
+        """
+        import pydra
```
arokem, line 980: I'd suggest adding: 

```
from dipy.utils.optpkg import optional_package

pydra, has_pydra, _ = optional_package('pydra')
``` 

At the top of the file for the pydra import, instead of importing within the function.

AFQ/api/group.py
```diff
@@ -916,6 +916,89 @@ def assemble_AFQ_browser(self, output_path=None, metadata=None,
...
+
+        if "cache_dir" not in orig.parallel_params:
+            orig.parallel_params["cache_dir"] = None
+
+        self.parallel_params = orig.parallel_params
+        self.pAFQ_kwargs = [pAFQ.kwargs for pAFQ in orig.pAFQ_list]
+
+        # Rename kwargs and clear "bids_info" and "base_fname"
+        # ParticipantAFQ takes in these parameters under one name but stores
+        # and uses them under another
```
arokem, line 934: This is awkward. Could we change this in the implementation of `ParticipantAFQ`, instead of implementing all this renaming and deleting?

arokem (changes requested):
This is a really neat implementation now! My main suggestion is to change ParticipantAFQ so that the names of kwargs and attributes are well-matched, so that you can make this even neater.

36000:
my thoughts, I will implement these:
1. add a similar looking `export` function to wrap export like the `export_all` implemented here
2. add pydra to setup.cfg
3. add a nighly test on 2 hbn subjects using concurrent futures for pydra

AFQ/api/group.py
```diff
@@ -916,6 +919,64 @@ def assemble_AFQ_browser(self, output_path=None, metadata=None,
...
+        # Submit to pydra
+        export_sub_task = export_sub(
+            pAFQ_kwargs=self.pAFQ_kwargs,
+            viz=viz,
+            xforms=xforms,
+            indiv=indiv,
+            cache_dir=self.parallel_params["cache_dir"]
+        ).split("pAFQ_kwargs")
+
+        with pydra.Submitter(
```
arokem, line 974: We should catch this error: https://github.com/nipype/pydra/issues/630 here (so that our users don't see it)

AFQ/api/participant.py
```diff
@@ -96,14 +96,13 @@ def __init__(self,
...
         self.logger = logging.getLogger('AFQ')
-        self.output_dir = output_dir
+        self.bids_info = _bids_info
 
         self.kwargs = dict(
-            dwi_path=dwi_data_file,
-            bval=bval_file,
-            bvec=bvec_file,
-            results_dir=output_dir,
-            bids_info=_bids_info,
```
arokem, line 106: @36000 : was there a reason for these substitutions? That is, that `dwi_data_file` becomes `dwi_path`, and so on? Would there be a cost associated with this change so that internal kwargs and external names are the same?

36000:
I made some minor changes here. I made pydra a required library for pyAFQ (its pip installable with few dependencies, this will just be easier for most users). We now try to catch that error Ariel mentioned (let me know if this doesnt work). And I added an `export` function based on the `export_all` function. Next I will add the nightly test then I think we can merge this

arokem:
Could you please push that test to this PR? I'm not sure what's up and would like to try to debug locally.

36000:
Sure, I figured out this was due to me using a different pydra version (the latest, 0.23). But now I am getting picking problems. I will push some of the changes I have made, but I think maybe we will have to talk about this in person at some point. I am running into a few issues

36000:
@teresamg mind trying this to see if it works?

teresamg:
Looks like it ran successfully, both locally and on Hyak!

arokem:
Did it generate the combined tract profiles file? I also ran a test on hyak: 

```
import os.path as op

from AFQ.api.group import ParallelGroupAFQ
from AFQ.definitions.image import RoiImage
import AFQ.api.bundle_dict as abd
import AFQ.data.fetch as afd


_, bids_path = afd.fetch_hbn_preproc(
        ["NDARZT957CWG",
         "NDARZU279XR3",
         "NDARZU401RCU",
         "NDARZU822WN3",
         "NDARZV421TCZ",
         "NDARZW262ZLV",
         "NDARZX163EWC",
         ],
        path="/gscratch/scrubbed/arokem/data/")
    
my_afq = ParallelGroupAFQ(
    bids_path=bids_path,
    preproc_pipeline="qsiprep",
    parallel_params={
        "submitter_params": {
            "plugin": "slurm",
            "sbatch_args": "-J test \
                            -p ckpt \
                            --nodes=1 \
                            --cpus-per-task=8 \
                            --gpus=1 \
                            -A escience \
                            --mem=64G \
                            --time=2:00:00 \
                            -o /gscratch/scrubbed/arokem/logs/test.out \
                            -e /gscratch/scrubbed/arokem/logs/test.err \
                            --mail-user=arokem@uw.edu \
                            --mail-type=ALL"
        },
        "cache_dir": "/gscratch/scrubbed/arokem/tmp"
    }
)

my_afq.export_all()
```

which worked great at individual subject level, but I can't find the final combined tract profiles file, which I was expecting to find.

teresamg:
It did for me... does each subject have the *_desc-profiles_dwi.csv?

AFQ/api/group.py
```diff
@@ -916,6 +1004,139 @@ def assemble_AFQ_browser(self, output_path=None, metadata=None,
...
+        ):
+            pAFQ = ParticipantAFQ(
+                pAFQ_kwargs.dwi_data_file,
+                pAFQ_kwargs.bval_file,
+                pAFQ_kwargs.bvec_file,
+                pAFQ_kwargs.results_dir,
+                **pAFQ_kwargs.kwargs)
+            pAFQ.export_all(viz, xforms, indiv)
+
+            for dir in finishing_params["output_dirs"]:
```
36000, line 1120: I am a little confused at what these lines here are trying to do

arokem: The idea behind this is that we want to generate a group-level csv file that contains the merged tract profiles. But we only want to do this once, so each task checks whether all of the other tasks have completed. If they have not, it returns and nothing happens. But if it is the last task running, the condition is always fulfilled and the code continues to lines 1124-1126, which run one last GroupAFQ, which would generate the combined tract profile. I worry that this is not super robust, though, and could potentially run into some funky race conditions. In my own test on hyak, I did not get the combined tract profile did not work. So, maybe we need to do something a bit more direct to create the group-level derivatives.

teresamg: 1120 is checking whether the process is the last process running by trawling for each subject's *_desc-profiles_dwi.csv. If all are present, it exports the GroupAFQ object to create the final tract_profiles.csv.

36000: OK that makes sense

36000: I wonder if pydra has some way to do this through their API

teresamg: Ariel, if all csvs are present, perhaps you can print "output_dirs" and double check that none of the paths are wonky or null? Is there anything unusual in test.out or test.err?

arokem: Oh yeah, good call:

```
slurmstepd: error: *** JOB 18301898 ON g3030 CANCELLED AT 2024-05-14T22:33:02 DUE TO TIME LIMIT ***
```

I will try this again with a longer max time limit than two hours.

arokem: I think it would be tricky to do this through pydra, because we'd have to somehow leave the submitting process running somehow for the duration of all the sub-processes, and that may become very cumbersome for very large datasets. I think that assuming we don't leave the program running until all tasks are completed, we have two options: 

1. Something like what is being done here. 
2. Don't merge and let users do that separately in a separate program. 

I suggest that if my current test works and we get what we expect that we go ahead and merge this PR as is, including this bit of code. I can follow up with a documentation example, based on my current experiments with the HBN data. Users need to know that the final merge into a tract profiles file will fail if _any_ of the sub-tasks fail, so maybe we just need to clarify this in the documentation.

Or is there anything else we need to address before merging?

36000: I agree. This looks good to me!

teresamg: I'm not sure what you mean about the submitting process... it ends as soon as the jobs are submitted, not completed, and its the last job that does the final merge (all of them check for it). Plus GroupAFQ also fails to produce tract_profiles.csv if anything fails.

AFQ/api/group.py
```diff
@@ -146,6 +161,8 @@ def __init__(self,
         if not isinstance(bids_layout_kwargs, dict):
             raise TypeError("bids_layout_kwargs must be a dict")
 
+        parallel_params["engine"] = parallel_params.get("engine", "serial")
```
arokem, line 164: I am getting some strange behaviors on hyak and wondering whether this is causing some issues, so I am going to remove it for now:
```suggestion
```

arokem:
I don't know if this is directly related to the content of this PR, but I am now getting an error in visualizing the standard set of bundles, where the visualization code is raising: 

```
INFO:AFQ:Generating colorful lines from tractography...
KeyError: 'Forceps Minor'

```

Presumably because it's looking for a tract that is now no longer part of the default set of tracts (because we're using the more granular set of CC tracts). 

@36000 : any chance this is related to changes you introduced here to how data is passed between GroupAFQ and ParticipantAFQ? 

If you think this is unrelated, I think we can probably merge this PR, and fix this issue elsewhere.

36000:
Not sure what would cause this, but I think it is unrelated, as removing overlapping bundle definitions happens in the __init__ method of the BundleDict, so I don't think a race condition is causing this error.

PR #1128, opened 2024-04-16
[ENH, WIP] More tracking options

36000:


AFQ/definitions/utils.py
```diff
@@ -72,8 +72,8 @@ def _arglist_to_string(args, get_attr=None):
 def name_from_path(path):
     file_name = op.basename(path)  # get file name
     file_name = drop_extension(file_name)  # remove extension
-    if "-" in file_name:
-        file_name = file_name.split("-")[-1]  # get suffix if exists
+    if "_" in file_name:
```
arokem, line 75: Are we even using this function?

arokem (commented):
What's the purpose of adding the bootstrap method here, if it's so slow as to be questionably useful?

36000:
We are going to try it for the baby data Howard is working with. We can remove it if it turns out not to be useful

36000:
It turns out ptt is also pretty slow, but thats only like two lines, so I think we can keep that.

AFQ/definitions/utils.py
36000: ParticipantAFQ uses this function to generate names for scalars that the user inputs (such as a t1 image from another pipeline). I should make it so the user can also define a name, and we call this function only if a name is not provided.

PR #1129, opened 2024-04-17
[ENH] Added RUMBA-SD to data.py

chiuhoward:
Tried to add RUMBA-SD based on https://docs.dipy.org/stable/examples_built/fiber_tracking/tracking_rumba.html and https://docs.dipy.org/stable/reference/dipy.reconst.html#dipy.reconst.rumba.RumbaSDModel

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -12,7 +12,7 @@ import dipy.reconst.dti as dpy_dti
 import dipy.reconst.fwdti as dpy_fwdti
 import dipy.reconst.msdki as dpy_msdki
 from dipy.reconst.gqi import GeneralizedQSamplingModel
-from dipy.reconst.rumba import RumbaSDModel as dpy_rumba
+from dipy.reconst.rumba import RumbaSDModel
 from dipy.reconst import shm
 from dipy.reconst.dki_micro import axonal_water_fraction
 
@@ -450,9 +450,8 @@ def gq_ai(gq_params):
     return AI, dict(GQParamsFile=gq_params)
 
 
-
-@pimms.calc("rumba_params", "rumba_iso", "rumba_aso")
-def rumba(base_fname, gtab, dwi_affine, data):
+@pimms.calc("rumba_fit")
+def rumba_fit(base_fname, gtab, dwi_affine, data):
     """
     full path to a nifti file containing
     parameters for the Rumba model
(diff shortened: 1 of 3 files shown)


AFQ/tasks/data.py
```diff
@@ -449,6 +450,50 @@ def gq_ai(gq_params):
...
+                         R=1,
+                         voxelwise=False, 
+                         use_tv=False, 
+                         sphere=None,
+                         verbose=False)
+    rumba_fit = rumbamodel.fit(data, mask=white_matter)
+    odf = rumba_fit.odf(rumbamodel, data)  # fODF
+    f_wm = rumba_fit.f_wm  # white matter volume fractions
+
+    GQ_shm, ASO, ISO, f_wm = extract_odf(odf)
```
36000, line 488: get rid of f_wm for this line
`GQ_shm, ASO, ISO = extract_odf(odf)`

36000:
Maybe we could break this up into several tasks. One would be `rumba_fit` which returns `rumba_fit`. Another would be `rumba_params` and that can return the shm params (so this would involve running rumba_fit.odf then  extract_odf). then more could be `rumba_f_wm`, `rumba_f_gm`, etc... and these functions are one-liners that accept the rumba_fit and  return that tissue property

AFQ/tasks/data.py
```diff
@@ -449,6 +450,90 @@ def gq_ai(gq_params):
...
+                         n_iter=600,
+                         recon_type='smf',
+                         n_coils=1,
+                         R=1,
+                         voxelwise=False,
+                         use_tv=False,
+                         sphere=None,
+                         verbose=False)
+    white_matter = (base_fname == 1) | (base_fname == 2)
+    rumba_fit = rumbamodel.fit(data, mask=white_matter)
```
36000, line 484: use brain_mask here instead of white_matter

AFQ/tasks/data.py
```diff
@@ -449,6 +450,90 @@ def gq_ai(gq_params):
...
+    ----------
+    wm_response: able to take response[0] from auto_response_ssst. default = array([0.0017, 0.0002, 0.0002])
+    voxelwise: voxelwise fit or global fit. global fit is faster. global fit requires 4D brain volume in fit. default = False
+    n_iter: number of iterations for fODF estimation. default = 600
+    use_tv: TV regularization turned off by default for efficiency, but can provide more coherent results in fiber tracking. can be only be applied to 4D brain volumes with no singleton dimensions
+    sphere: Sphere on which to construct fODF. If None, uses repulsion724. Default: None
+    """
+
+    rumbamodel = RumbaSDModel(gtab,
+                         wm_response=array([0.0017, 0.0002, 0.0002]),
```
36000, line 472: let the user set these

36000:
@arokem  this is ready for review/merge

AFQ/tasks/data.py
```diff
@@ -449,6 +451,103 @@ def gq_ai(gq_params):
...
 
 
+@pimms.calc("rumba_model")
+def rumba_model(gtab,
+                rumba_wm_response=[0.0017, 0.0002, 0.0002],
+                rumba_gm_response=0.0008,
+                rumba_csf_response=0.003,
+                rumba_n_iter=600):
+    """
+    fit for RUMBA-SD model as documented on dipy reconstruction options
```
arokem, line 461:
```suggestion
    Initializes RUMBA-SD model.
```

AFQ/tasks/data.py
```diff
@@ -449,6 +451,103 @@ def gq_ai(gq_params):
...
+        use_tv=False,
+        sphere=default_sphere,
+        verbose=True)
+
+
+@pimms.calc("rumba_params")
+@as_file(suffix='_odfmodel-RUMBA_desc-diffmodel_dwi.nii.gz')
+@as_img
+def rumba_params(rumba_model, data, brain_mask):
+    """RUMBA SHM Parameters"""
```
arokem, line 500:
```suggestion
    """Fits the RUMBA SD model. """
```

Would be good to finish this docstring, including a "Parameters" and a "Returns" sections.

AFQ/tasks/data.py
```diff
@@ -449,6 +451,103 @@ def gq_ai(gq_params):
...
+@pimms.calc("rumba_fit")
+def rumba_fit(rumba_model, rumba_params):
+    """RUMBA FIT"""
+    return RumbaFit(
+        rumba_model,
+        nib.load(rumba_params).get_fdata())
+
+
+@pimms.calc("rumba_f_csf")
+@as_file(suffix='_odfmodel-RUMBA_desc-CSF_dwi.nii.gz')
```
arokem, line 519: Maybe use `probseg` as suffix? https://bids-specification.readthedocs.io/en/stable/glossary.html#probseg-suffixes

AFQ/tasks/data.py
```diff
@@ -449,6 +451,103 @@ def gq_ai(gq_params):
...
+
+
+@pimms.calc("rumba_f_wm")
+@as_file(suffix='_odfmodel-RUMBA_desc-WM_dwi.nii.gz')
+@as_fit_deriv('RUMBA')
+def rumba_f_wm(rumba_fit):
+    """
+    full path to a nifti file containing
+    the white matter volume fraction for each voxel.Equivalent to sum of fODF.
+    """
```
arokem, line 547:
```suggestion
    """
    Full path to a nifti file containing the white matter 
    volume fraction for each voxel. Equivalent to sum of fODF.
    """
```

What does it mean that it's equivalent to the sum of the fODF?

arokem (commented):
A few suggestions, mostly related to documentation, but also about BIDS file naming.

AFQ/tasks/data.py
chiuhoward: I’m not sure! I just copied it from the dipy documentation here to save users from having to go to the original documentation.

https://docs.dipy.org/stable/reference/dipy.reconst.html#rumbafit

AFQ/tasks/data.py
chiuhoward: @arokem if i’m understanding the docs correctly, each filename can only have one suffix after all of the keyword-value pairs, so the correct name would then be 

`@as_file(suffix='_odfmodel-RUMBA_desc-CSF_probseg.nii.gz’)`?

AFQ/tasks/data.py
chiuhoward: @36000 would this docstring work? I don’t think there’s much room for user configuration in this function, and I’m not too sure what `meta` does… took a stab at defining what the function returns by looking at `QBallTP.py` 
  
   ```
"""
   Takes the fitted RUMBA-SD model as input and returns the spherical harmonics coefficients (SHM).

    Parameters
    ----------
    rumba_fit: use output from rumba_fit as input for rumba_params

    Returns
    ----------
    rumba_shm: spherical harmonics coefficients from an ODF without the isotropic and anisotropic diffusion components

    """
```

File: AFQ/tasks/data.py (hunk 6 above)
Comment (36000) [reply AFQ/tasks/data.py:519]: Thats correct @chiuhoward !

File: AFQ/tasks/data.py (hunk 5 above)
Comment (36000) [reply AFQ/tasks/data.py:500]: This is a part of the task system so we typically only have parameters when theres user input via named arguments, as you point out Howard. We also typically don't have returns because we usually just describe it in the main docstring. These task functions are all only called by pimms. So I think we could do something like this is good:
```
Takes the fitted RUMBA-SD model as input and returns the spherical harmonics coefficients (SHM).
```

File: AFQ/tasks/data.py (hunk 6 above)
Comment (36000) [reply AFQ/tasks/data.py:519]: I think its correct! I just use `dwi` when I don't have anything else to put.

Comment (36000) [reply AFQ/tasks/data.py:519]: So `probseg` is better here

Comment (chiuhoward) [reply AFQ/tasks/data.py:519]: Sorry one last question @36000, `probseg` should be the suffix not just for `@pimms.calc("rumba_f_gm”)` but also for `@pimms.calc("rumba_f_csf”)`, `@pimms.calc("rumba_f_wm”)` and `@pimms.calc("rumba_params”)`?

**Comment by chiuhoward** [comment]
I think I addressed all of the changes and did the new merge correctly? =s @36000 @arokem

**Comment by 36000** [comment]
LGTM, Ariel?

File: AFQ/tasks/data.py [hunk 8]
```diff
@@ -4,6 +4,8 @@
 
 from dipy.io.gradients import read_bvals_bvecs
 import dipy.core.gradients as dpg
+from dipy.data import default_sphere
+from dipy.data import default_sphere
```
arokem, line 8: Duplicate import

```suggestion
```

arokem (commented):
I think some other code got pulled into this PR from other PRs. Could you try merging from master? Also, I spotted one more small thing:

36000:
Actually, now that I look at this, this also removed the refactoring I did so that we could use the SHM from disk from previous runs. Let me see If I can do this merge

36000:
I think this should be good to go! @chiuhoward look it over and make sure i didnt delete any changes u made recently.

chiuhoward:
@36000 looks good! suffixes and docstrings were all that I changed and they’re there

36000:
OK, once the docs finish I will merge, unless @arokem has any more comments. overall, good work everyone!

arokem:
Yeah - looks great! When you get a chance, maybe you can write a PR that has an example? @36000 showed me some really impressive tissue type maps from this model using the Stanford HARDI dataset. Might be nice to show slices from these as part of that.

PR #1130, opened 2024-04-18
Make altair required

36000:
It's a lightweight library and currently it is used in the examples. So, currently, a user could install pyAFQ but get an error from trying to run an example from not having altair. currently, they would have to do pip install pyAFQ[plot]


arokem:
Yeah - seems reasonable to me!

PR #1132, opened 2024-04-29
[REF] Refactor segmentation code into many files and pimms system

36000:
This refactors the massive segmentation class into many files. It should make it a lot easier to change segmentation code in the future, ie, to make it so bundles can depend on the position of other bundles. It also makes recobundles into "just another" step in the pipeline, not as an entirely separate pipeline.


36000:

AFQ/bundle_rec/recognize.py
```diff
@@ -0,0 +1,305 @@
...
+            "Conflicts in bundle assignment detected. "
+            f"{conflicts} conflicts detected in total out of "
+            f"{n_streamlines} total streamlines."
+            "Defaulting to whichever bundle appears first"
+            "in the bundle_dict."))
+    bundle_decisions = np.concatenate((
+        bundle_decisions, np.ones((n_streamlines, 1))), axis=1)
+    bundle_decisions = np.argmax(bundle_decisions, -1)
+
+    # Helper functions for formatting the results
```
arokem, line 214: Maybe move these helper functions out in front of this function? I think that would improve readability (a tiny bit).

AFQ/bundle_rec/preprocess.py
```diff
@@ -0,0 +1,73 @@
...
+import dipy.tracking.streamline as dts
+
+import AFQ.bundle_rec.utils as abu
+
+
+logger = logging.getLogger('AFQ')
+
+
+@pimms.calc("tol", "dist_to_atlas", "vox_dim")
+def incexc_tol(img, dist_to_waypoint, input_dist_to_atlas):
```
36000, line 15: tolerance_mm_to_vox

36000:
I think this is good to go, lets wait for #1124

36000:
@arokem i added some testing and modified the criteria system to not use pimms

AFQ/bundle_rec/recognize.py
```diff
@@ -0,0 +1,310 @@
...
+            "Defaulting to whichever bundle appears first"
+            "in the bundle_dict."))
+    bundle_decisions = np.concatenate((
+        bundle_decisions, np.ones((n_streamlines, 1))), axis=1)
+    bundle_decisions = np.argmax(bundle_decisions, -1)
+
+    # We do another round through, so that we can orient all the
+    # streamlines within a bundle in the same orientation with respect to
+    # the ROIs. This order is ARBITRARY but CONSISTENT (going from ROI0
+    # to ROI1).
```
arokem, line 217: This is no longer ARBITRARY! And in general, this is a lot more systematic as of recently.

AFQ/tests/test_roi.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
```
arokem, line 1: Maybe we should have a tests folder in the new module (which will be called "recognition"?)

arokem (commented):
First pass for now. This is a lot of code 😅

AFQ/bundle_rec/recognize.py
36000: Yes, and we have documented the LPI convention elsewhere. If the user provides their own bundle, however, it could be in any orientation. Also, we don't necessarily use ROIs, although we can. So I think I will remove this comment and instead give a more general one.

36000:
@arokem  this is ready for another round of review

PR #1133, opened 2024-05-02
WIP: Bval error check

teresamg:
Checks if dki is specified when there are less than three bvals

### DIFF
--- a/AFQ/api/participant.py
+++ b/AFQ/api/participant.py
@@ -23,6 +23,10 @@ from AFQ.tasks.utils import get_base_fname
 from AFQ.utils.path import apply_cmd_to_afq_derivs
 from AFQ.viz.utils import BEST_BUNDLE_ORIENTATIONS, trim, get_eye
 
+from AFQ.definitions.image import ScalarImage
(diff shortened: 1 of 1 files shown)


36000:
I wonder if this could be a lot less lines of code if you do this check in dki_params. If you don't want to do this because you want the error to happen during initialization of ParticipantAFQ, not runtime during and export call, you could move it to get_data_plan. At line 1024, there is already this check for defaults:
```
    if "scalars" not in kwargs:
        bvals, _ = read_bvals_bvecs(kwargs["bval_file"], kwargs["bvec_file"])
        if len(dpg.unique_bvals_magnitude(bvals)) > 2:
            kwargs["scalars"] = [
                "dki_fa", "dki_md",
                "dki_kfa", "dki_mk"]
        else:
            kwargs["scalars"] = [
                "dti_fa", "dti_md"]
``` 
You could do something similar nearby. That way its with the data code and not in the participantAFQ code (just the way I have been organizing things).

36000:
I think this will get the job done: https://github.com/yeatmanlab/pyAFQ/pull/1139

PR #1136, opened 2024-05-14
Streamlines task in parallel using ray

asagilmore:
This adds a num_chunks argument to tracking_params which causes the streamlines task to generate multiple trx files in parallel using ray, and then concatenate them at the end.

### DIFF
--- a/AFQ/tasks/tractography.py
+++ b/AFQ/tasks/tractography.py
@@ -159,9 +159,10 @@ def streamlines(data_imap, seed, stop,
 
                 def trx_from_lazy_tractogram(self, lazyt_id, seed, dtype_dict):
                     id = self.objects[lazyt_id]
-                    return self.TrxFile.from_lazy_tractogram(id,
-                                                             seed,
-                                                             dtype_dict)
+                    return self.TrxFile.from_lazy_tractogram(
+                        id,
+                        seed,
+                        dtype_dict=dtype_dict)
 
                 def create_lazyt(self, id, *args, **kwargs):
                     self.objects[id] = self.aft.track(*args, **kwargs)
@@ -177,7 +178,7 @@ def streamlines(data_imap, seed, stop,
             tracking_params_copy['n_seeds'] = this_tracking_params['n_seeds'] // num_chunks
             # create lazyt inside each actor
             tasks = [ray_actor.create_lazyt.remote(object_id, params_file,
-                     **this_tracking_params) for ray_actor in actors]
+                     **tracking_params_copy) for ray_actor in actors]
             ray.get(tasks)
 
             # create trx from lazyt
--- a/AFQ/tasks/tractography.py
+++ b/AFQ/tasks/tractography.py
@@ -146,7 +146,7 @@ def streamlines(data_imap, seed, stop,
     if is_trx:
         start_time = time()
         dtype_dict = {'positions': np.float16, 'offsets': np.uint32}
-        if num_chunks:
+        if num_chunks and num_chunks > 1:
             if not has_ray:
                 raise ImportError("Ray is required to perform tractography in"
                                   "parallel, please install ray or remove the"
--- a/AFQ/tasks/tractography.py
+++ b/AFQ/tasks/tractography.py
@@ -173,6 +173,8 @@ def streamlines(data_imap, seed, stop,
             actors = [TractActor.remote() for _ in range(num_chunks)]
             object_id = 1
 
+            tracking_params_copy = this_tracking_params.copy()
+            tracking_params_copy['n_seeds'] = this_tracking_params['n_seeds'] // num_chunks
             # create lazyt inside each actor
             tasks = [ray_actor.create_lazyt.remote(object_id, params_file,
                      **this_tracking_params) for ray_actor in actors]
--- a/AFQ/tasks/tractography.py
+++ b/AFQ/tasks/tractography.py
@@ -141,7 +141,7 @@ def streamlines(data_imap, seed, stop,
 
     is_trx = this_tracking_params.get("trx", False)
 
-    num_chunks = this_tracking_params.pop("n_chunks", False)
+    num_chunks = this_tracking_params.pop("num_chunks", False)
 
     if is_trx:
         start_time = time()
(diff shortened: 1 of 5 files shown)


AFQ/tasks/tractography.py
```diff
@@ -134,14 +139,62 @@ def streamlines(data_imap, seed, stop,
...
+            tasks = [ray_actor.trx_from_lazy_tractogram.remote(object_id, seed,
+                     dtype_dict) for ray_actor in actors]
+            sfts = ray.get(tasks)
+
+            # cleanup objects
+            tasks = [ray_actor.delete_lazyt.remote(object_id) for ray_actor in
+                     actors]
+            ray.get(tasks)
+
+            sft = concatenate(sfts)
```
arokem, line 191: For clarity, I would 

```
from trx.trx_file_memmap import concatenate as trx_concat
``` 

at the top and then here call: 

```suggestion
            sft = trx_concat(sfts)
```

arokem (changes requested):
Can we incorporate this into the example? 

https://github.com/yeatmanlab/pyAFQ/blob/master/examples/tutorial_examples/plot_001_afq_api.py#L62

@36000 : any objections? We'd need to install ray on the CI.

asagilmore:
To handle non-random seeding, where there may be less that 1 seed per chunk after splitting up the seeds for ray workers. To make this work I am running gen_seeds outside of track function and then passing the seeds as an array here:

`def _gen_seeds(n_seeds, params_file, seed_mask=None, seed_threshold=0,
               thresholds_as_percentages=False,
               random_seeds=False, rng_seed=None):
    if isinstance(params_file, str):
        params_img = nib.load(params_file)
    else:
        params_img = params_file

    affine = params_img.affine

    return gen_seeds(seed_mask, seed_threshold, n_seeds,
                     thresholds_as_percentages,
                     random_seeds, rng_seed, affine)`
                     
And then using that function here: 
`                    seeds = _gen_seeds(this_tracking_params['n_seeds'],
                                       params_file, **this_tracking_params)
                    seed_chunks = np.array_split(seeds, num_chunks)
                    tracking_params_list = [this_tracking_params.copy() for _
                                            in range(num_chunks)]
                    for i in range(num_chunks):
                        tracking_params_list[i]['n_seeds'] = seed_chunks[i]`



This works well, but this bypasses the default arguments from track(), which I have copied to the helper function. If those are ever changed this would not exhibit expected behavior. Is there any way to make this always match the track() default arugments?

AFQ/tasks/tractography.py
asagilmore: fixed in lastest push

AFQ/tasks/tractography.py
```diff
@@ -348,3 +458,18 @@ def get_tractography_plan(kwargs):
...
+
+def _gen_seeds(n_seeds, params_file, seed_mask=None, seed_threshold=0,
+               thresholds_as_percentages=False,
+               random_seeds=False, rng_seed=None):
+    if isinstance(params_file, str):
+        params_img = nib.load(params_file)
+    else:
+        params_img = params_file
+
+    affine = params_img.affine
```
36000, line 471: For this, I would just ask for affine for _gen_seeds instead of params_file. Then, use data_imap['dwi_affine'] to get the affine. It will be the same as the params_img affine, and then you don't have to load the params_img just for the affine

36000: In fact, you probably don't need this function. Instead of:
```
                    seeds = _gen_seeds(this_tracking_params['n_seeds'],
                                       params_file, **this_tracking_params)
```
do:
```
                   seeds = gen_seeds(
                       this_tracking_params['seed_mask'],
                       this_tracking_params['seed_threshold'],
                       this_tracking_params['n_seeds'],
                       this_tracking_params['thresholds_as_percentages'],
                       this_tracking_params['random_seeds'],
                       data_imap["dwi_affine"])
```
I guess this is a little longer which is part of what you are avoiding with the keyword arguments, but I think it's the way to go here.

36000: Another approach would be to have `track` and `gen_seeds` accept `tracking_params` outright instead of accepting each of these arguments individually. You could even then have tracking params be a class with documentation elsewhere, like:
```
class TrackingParams:
    seed_mask = None
    seed_threshold = 0
    ...
```

asagilmore: Some of these arguments are optional, would doing this not override the default arguments with None? For example thresholds_as_percentages would be set to None instead of False here?

36000: No, when your in `streamlines`, your in the task system that gets called from either GroupAFQ or ParticipantAFQ. In here, we try to make sure that everything is set to a reasonable default, sometimes based on other options people set in those objects. In particular, these lines 257-281: https://github.com/yeatmanlab/pyAFQ/blob/6d0ffb699b975dde309dc8ada52e7b70976325f4/AFQ/tasks/tractography.py#L257
make sure that thresholds_as_percentages will either be directly set by the user whatever the default is in `track`

asagilmore: Got it! Thanks for your help, the new commit implements your advice.

AFQ/tasks/tractography.py
```diff
@@ -134,14 +142,109 @@ def streamlines(data_imap, seed, stop,
...
+
+                def delete_lazyt(self, id):
+                    if id in self.objects:
+                        del self.objects[id]
+            actors = [TractActor.remote() for _ in range(num_chunks)]
+            object_id = 1
+            tracking_params_list = []
+
+            # random seeds case
+            if isinstance(this_tracking_params.get("n_seeds"), int) and \
```
arokem, line 185: When there is no default value, it is probably clear to use square-bracket indexing:

```suggestion
            if isinstance(this_tracking_params["n_seeds"], int) and \
```

AFQ/tasks/tractography.py
```diff
@@ -134,14 +142,109 @@ def streamlines(data_imap, seed, stop,
...
+                seed_chunks = np.array_split(n_seeds, num_chunks)
+                tracking_params_list = [this_tracking_params.copy() for _ in
+                                        range(num_chunks)]
+
+                for i in range(num_chunks):
+                    tracking_params_list[i]['n_seeds'] = seed_chunks[i]
+
+            else:
+                seeds = gen_seeds(
+                    this_tracking_params.get('seed_mask', None),
```
36000, line 211: You don't need the get() calls here either, as this_tracking_params should already be filled out previously in the task system. So these can be:
```
this_tracking_params['seed_mask']
```
etc.

AFQ/tasks/tractography.py
asagilmore: got it, fixed in most recent commit.

AFQ/tasks/tractography.py
asagilmore: good to know thanks! Switched them to [] index's in the most recent commit

36000:
From my perspective this looks good now. You can add `ray` in setup.cfg (https://github.com/yeatmanlab/pyAFQ/blob/2889d7afc48ffc9d700cce73a6a7416c55721588/setup.cfg#L97) by adding this to the end:
```
ray =
    ray

all =
    %(dev)s
    %(fury)s
    %(fsl)s
    %(afqbrowser)s
    %(plot)s
    %(trx)s
    %(ray)s
```

Then you can also try making it so the `plot_001_afq_api.py` example (which already uses trx) does `num_chunks`, so that this implementation will also be automatically tested. then, I think we can merge.

arokem:
+1

Have you had a chance to run some more experiments with this? Are you still consistently seeing pretty substantial speedup with this?

asagilmore:
> +1
> 
> Have you had a chance to run some more experiments with this? Are you still consistently seeing pretty substantial speedup with this?

yes, it shows over a 25 times speedup on a 72 core machine. 

image

image

arokem:
Nice! ⚡

I think that we can mark this ready to review and after you add it to the dependencies and  documentation, we can go ahead and merge it. For now, we can keep this as an optional feature, but we might consider setting this as default on some future release (probably after we let users kick the tires on this some more).

asagilmore:
> Nice! ⚡
> 
> I think that we can mark this ready to review and after you add it to the dependencies and documentation, we can go ahead and merge it. For now, we can keep this as an optional feature, but we might consider setting this as default on some future release (probably after we let users kick the tires on this some more).

I think I have correctly added it to the documentation and dependencies, but please let me know if I have done it incorrectly I am not sure exactly where everything is supposed to be. I have added ray as a dependency under the TRX option, let me know if you think I should separate it into its own option.

examples/tutorial_examples/plot_001_afq_api.py
```diff
@@ -55,14 +55,17 @@
 # ---------------------------------------
 # We make create a `tracking_params` variable, which we will pass to the
 # GroupAFQ object which specifies that we want 25,000 seeds randomly
-# distributed in the white matter.
+# distributed in the white matter. We also set num_chunks to True,
+# which will use ray to parallelize the tracking across each core.
```
arokem, line 59: Should this say the following instead? 
```suggestion
# which will use ray to parallelize the tracking across all cores.
```

examples/tutorial_examples/plot_001_afq_api.py
```diff
@@ -55,14 +55,17 @@
...
+# This can be removed to process in serial.
```
arokem, line 60: Also could add here: 
```suggestion
# This can be removed to process in serial or set to use a particular distribution of work by setting `n_chunks` to an integer number.
```

setup.cfg
```diff
@@ -93,6 +93,8 @@ plot =
     ipython>=7.13.0,<=7.20.0
 trx =
     trx-python
+ray =
+    ray
```
arokem, line 97: We could combine these: 

```suggestion
    ray
```

If we go with that, you would have to remove the "ray" option below on L106

arokem (commented):
Just some small comments. Your last commit is now running on the CI, so we'll see how that works on the documentation build action.

arokem:
I've set this to "Ready for review" and removed the "WIP" from the title of the PR, because I think we're close.

setup.cfg
asagilmore: changed in last commit

asagilmore: done!

PR #1138, opened 2024-05-18
[FIX] different CST ROI 2 for L/R

36000:


36000:
image

arokem:
OK - just to document what is going on here: in the current software, the inferior ROI for CST contains waypoints in both the left and the right hemispheres. However, because we are excluding streamlines that cross the midline, it makes no sense to have a waypoint in the opposite hemisphere. This PR points the software to new waypoint ROIs that only have voxels in the relevant hemisphere for each CST. Is that correct, @36000? Does that make sense to you as well, @jyeatman?

36000:
This is correct

jyeatman:
Previously we split fibers that crossed rather than excluding them and that’s why the ROIs were in both hemispheres. But this is fine

Jason D. Yeatman, Ph.D.<https://profiles.stanford.edu/jason-yeatman>
Associate Professor
Stanford University, Graduate School of Education and Department of Psychology
Stanford School of Medicine, Division Developmental Behavioral Pediatrics
Brain Development & Education Lab<http://brainandeducation.com/>
________________________________

arokem (approved):

PR #1139, opened 2024-05-20
Check bvals on dki

36000:
Supersedes #1133

### DIFF
--- a/AFQ/tasks/data.py
+++ b/AFQ/tasks/data.py
@@ -219,7 +219,7 @@ def dki_params(brain_mask, gtab, data):
     full path to a nifti file containing
     parameters for the DKI fit
     """
-    if len(dpg.unique_bvals_magnitude(gtab.bvals)) < 2:
+    if len(dpg.unique_bvals_magnitude(gtab.bvals)) < 3:
         raise ValueError("Not enough bvals to fit DKI.")
     mask =\
         nib.load(brain_mask).get_fdata()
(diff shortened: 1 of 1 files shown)


AFQ/tasks/data.py
```diff
@@ -219,6 +219,8 @@ def dki_params(brain_mask, gtab, data):
     full path to a nifti file containing
     parameters for the DKI fit
     """
+    if len(dpg.unique_bvals_magnitude(gtab.bvals)) < 2:
```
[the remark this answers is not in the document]
36000: You are correct!

AFQ/tasks/data.py
```diff
@@ -219,6 +219,8 @@ def dki_params(brain_mask, gtab, data):
     full path to a nifti file containing
     parameters for the DKI fit
     """
+    if len(dpg.unique_bvals_magnitude(gtab.bvals)) < 3:
+        raise ValueError("Not enough bvals to fit DKI.")
```
arokem, line 223: Also, maybe we can make the error a bit more helpful? Something like: 

```suggestion
        raise ValueError(f"The DKI model requires at least 2 non-zero b-values, but you provided {len(dpg.unique_bvals_magnitude(gtab.bvals))} b-values (including b=0).")
```

36000: done!

arokem (approved):

PR #1140, opened 2024-05-23
Test against numpy 2.0

arokem:


arokem:
Rebased on #1165 so we can run the documentation build.

arokem:
I believe this should work when a new version of DIPY is released in https://github.com/dipy/dipy/pull/3364

PR #1142, opened 2024-06-03
[FIX] Remove unused parameter, filter_by_endpoints

36000:
I broke this parameter over a year ago it looks like. I would like to remove it though, instead of fixing it, because it's redundant. If you don't want endpoints, don't put them in your bundledict!


36000:
@arokem  this is ready for review/merge

PR #1143, opened 2024-06-04
[FIX] Update fetch.py with pediatric ARC ROI mapping

chiuhoward:
Current mapping is:
ARC include0 = SLF include0
ARC include1 = SLF exclude0
ARC include2 = SLF include1

This results in premature clipping of the ARC as ARC include2 is between ARC include0 and ARC include1.

Swapping the mapping ensures that ARC is clipped correctly by redefining the ROIs mapped from the SLF.

Before
image

After
image


chiuhoward:
New mapping would therefore be:

ARC include0 = SLF include0  
ARC include1 = SLF include1 
ARC include2 = SLF exclude0

36000:
LGTM!

PR #1144, opened 2024-06-04
[ENH] add system for finding bundles relative to other bundles

36000:


36000:
There is a bug here (which can be see with qsiprep subjects) where lines like `'core': 'Right'` are doing the opposite of what we intend. we need to consider the affine

36000:
moved to https://github.com/tractometry/pyAFQ/pull/7

Issue #1156, opened 2024-08-26
Issues with setting max_bval


audreycluo:
Hi team! I've been trying to run pyAFQ while excluding the highest bvalue. I've updated my optional arguments from having `"max_bval": ""` to the following:
```
"min_bval": "",
"max_bval": 1500, 
"filter_b": true,
"robust_tensor_fitting": false,
"csd_response": "",
```
After changing my `max_bval`, I'm getting this error:
```
WARNING:AFQ:Failed to export warped b0. This could be because your mapping type is only compatible with transformation from template to subject space. 
The error is: boolean index did not match indexed array along dimension 3; dimension is 398 but corresponding boolean dimension is 172
```
For context, I've run pyAFQ through qsiprep 0.22.0, which includes pyAFQ optional arguments as part of a json file, and things have worked beautifully before. In the steps before pyAFQ, I use mrtrix msmt csd and tractography with all my b-values. Now that I'm excluding the max bval, I’m wondering if this is tripping up the input into pyAFQ somehow? I assume I should be able to use all the available b-values for mrtrix, then exclude my highest b-values for pyafq right?

arokem:
My guess, based on this error, is that we have a bug that has to do with the b-value threshold being applied in some places, but not in others, so you get this kind of mismatch here. Unfortunately, I am not going to have time to look into this more closely before Friday at the very earliest.

smeisler:
I am going to guess that it comes from the definition of `mapping` (https://github.com/yeatmanlab/pyAFQ/blob/cef7b1f9692f121184077e25b189648cb415263f/AFQ/tasks/mapping.py#L74) which is used in the `export_registered_b0` function  (https://github.com/yeatmanlab/pyAFQ/blob/cef7b1f9692f121184077e25b189648cb415263f/AFQ/tasks/mapping.py#L25).

```
def mapping(base_fname, dwi_data_file, reg_subject, data_imap,
            mapping_definition=None):
    """
    mapping from subject to template space.

    Parameters
    ----------
    mapping_definition : instance of `AFQ.definitions.mapping`, optional
        This defines how to either create a mapping from
        each subject space to template space or load a mapping from
        another software. If creating a map, will register reg_subject and
        reg_template.
        If None, use SynMap()
        Default: None
    """
    reg_template = data_imap["reg_template"]
    if mapping_definition is None:
        mapping_definition = SynMap()
    if not isinstance(mapping_definition, Definition):
        raise TypeError(
[...]

PR #1158, opened 2024-08-29
[FIX] Fix bug in filtering by bval, where b0 calculation used mismatched dwi and gtab

36000:
Fixes the bug found in #1156 . Without the changes, the new test will throw the error:
```
IndexError: boolean index did not match indexed array along dimension 3; dimension is 160 but corresponding boolean dimension is 10
```

### DIFF
--- a/AFQ/tests/test_api.py
+++ b/AFQ/tests/test_api.py
@@ -604,6 +604,15 @@ def test_AFQ_pydra():
     pga.export("dti_fa")
 
 
+def test_AFQ_filterb():
+    _, bids_path, _ = get_temp_hardi()
+    myafq = GroupAFQ(
+        bids_path=bids_path,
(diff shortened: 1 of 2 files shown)


AFQ/tests/test_api.py
```diff
@@ -604,6 +604,15 @@ def test_AFQ_pydra():
     pga.export("dti_fa")
 
 
+def test_AFQ_filterb():
+    _, bids_path, _ = get_temp_hardi()
+    myafq = GroupAFQ(
+        bids_path=bids_path,
+        preproc_pipeline='vistasoft',
+        max_bval=1000)
```
arokem, line 612: Would this actually test this functionality? This data doesn't have b-vals above 1000 does it? I'll merge for now but just wondering whether there's a better test.

arokem (approved):

Issue #1159, opened 2024-09-03
ValueError: Expected subject IDs to match


teresamg:
We've been getting a similar error running GroupAFQ on our own data and the example at https://github.com/yeatmanlab/pyAFQ/blob/master/examples/howto_examples/use_subject_space_rois_from_freesurfer.py:

`
ValueError: Expected subject IDs to match for the retrieved image file and the supplied 'from_path' file. Got sub-01 from image file /home/tmgomez/AFQ_data/stanford_hardi/derivatives/freesurfer/sub-01/ses-01/anat/sub-01_ses-01_desc-leftThal_mask.nii.gz and sub-None from 'from_path' file /home/tmgomez/AFQ_data/stanford_hardi.
`

Running ParticipantAFQ with direct file paths still gives an error (failed to export warped b0). Could this be a bug?

36000:
Good catch! I believe this is a type and this PR fixes it: https://github.com/yeatmanlab/pyAFQ/pull/1160 

The "failed to export warped b0" bug might have been fixed by this recently merged PR: https://github.com/yeatmanlab/pyAFQ/pull/1158

arokem:
@teresamg : could you please confirm that #1160 closes this issue with the data you are working with?

teresamg:
Yes

PR #1160, opened 2024-09-03
[FIX] bug in finding ROI from freesurfer using pybids

36000:


arokem:
Thanks for jumping on this! Do you think we should run that example as part of the CI documentation build?

arokem:
Maybe we do that separately. For now, closes #1159

